PR #170, opened 2021-07-16
[#89] Add placement policy

### LINKED ISSUE #89: Container policy setting
Briefly touched in #47. There is a problem of container policy setting, `CreateBucket` doesn't have anything similar to this and `PutBucketPolicy` solves a bit different problem and implies that bucket already exists while container policy is set upon creation and can't be changed. I only see two possibilities at the moment:
 * misuse `LocationConstraint`, assign a set of magic predefined policies to each of locations and use them (which allows to create different containers, but looks weird)
 * store default policy in the session token (but it probably doesn't need that) or access box (which will make this access id/key combination always create containers with the same policy, but you can always have a number of ids/keys)
 * a combination of two is mapping location to policy in access box

Maybe there is something else we can do.

KirillovDenis:
closes #89 

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -1,6 +1,8 @@
 package handler
 
 import (
+	"context"
+	"encoding/xml"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -11,6 +13,7 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/policy"
 	"github.com/nspcc-dev/neofs-s3-gw/api"
 	"github.com/nspcc-dev/neofs-s3-gw/api/layer"
+	"github.com/nspcc-dev/neofs-s3-gw/creds/accessbox"
 	"go.uber.org/zap"
 )
 
@@ -23,6 +26,11 @@ const (
 
 	publicBasicRule = 0x0FFFFFFF
 )
+
+type createBucketParams struct {
+	XMLName            xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ CreateBucketConfiguration" json:"-"`
+	LocationConstraint string
+}
 
 func (h *handler) PutObjectHandler(w http.ResponseWriter, r *http.Request) {
 	var (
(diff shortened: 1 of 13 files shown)


api/handler/put.go
```diff
@@ -155,3 +173,22 @@ func parseBasicACL(basicACL string) (uint32, error) {
 		return uint32(value), nil
 	}
 }
+
+func (h *handler) writeError(w http.ResponseWriter, r *http.Request, msg, rid string, err error) {
```
masterSplinter01, line 177: Isn't this function too similar to `func (h *handler) registerAndSendError(w http.ResponseWriter, r *http.Request, err error, logText string)`?

KirillovDenis: You're right

masterSplinter01 (approved):

api/layer/layer.go
```diff
@@ -498,11 +498,11 @@ func (n *layer) DeleteObjects(ctx context.Context, bucket string, objects []stri
 	return errs
 }
 
-func (n *layer) CreateBucket(ctx context.Context, p *CreateBucketParams) (*cid.ID, error) {
+func (n *layer) CreateBucket(ctx context.Context, p *CreateBucketParams, boxData *accessbox.Box) (*cid.ID, error) {
```
alexvanin, line 501: The same comment about `boxData` in `p`

alexvanin (approved):
LGTM

api/layer/container.go
```diff
@@ -123,21 +123,14 @@ func (n *layer) containerList(ctx context.Context) ([]*BucketInfo, error) {
 	return list, nil
 }
 
-func (n *layer) createContainer(ctx context.Context, p *CreateBucketParams) (*cid.ID, error) {
+func (n *layer) createContainer(ctx context.Context, p *CreateBucketParams, boxData *accessbox.Box) (*cid.ID, error) {
```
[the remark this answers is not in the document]
KirillovDenis: Fixed

api/layer/layer.go
KirillovDenis: Fixed

api/handler/put.go
```diff
@@ -90,43 +97,42 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
...
+		h.registerAndSendError(w, r, err, "could not parse body")
+		return
+	}
 
-		api.WriteErrorResponse(r.Context(), w, api.Error{
-			Code:           api.GetAPIError(api.ErrBadRequest).Code,
-			Description:    err.Error(),
-			HTTPStatusCode: http.StatusBadRequest,
-		}, r.URL)
+	boxData, err := getBoxData(r.Context())
```
alexvanin, line 110: You can even 
```suggestion
	p.boxData, err = getBoxData(r.Context())
```

PR #171, opened 2021-07-16
[#149] Remove grpc errors

### LINKED ISSUE #149: Remove gRPC status errors
gRPC status errors are suitable for gRPC p2p communication between clients. I don't think we need to work with it it gates. S3 gate defines it's own set of errors so we should use it ubiquitous. 

NeoFS API Client will provide generic go errors in the future (https://github.com/nspcc-dev/neofs-api/issues/150) but now we can rely only on text descriptions. So the common pattern should be to parse API error (basically text description) and throw S3 error.

KirillovDenis:
closes #149 


alexvanin:
Conflicts after https://github.com/nspcc-dev/neofs-s3-gw/pull/173

alexvanin (approved):

PR #175, opened 2021-07-22
ListObjectsV2 cache

### LINKED ISSUE #112: Support continuation-token for ListObjectsV2
IIUC `continutation-token` requires us to save some stable object list on the gateway and return results from it in subsequent requests. Refs. #111.

masterSplinter01:
Closes #112

### DIFF
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -22,8 +22,9 @@ import (
 
 type (
 	layer struct {
-		pool pool.Pool
-		log  *zap.Logger
+		pool  pool.Pool
+		log   *zap.Logger
+		cache ObjectsListV2Cache
 	}
 
 	// Params stores basic API parameters.
@@ -128,8 +129,9 @@ const (
 // and establishes gRPC connection with node.
 func NewLayer(log *zap.Logger, conns pool.Pool) Client {
 	return &layer{
-		pool: conns,
-		log:  log,
+		pool:  conns,
+		log:   log,
+		cache: newListObjectsCache(),
 	}
 }
 
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -55,7 +55,7 @@ type (
 	}
 
 	allObjectParams struct {
-		Bucket     string
+		Bucket     *BucketInfo
 		Delimiter  string
 		Prefix     string
 		StartAfter string
@@ -210,14 +210,19 @@ func (n *layer) ListObjectsV1(ctx context.Context, p *ListObjectsParamsV1) (*Lis
 	var (
 		err    error
 		result ListObjectsInfoV1
+		bkt    *BucketInfo
 	)
 
 	if p.MaxKeys == 0 {
 		return &result, nil
 	}
 
+	if bkt, err = n.GetBucketInfo(ctx, p.Bucket); err != nil {
+		return nil, err
+	}
+
 	allObjects, err := n.listSortedAllObjects(ctx, allObjectParams{
-		Bucket:     p.Bucket,
+		Bucket:     bkt,
 		Prefix:     p.Prefix,
 		Delimiter:  p.Delimiter,
 		StartAfter: p.Marker,
@@ -252,17 +257,22 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
 		err        error
 		result     ListObjectsInfoV2
 		allObjects []*ObjectInfo
+		bkt        *BucketInfo
 	)
 
 	if p.MaxKeys == 0 {
 		return &result, nil
 	}
 
+	if bkt, err = n.GetBucketInfo(ctx, p.Bucket); err != nil {
+		return nil, err
+	}
+
 	if p.ContinuationToken != "" {
 		// find cache with continuation token
 	} else {
 		allObjects, err = n.listSortedAllObjects(ctx, allObjectParams{
-			Bucket:     p.Bucket,
+			Bucket:     bkt,
 			Prefix:     p.Prefix,
 			Delimiter:  p.Delimiter,
 			StartAfter: p.StartAfter,
@@ -292,14 +302,11 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
 func (n *layer) listSortedAllObjects(ctx context.Context, p allObjectParams) ([]*ObjectInfo, error) {
 	var (
 		err       error
-		bkt       *BucketInfo
 		ids       []*object.ID
 		uniqNames = make(map[string]bool)
 	)
 
-	if bkt, err = n.GetBucketInfo(ctx, p.Bucket); err != nil {
-		return nil, err
-	} else if ids, err = n.objectSearch(ctx, &findParams{cid: bkt.CID}); err != nil {
+	if ids, err = n.objectSearch(ctx, &findParams{cid: p.Bucket.CID}); err != nil {
 		return nil, err
 	}
 
@@ -308,14 +315,14 @@ func (n *layer) listSortedAllObjects(ctx context.Context, p allObjectParams) ([]
 	for _, id := range ids {
 		addr := object.NewAddress()
 		addr.SetObjectID(id)
-		addr.SetContainerID(bkt.CID)
+		addr.SetContainerID(p.Bucket.CID)
 
 		meta, err := n.objectHead(ctx, addr)
 		if err != nil {
 			n.log.Warn("could not fetch object meta", zap.Error(err))
 			continue
 		}
-		if oi := objectInfoFromMeta(bkt, meta, p.Prefix, p.Delimiter); oi != nil {
+		if oi := objectInfoFromMeta(p.Bucket, meta, p.Prefix, p.Delimiter); oi != nil {
 			// use only unique dir names
 			if _, ok := uniqNames[oi.Name]; ok {
 				continue
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -13,6 +13,7 @@ import (
 	cid "github.com/nspcc-dev/neofs-api-go/pkg/container/id"
 	"github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-s3-gw/api"
+	"github.com/nspcc-dev/neofs-s3-gw/creds/accessbox"
 	"go.uber.org/zap"
 )
 
@@ -258,6 +259,8 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
 		result     ListObjectsInfoV2
 		allObjects []*ObjectInfo
 		bkt        *BucketInfo
+		cacheKey   string
+		box        *accessbox.Box
 	)
 
 	if p.MaxKeys == 0 {
@@ -268,9 +271,18 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
 		return nil, err
 	}
 
+	if box, err = GetBoxData(ctx); err != nil {
+		return nil, err
+	}
+
+	cacheKey = createKey(box.Gate.AccessKey, bkt.CID)
+
 	if p.ContinuationToken != "" {
-		// find cache with continuation token
-	} else {
+		allObjects = n.cache.Get(p.ContinuationToken, cacheKey)
+		allObjects = trimStartAfter(p.StartAfter, allObjects)
+	}
+
+	if allObjects == nil {
 		allObjects, err = n.listSortedAllObjects(ctx, allObjectParams{
 			Bucket:     bkt,
 			Prefix:     p.Prefix,
@@ -280,13 +292,20 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
 		if err != nil {
 			return nil, err
 		}
+
+		if p.ContinuationToken != "" {
+			allObjects = trimAfterObjectID(p.ContinuationToken, allObjects)
+		}
 	}
 
 	if len(allObjects) > p.MaxKeys {
 		result.IsTruncated = true
 
+		restObjects := allObjects[p.MaxKeys:]
+		n.cache.Put(cacheKey, restObjects)
+		result.NextContinuationToken = restObjects[0].id.String()
+
 		allObjects = allObjects[:p.MaxKeys]
-		// add  creating of cache here
 	}
 
 	for _, ov := range allObjects {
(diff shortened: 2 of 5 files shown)


api/layer/object_cache.go
```diff
@@ -0,0 +1,111 @@
...
+		mtx    sync.RWMutex
+	}
+	cache struct {
+		list    []*ObjectInfo
+		created time.Time
+	}
+)
+
+func (l *listObjectsCache) GetCache(token, key string) []*ObjectInfo {
+	if val, ok := l.caches[key]; ok {
```
KirillovDenis, line 55: If we have  `RWMutex` why don't we protect a read operation?

KirillovDenis:
Did you check the following compatibility tests?
1. s3tests_boto3.functional.test_s3:test_bucket_listv2_continuationtoken_empty 
2. s3tests_boto3.functional.test_s3:test_bucket_listv2_continuationtoken 
3. s3tests_boto3.functional.test_s3:test_bucket_listv2_both_continuationtoken_startafter 

I've tried, tests failed

cmd/s3-gw/app.go
```diff
@@ -175,6 +175,8 @@ func (a *App) Server(ctx context.Context) {
 	srv.Handler = router
 	srv.ErrorLog = zap.NewStdLog(a.log)
 
+	go a.obj.StartTimer()
```
alexvanin, line 178: `a.obj.StartWorkers()` will be nice to see there.

api/layer/object_cache.go
```diff
@@ -0,0 +1,111 @@
...
+	)
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+	c.list = objects
+	c.created = time.Now().UTC()
+	l.caches[key] = c
+}
+
+func (l *listObjectsCache) StartTimer() {
+	for {
```
alexvanin, line 78: I think it will be better to hide new go routine there, instead of using go keyword outside of the module.

api/layer/object_cache.go
```diff
@@ -0,0 +1,111 @@
...
+		select {
+		case <-ticker.C:
+			cur := time.Now()
+			for key, obj := range l.caches {
+				if cur.Sub(obj.created) >= time.Minute {
```
KirillovDenis, line 83: We have `timerDefaultTickerTime` for this

api/layer/object_cache.go
```diff
@@ -0,0 +1,111 @@
...
+	c.created = time.Now().UTC()
+	l.caches[key] = c
+}
+
+func (l *listObjectsCache) StartTimer() {
+	for {
+		select {
+		case <-ticker.C:
+			cur := time.Now()
+			for key, obj := range l.caches {
```
KirillovDenis, line 82: I suppose the read operation needs to be protected here as well.

api/layer/object_cache.go
```diff
@@ -0,0 +1,111 @@
...
+
+	cid "github.com/nspcc-dev/neofs-api-go/pkg/container/id"
+)
+
+/*
+	This is an implementation of a cache for ListObjectsV2 which we return to users by ContinuationToken.
+
+	The cache consists of a map with data and a timer.
+	The map has a key: (access_key from AccessBox) + container_id and a value: list of objects with creation time.
+	The timer every timerDefaultTickerTime checks the map with caches and removes the expired caches.
```
KirillovDenis, line 15: Is it too expensive if we run `time.AfterFunc()` on each key with a removal function? Maybe it will be more profitable if this cache is empty most of the time.

api/layer/object_cache.go
masterSplinter01: Added rwlock

api/layer/object_cache.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"sync"
+	"time"
+
+	cid "github.com/nspcc-dev/neofs-api-go/pkg/container/id"
+)
+
+/*
+	This is an implementation of a cache for ListObjectsV2 which we return to users by ContinuationToken.
```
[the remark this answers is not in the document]
masterSplinter01: Created an issue

api/layer/object_cache.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+		case <-l.quit:
+			ticker.Stop()
+			return
+		}
+	}
+}
+
+func (l *listObjectsCache) StopTimer() {
+	l.quit <- struct{}{}
```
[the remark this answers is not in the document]
masterSplinter01: Fixed

api/layer/layer.go
```diff
@@ -25,8 +25,9 @@ import (
 
 type (
 	layer struct {
-		pool pool.Pool
-		log  *zap.Logger
+		pool  pool.Pool
+		log   *zap.Logger
+		cache ObjectsListV2Cache
```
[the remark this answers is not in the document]
masterSplinter01: Fixed

api/layer/layer.go
```diff
@@ -97,9 +98,16 @@ type (
...
+	// Timer provides basic interface for timer  ¯\_(ツ)_/¯.
+	Timer interface {
+		StartTimer()
+		StopTimer()
+	}
+
 	// Client provides S3 API client interface.
 	Client interface {
 		NeoFS
+		Timer
```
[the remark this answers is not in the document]
masterSplinter01: In App struct we keep layer object as layer.Client interface. But we need to be able start/stop workers from App. That's why we need this.

api/layer/layer.go
```diff
@@ -465,3 +475,12 @@ func (n *layer) ListObjectVersions(ctx context.Context, p *ListObjectVersionsPar
 	}
 	return &res, nil
 }
+
+func (n *layer) StartTimer() {
+	n.cache.StartTimer()
```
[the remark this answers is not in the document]
masterSplinter01: Done

masterSplinter01:
All tests pass, but in case when start after and continuation token are set both it passes falsely. While implementing this I decided to add logic for start-after and continuation token later because of #117.  Or I can add it within this PR in a separate commit.

api/layer/object_cache.go
masterSplinter01: Thanks, time.After() is not too expensive, changed a way to clean records of the cache.

cmd/s3-gw/app.go
```diff
@@ -174,7 +174,6 @@ func (a *App) Server(ctx context.Context) {
 	// Use mux.Router as http.Handler
 	srv.Handler = router
 	srv.ErrorLog = zap.NewStdLog(a.log)
-
```
KirillovDenis, line 177: Not necessary changes

cmd/s3-gw/app.go
```diff
@@ -201,7 +200,6 @@ func (a *App) Server(ctx context.Context) {
 
 	ctx, cancel := context.WithTimeout(context.Background(), defaultShutdownTimeout)
 	defer cancel()
-
```
KirillovDenis, line 204: Not necessary changes

api/handler/response.go
```diff
@@ -34,7 +34,7 @@ type ListObjectsV2Response struct {
 	XMLName               xml.Name       `xml:"http://s3.amazonaws.com/doc/2006-03-01/ ListBucketResult" json:"-"`
 	CommonPrefixes        []CommonPrefix `xml:"CommonPrefixes"`
 	Contents              []Object       `xml:"Contents"`
-	ContinuationToken     string         `xml:"ContinuationToken,omitempty"`
+	ContinuationToken     string         `xml:"ContinuationToken"`
```
KirillovDenis, line 37: Why we need this change? If I understand correctly `ContinuationToken` is included in response only if it was in request.

api/layer/object_cache.go
```diff
@@ -0,0 +1,86 @@
...
+		mtx    sync.RWMutex
+	}
+	cache struct {
+		list    []*ObjectInfo
+		created time.Time
+	}
+)
+
+func newListObjectsCache() *listObjectsCache {
+	c := listObjectsCache{}
```
KirillovDenis, line 49: We can use:
```
return &listObjectsCache{
	caches: make(map[string]cache),
}
```

api/layer/object_cache.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+type (
+	listObjectsCache struct {
+		caches map[string]cache
+		mtx    sync.RWMutex
+	}
+	cache struct {
+		list    []*ObjectInfo
+		created time.Time
```
KirillovDenis, line 44: This field is not used (we only set it).

api/layer/layer.go
```diff
@@ -127,9 +128,11 @@ const (
 // NewLayer creates instance of layer. It checks credentials
 // and establishes gRPC connection with node.
 func NewLayer(log *zap.Logger, conns pool.Pool) Client {
+	cache := newListObjectsCache()
 	return &layer{
-		pool: conns,
-		log:  log,
+		pool:  conns,
+		log:   log,
+		cache: cache,
```
KirillovDenis, line 135: We can set cache like this:
```
return &layer{
	pool:  conns,
	log:   log,
	cache: newListObjectsCache(),
}
```

api/layer/object.go
```diff
@@ -252,31 +259,61 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
...
 		return &result, nil
 	}
 
-	if p.ContinuationToken != "" {
-		// find cache with continuation token
+	if bkt, err = n.GetBucketInfo(ctx, p.Bucket); err != nil {
+		return nil, err
+	}
+
+	if data, ok := ctx.Value(api.BoxData).(*accessbox.Box); ok && data != nil {
```
KirillovDenis, line 275: Maybe we can somehow reuse `getBoxData` from `put.go` file

api/layer/object.go
```diff
@@ -252,31 +259,61 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
...
 			Prefix:     p.Prefix,
 			Delimiter:  p.Delimiter,
 			StartAfter: p.StartAfter,
 		})
 		if err != nil {
 			return nil, err
 		}
+
+		if p.ContinuationToken != "" {
+			for i, obj := range allObjects {
```
KirillovDenis, line 300: I supposes we should extract this:
```
for i, obj := range allObjects {
	if obj.ID().String() == p.ContinuationToken {
		allObjects = allObjects[i:]
		break
	}
}
``` 
to separate function and use it here and in `GetCache`

api/handler/response.go
masterSplinter01: Yep, but how we can differ when ContinuationToken is set as empty and when ContinuationToken is not set? 
It has not passed  s3tests_boto3.functional.test_s3:test_bucket_listv2_continuationtoken_empty because of it. 
I'm not sure it matters a lot if we return a response with empty continuation token.

api/layer/object_cache.go
```diff
@@ -0,0 +1,86 @@
...
+	to other gateways and they should be able to get a list of objects.
+	When we receive the token from the user we just try to find the cache and then we return the list of objects which
+	starts from this token (i.e. objectID).
+*/
+
+// ObjectsListV2Cache provides interface for cache of ListObjectsV2 in a layer struct.
+type (
+	ObjectsListV2Cache interface {
+		GetCache(token string, key string) []*ObjectInfo
+		PutCache(key string, objects []*ObjectInfo)
```
alexvanin, line 29: Don't you think `cache.Get()` and `cache.Put()` looks better than `cache.GetCache()` and `cache.PutCache()`? If it is possible to avoid repetition, I would do so.

api/layer/object_cache.go
alexvanin: I guess with `time.AfterFunc()` we can just

```go
listObjectsCache struct {
	caches map[string][]*ObjectInfo
	mtx    sync.RWMutex
}
```

api/layer/object.go
masterSplinter01: Didn't notice repeated code, fixed

api/layer/object.go
masterSplinter01: Moved getBoxData to layer and made it exported because of cycled dependencies issues

api/layer/layer.go
masterSplinter01: Fixed

api/layer/object_cache.go
masterSplinter01: Maybe we can leave cache struct, doesn't it make code more readable? And this struct make more clear what does it mean `cache` in code.

api/layer/object_cache.go
masterSplinter01: Fixed

api/layer/object.go
KirillovDenis: I suppose now we can change:
```
if data, err := GetBoxData(ctx); data != nil && err == nil {
	accessKey = data.Gate.AccessKey
} else {
	return nil, fmt.Errorf("couldn't get access key from the accessbox: %s", err)
}
cacheKey = createKey(accessKey, bkt.CID)
```
to 
```
if data, err = GetBoxData(ctx); err != nil {
	return nil, err 
}
cacheKey := createKey(data.Gate.AccessKey, bkt.CID)
```

api/handler/response.go
KirillovDenis: Then I'm not sure if we need to pass such test. @alexvanin ?

alexvanin: Let's see what actual AWS will tell us about it (looking into it).

masterSplinter01: Minio response with an error
[kira@kira-20n4000brt /home/kira/work/neofs-s3-gw]$ aws s3api list-objects-v2 --bucket start --endpoint-ur http://127.0.0.1:9000 --continuation-token ""

An error occurred (InvalidArgument) when calling the ListObjectsV2 operation: The continuation token provided is incorrect
[kira@kira-20n4000brt /home/kira/work/neofs-s3-gw]$ aws s3api list-objects-v2 --bucket start --endpoint-ur http://127.0.0.1:9000 --continuation-token ''

An error occurred (InvalidArgument) when calling the ListObjectsV2 operation: The continuation token provided is incorrect

alexvanin: AWS eu-north-1 returns the same error.

```
$ aws --endpoint-url https://s3.eu-north-1.amazonaws.com s3api list-objects-v2  --bucket avbexamplebucket --continuation-token ''

An error occurred (InvalidArgument) when calling the ListObjectsV2 operation: The continuation token provided is incorrect
```

I suggest to keep `omitempty` in this PR and make a note about this test in the docs. 
In separate issue we can mimic this behavior with error output.

alexvanin (approved):

PR #176, opened 2021-07-23
[#49] Add acl translation

### LINKED ISSUE #49: Support ACL operations via translation to eACL request with Session Token
S3 API has [ACL support](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html). Most of it can be translated into NeoFS eACL. The proposal is to convert S3 ACL changes into NeoFS eACL inside S3 Protocol Gateway and sen NeoFS API gRPC requests attaching Session Token with container Context defined.

KirillovDenis:
closes #49 

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -1,0 +1,1158 @@
+package handler
+
+import (
+	"crypto/ecdsa"
+	"encoding/hex"
+	"encoding/json"
+	"encoding/xml"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neofs-api-go/pkg/acl/eacl"
+	"github.com/nspcc-dev/neofs-api-go/pkg/object"
+	"github.com/nspcc-dev/neofs-s3-gw/api"
+	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
+	"github.com/nspcc-dev/neofs-s3-gw/api/layer"
+)
+
+var (
+	writeOps = []eacl.Operation{eacl.OperationPut, eacl.OperationDelete}
+	readOps  = []eacl.Operation{eacl.OperationGet, eacl.OperationHead,
+		eacl.OperationSearch, eacl.OperationRange, eacl.OperationRangeHash}
+	fullOps = []eacl.Operation{eacl.OperationGet, eacl.OperationHead, eacl.OperationPut,
+		eacl.OperationDelete, eacl.OperationSearch, eacl.OperationRange, eacl.OperationRangeHash}
+)
+
+var actionToOpMap = map[string][]eacl.Operation{
+	s3DeleteObject: {eacl.OperationDelete},
+	s3GetObject:    readOps,
+	s3PutObject:    {eacl.OperationPut},
+	s3ListBucket:   readOps,
+}
+
+const (
+	arnAwsPrefix     = "arn:aws:s3:::"
+	allUsersWildcard = "*"
+	allUsersGroup    = "http://acs.amazonaws.com/groups/global/AllUsers"
+
+	s3DeleteObject               = "s3:DeleteObject"
+	s3GetObject                  = "s3:GetObject"
+	s3PutObject                  = "s3:PutObject"
+	s3ListBucket                 = "s3:ListBucket"
+	s3ListBucketVersions         = "s3:ListBucketVersions"
+	s3ListBucketMultipartUploads = "s3:ListBucketMultipartUploads"
+	s3GetObjectVersion           = "s3:GetObjectVersion"
+)
+
+// AWSACL is aws permission constants.
+type AWSACL string
+
+const (
+	aclFullControl AWSACL = "FULL_CONTROL"
+	aclWrite       AWSACL = "WRITE"
+	aclRead        AWSACL = "READ"
+)
+
+// GranteeType is aws grantee permission type constants.
+type GranteeType string
+
+const (
+	acpCanonicalUser         GranteeType = "CanonicalUser"
+	acpAmazonCustomerByEmail GranteeType = "AmazonCustomerByEmail"
+	acpGroup                 GranteeType = "Group"
+)
+
+type bucketPolicy struct {
+	Version   string      `json:"Version"`
+	ID        string      `json:"Id"`
+	Statement []statement `json:"Statement"`
+}
+
+type statement struct {
+	Sid       string    `json:"Sid"`
+	Effect    string    `json:"Effect"`
+	Principal principal `json:"Principal"`
+	Action    []string  `json:"Action"`
+	Resource  []string  `json:"Resource"`
+}
+
+type principal struct {
+	AWS           string `json:"AWS,omitempty"`
+	CanonicalUser string `json:"CanonicalUser,omitempty"`
+}
+
+type ast struct {
+	Resources []*astResource
+}
+
+type astResource struct {
+	Name       string
+	Operations []*astOperation
+}
+
+type astOperation struct {
+	Users  []string
+	Role   eacl.Role
+	Op     eacl.Operation
+	Action eacl.Action
+}
+
+func (h *handler) GetBucketACLHandler(w http.ResponseWriter, r *http.Request) {
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bucketACL, err := h.obj.GetBucketACL(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not fetch bucket acl", reqInfo, err)
+		return
+	}
+
+	if err := checkOwner(bucketACL.Info, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	if err = api.EncodeToResponse(w, h.encodeBucketACL(bucketACL)); err != nil {
+		h.logAndSendError(w, "something went wrong", reqInfo, err)
+	}
+}
+
+func (h *handler) PutBucketACLHandler(w http.ResponseWriter, r *http.Request) {
+	var (
+		err     error
+		reqInfo = api.GetReqInfo(r.Context())
+	)
+
+	list := &AccessControlPolicy{}
+	if r.ContentLength == 0 {
+		list, err = parseACLHeaders(r)
+		if err != nil {
+			h.logAndSendError(w, "could not parse bucket acl", reqInfo, err)
+			return
+		}
+	} else if err := xml.NewDecoder(r.Body).Decode(list); err != nil {
+		h.logAndSendError(w, "could not parse bucket acl", reqInfo, err)
+		return
+	}
+
+	list.Resource = reqInfo.BucketName
+	list.IsBucket = true
+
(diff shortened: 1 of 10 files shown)


masterSplinter01:
Could you please add some examples of usage of supported methods?

KirillovDenis:
## PutBucketPolicy
`aws s3api put-bucket-policy --endpoint-url http://127.0.0.1:8084 --bucket test-acl8  --policy file://policy.json`
Policy example to allow `ListObjects/v2` to all users:
```
{
    "Version": "2012-10-17",
    "Id": "ExamplePolicy01",
    "Statement": [
        {
            "Sid": "ExampleStatement01",
            "Effect": "Allow",
            "Principal": {
                "AWS": "*"
            },
            "Action": [
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::test-acl8"
            ]
        }
    ]
}
```
Other examples of policy [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-bucket-policies.html).
**Note:** currently aws conditions and wildcard are not supported in resources.

## PutBucketAcl
`aws s3api put-bucket-acl --endpoint-url http://127.0.0.1:8084 --bucket test-acl8 --access-control-policy file://acl.json`
Acl example to allow all operations to specific user:
```
{
  "Owner": {
    "DisplayName": "NbUgTSFvPmsRxmGeWpuuGeJUoRoi6PErcM",
    "ID": "021dc56fc6d81d581ae7605a8e00e0e0bab6cbad566a924a527339475a97a8e38e"
  },
  "Grants": [
    {
      "Grantee": {
        "ID": "021dc56fc6d81d581ae7605a8e00e0e0bab6cbad566a924a527339475a97a8e38e",
        "Type": "CanonicalUser"
      },
      "Permission": "FULL_CONTROL"
    }
  ]
}
```
More information [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html)
**Note:** currently only `CanonicalUser` with hex encoded public key and `Group` with `http://acs.amazonaws.com/groups/global/AllUsers` value are supported.

## PutObjectAcl
Similar to `PutBucketAcl`
`aws s3api put-object-acl --endpoint-url http://127.0.0.1:8084 --bucket test-acl8 --key obj1 --access-control-policy file://acl.json

Also `--acl` and  `--grant-*` parameters are supported.

api/handler/acl.go
```diff
@@ -0,0 +1,1185 @@
...
+		h.writeError(w, r, "expected owner doesn't match", rid, bkt, err)
+		return
+	}
+
+	if err = api.EncodeToResponse(w, h.encodeBucketACL(bucketACL)); err != nil {
+		h.registerAndSendError(w, r, err, "something went wrong")
+	}
+}
+
+func (h *handler) writeError(w http.ResponseWriter, r *http.Request, msg, rid, bkt string, err error) {
```
masterSplinter01, line 118: This function duplicates `func (h *handler) registerAndSendError(w http.ResponseWriter, r *http.Request, err error, logText string)`. 

We need the function like `writeError` everywhere in handler.  Maybe we should create a separate file (util.go, for example) and put functions like this into this file. If you don't like a name `registerAndSendError` you can use `writeError`, but could you please replace all usages of `registerAndSendError` by `WriteError`?

masterSplinter01: Also as I see we use `rid` (at this moment) only in logs messages. Maybe we can extract `rid` from `http.Request.Context()` inside `writeError` and don't pass `rid` together with `http.Request`?

api/handler/response.go
```diff
@@ -52,6 +52,42 @@ type Bucket struct {
 	CreationDate string // time string of format "2006-01-02T15:04:05.000Z"
 }
 
+// AccessControlPolicy contains ACL.
+type AccessControlPolicy struct {
+	XMLName xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ AccessControlPolicy" json:"-"`
+	Owner   Owner
+	//AccessControlList AccessControlList `xml:"AccessControlList"`
```
masterSplinter01, line 59: What does this comment mean?

api/handler/acl.go
KirillovDenis: I think we can create a separate issue for this (refactor so that the `registerAndSendError` function takes `api.ReqInfo` to log at least method and bucket name). For now, let's use the `registerAndSendError` function everywhere.

api/handler/response.go
KirillovDenis: Fixed

api/handler/acl.go
```diff
@@ -0,0 +1,1185 @@
...
+		rid = api.GetRequestID(r.Context())
+	)
+
+	bucketACL, err := h.obj.GetBucketACL(r.Context(), bkt)
+	if err != nil {
+		h.writeError(w, r, "could not fetch bucket acl", rid, bkt, err)
+		return
+	}
+
+	if err := checkOwner(bucketACL.Info, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
```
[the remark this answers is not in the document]
KirillovDenis: I'm not sure if this is necessary.

api/layer/layer.go
```diff
@@ -193,6 +203,38 @@ func (n *layer) GetBucketInfo(ctx context.Context, name string) (*BucketInfo, er
...
+}
+
+// PutBucketACL put bucket acl by name.
+func (n *layer) PutBucketACL(ctx context.Context, param *PutBucketACLParams) error {
+	inf, err := n.GetBucketInfo(ctx, param.Name)
+	if err != nil {
+		return err
+	}
+
+	if err = n.setContainerEACLTable(ctx, inf.CID, param.EACL); err != nil {
```
alexvanin, line 231:
```suggestion
	return n.setContainerEACLTable(ctx, inf.CID, param.EACL)
```

api/layer/container.go
```diff
@@ -130,19 +135,15 @@ func (n *layer) createContainer(ctx context.Context, p *CreateBucketParams) (*ci
 		return nil, err
 	}
 
-	if err := n.setContainerEACL(ctx, cid, p.BoxData.Gate.GateKey); err != nil {
+	if err := n.setContainerEACLTable(ctx, cid, p.EACL); err != nil {
```
alexvanin, line 138: Did I understand correctly, that default eACL will be set up on higher lever (with the help of `addPredefinedACP`) and then passed as `p.EACL`?

api/layer/container.go
```diff
@@ -154,25 +155,12 @@ func (n *layer) setContainerEACL(ctx context.Context, cid *cid.ID, gateKey *keys
...
-		eacl.AddFormedTarget(record, eacl.RoleUser, gateKey)
-		table.AddRecord(record)
-
-		record2 := eacl.NewRecord()
-		record2.SetOperation(op)
-		record2.SetAction(eacl.ActionDeny)
-		eacl.AddFormedTarget(record2, eacl.RoleOthers)
-		table.AddRecord(record2)
+func (n *layer) GetContainerEACL(ctx context.Context, cid *cid.ID) (*eacl.Table, error) {
+	signedEacl, err := n.pool.GetEACL(ctx, cid)
```
alexvanin, line 159: Can you create an issue on caching of EACL values? I am not sure if we need that, because it is already cached in Storage node, thus we can discuss it there.

api/handler/acl.go
```diff
@@ -0,0 +1,1149 @@
...
+		eacl.OperationSearch, eacl.OperationRange, eacl.OperationRangeHash}
+	fullOps = []eacl.Operation{eacl.OperationGet, eacl.OperationHead, eacl.OperationPut,
+		eacl.OperationDelete, eacl.OperationSearch, eacl.OperationRange, eacl.OperationRangeHash}
+)
+
+var actionToOpMap = map[string][]eacl.Operation{
+	"s3:DeleteObject": {eacl.OperationDelete},
+	"s3:GetObject":    readOps,
+	"s3:PutObject":    {eacl.OperationPut},
+	"s3:ListBucket":   readOps,
```
alexvanin, line 31: Is this mapping suppose to be _precise_? For example to control `s3:GetObject` you don't need to control `eacl.OperationSearch`. For `s3:ListBucket` it seems that `eacl.OperationSearch` should be enough (I don't see how one can abuse listing with HEAD, GET, etc. requests).

Larger scopes are not bad, but maybe they can interfere each other (restrict get + allow search = ?).

api/handler/acl.go
```diff
@@ -0,0 +1,1149 @@
...
+		}
+	case aclWrite:
+		if isBucket {
+			res = []string{"s3:PutObject", "s3:DeleteObject"}
+		}
+	case aclFullControl:
+		if isBucket {
+			res = []string{"s3:ListBucket", "s3:ListBucketVersions", "s3:ListBucketMultipartUploads", "s3:PutObject", "s3:DeleteObject"}
+		} else {
+			res = []string{"s3:GetObject", "s3:GetObjectVersion"}
```
alexvanin, line 922: Define constants for those `s3:*` strings.

alexvanin (approved):
Overall LGTM. Tests are very nice, keep adding them for more complex use cases later! :+1: 

One thing I was curios is a Storage node behavior when it parses empty filter list in eACL record. Behavior is not defined in [NeoFS API](https://github.com/nspcc-dev/neofs-api/blob/7f67c378a74adc0f9d47b6c604822e95c79d1f5c/acl/types.proto#L143-L144). According to implementation, Storage node should match that record on any object and process Action (Allow\Deny). I didn't test it though.

If S3 gate heavily depends on this behavior for bucket ACL implementation, we should mention it in explicitly in the spec. 

/cc @realloc

realloc:
> One thing I was curios is a Storage node behavior when it parses empty filter list in eACL record. Behavior is not defined in [NeoFS API](https://github.com/nspcc-dev/neofs-api/blob/7f67c378a74adc0f9d47b6c604822e95c79d1f5c/acl/types.proto#L143-L144). According to implementation, Storage node should match that record on any object and process Action (Allow\Deny). I didn't test it though.
> 
> If S3 gate heavily depends on this behavior for bucket ACL implementation, we should mention it in explicitly in the spec.

That's the behaviour described in PR for ACL section of NeoFS Tech Spec. Empty filter matches anything.

api/layer/container.go
KirillovDenis: Yes. We set up default eACL (initialy as `AccessControlPolicy`) in `parseACLHeaders` function. And if `X-Amz-Acl` is provided we complement eACL in `addPredefinedACP` function.

api/layer/container.go
KirillovDenis: #202

api/handler/acl.go
KirillovDenis: > Is this mapping suppose to be _precise_? For example to control `s3:GetObject` you don't need to control `eacl.OperationSearch`. For `s3:ListBucket` it seems that `eacl.OperationSearch` should be enough (I don't see how one can abuse listing with HEAD, GET, etc. requests).
> 
> Larger scopes are not bad, but maybe they can interfere each other (restrict get + allow listing = ?).

I think we can reconsider this mapping when we add cache to s3-gw. For example at the moment `GetObject` has to perform `search`, `get`, `head`, `range`(sometimes).

alexvanin (approved):
Don't forget to squash \ fixup commits before merge.

masterSplinter01 (approved):

PR #181, opened 2021-07-29
Add cache to ListObjectsV1

### LINKED ISSUE #179: Implement cache for ListObjectsV1
After #175

masterSplinter01:
Closes #179

### DIFF
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -13,7 +13,6 @@ import (
 	cid "github.com/nspcc-dev/neofs-api-go/pkg/container/id"
 	"github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-s3-gw/api"
-	"github.com/nspcc-dev/neofs-s3-gw/creds/accessbox"
 	"go.uber.org/zap"
 )
 
@@ -56,10 +55,9 @@ type (
 	}
 
 	allObjectParams struct {
-		Bucket     *BucketInfo
-		Delimiter  string
-		Prefix     string
-		StartAfter string
+		Bucket    *BucketInfo
+		Delimiter string
+		Prefix    string
 	}
 )
 
@@ -209,45 +207,34 @@ func (n *layer) objectDelete(ctx context.Context, address *object.Address) error
 // ListObjectsV1 returns objects in a bucket for requests of Version 1.
 func (n *layer) ListObjectsV1(ctx context.Context, p *ListObjectsParamsV1) (*ListObjectsInfoV1, error) {
 	var (
-		err    error
-		result ListObjectsInfoV1
-		bkt    *BucketInfo
+		err        error
+		result     ListObjectsInfoV1
+		allObjects []*ObjectInfo
 	)
 
 	if p.MaxKeys == 0 {
 		return &result, nil
 	}
 
-	if bkt, err = n.GetBucketInfo(ctx, p.Bucket); err != nil {
+	if allObjects, err = n.listAllObjects(ctx, p.ListObjectsParamsCommon); err != nil {
 		return nil, err
 	}
 
-	allObjects, err := n.listSortedAllObjects(ctx, allObjectParams{
-		Bucket:     bkt,
-		Prefix:     p.Prefix,
-		Delimiter:  p.Delimiter,
-		StartAfter: p.Marker,
-	})
-	if err != nil {
-		return nil, err
+	if len(allObjects) == 0 {
+		return &result, nil
+	}
+
+	if p.Marker != "" {
+		allObjects = trimAfterObjectName(p.Marker, allObjects)
 	}
 
 	if len(allObjects) > p.MaxKeys {
 		result.IsTruncated = true
-
-		nextObject := allObjects[p.MaxKeys-1]
-		result.NextMarker = nextObject.Name
-
 		allObjects = allObjects[:p.MaxKeys]
+		result.NextMarker = allObjects[len(allObjects)-1].Name
 	}
 
-	for _, ov := range allObjects {
-		if ov.isDir {
-			result.Prefixes = append(result.Prefixes, ov.Name)
-		} else {
-			result.Objects = append(result.Objects, ov)
-		}
-	}
+	result.Prefixes, result.Objects = triageObjects(allObjects)
 
 	return &result, nil
 }
@@ -258,67 +245,40 @@ func (n *layer) ListObjectsV2(ctx context.Context, p *ListObjectsParamsV2) (*Lis
 		err        error
 		result     ListObjectsInfoV2
 		allObjects []*ObjectInfo
-		bkt        *BucketInfo
-		cacheKey   string
-		box        *accessbox.Box
 	)
 
 	if p.MaxKeys == 0 {
 		return &result, nil
 	}
 
-	if bkt, err = n.GetBucketInfo(ctx, p.Bucket); err != nil {
+	if allObjects, err = n.listAllObjects(ctx, p.ListObjectsParamsCommon); err != nil {
 		return nil, err
 	}
 
-	if box, err = GetBoxData(ctx); err != nil {
-		return nil, err
+	if len(allObjects) == 0 {
+		return &result, nil
 	}
-
-	cacheKey = createKey(box.Gate.AccessKey, bkt.CID)
 
 	if p.ContinuationToken != "" {
-		allObjects = n.cache.Get(p.ContinuationToken, cacheKey)
-		allObjects = trimStartAfter(p.StartAfter, allObjects)
+		allObjects = trimAfterObjectID(p.ContinuationToken, allObjects)
 	}
 
-	if allObjects == nil {
-		allObjects, err = n.listSortedAllObjects(ctx, allObjectParams{
-			Bucket:     bkt,
-			Prefix:     p.Prefix,
-			Delimiter:  p.Delimiter,
-			StartAfter: p.StartAfter,
-		})
-		if err != nil {
-			return nil, err
-		}
-
-		if p.ContinuationToken != "" {
-			allObjects = trimAfterObjectID(p.ContinuationToken, allObjects)
-		}
+	if p.StartAfter != "" {
+		allObjects = trimAfterObjectName(p.StartAfter, allObjects)
 	}
 
 	if len(allObjects) > p.MaxKeys {
 		result.IsTruncated = true
-
-		restObjects := allObjects[p.MaxKeys:]
-		n.cache.Put(cacheKey, restObjects)
-		result.NextContinuationToken = restObjects[0].id.String()
-
 		allObjects = allObjects[:p.MaxKeys]
+		result.NextContinuationToken = allObjects[len(allObjects)-1].id.String()
 	}
 
-	for _, ov := range allObjects {
-		if ov.isDir {
-			result.Prefixes = append(result.Prefixes, ov.Name)
-		} else {
-			result.Objects = append(result.Objects, ov)
-		}
-	}
+	result.Prefixes, result.Objects = triageObjects(allObjects)
+
 	return &result, nil
 }
 
-func (n *layer) listSortedAllObjects(ctx context.Context, p allObjectParams) ([]*ObjectInfo, error) {
+func (n *layer) listSortedObjectsFromNeoFS(ctx context.Context, p allObjectParams) ([]*ObjectInfo, error) {
 	var (
 		err       error
 		ids       []*object.ID
@@ -346,9 +306,6 @@ func (n *layer) listSortedAllObjects(ctx context.Context, p allObjectParams) ([]
 			if _, ok := uniqNames[oi.Name]; ok {
 				continue
 			}
-			if len(p.StartAfter) > 0 && oi.Name <= p.StartAfter {
-				continue
-			}
 
 			uniqNames[oi.Name] = oi.isDir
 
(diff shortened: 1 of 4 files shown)


api/layer/object_cache.go
```diff
@@ -74,6 +86,31 @@ func (l *listObjectsCache) Put(key string, objects []*ObjectInfo) {
...
+	box, err := GetBoxData(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	return box.Gate.AccessKey + cid.String(), nil
+}
+
+func trimAfterObjectName(startAfter string, objects []*ObjectInfo) []*ObjectInfo {
+	if objects != nil && len(startAfter) != 0 && objects[0].Name <= startAfter {
```
KirillovDenis, line 99: If objects is `[]*ObjectInfo{}` panic will occur.

api/layer/object.go
```diff
@@ -363,22 +356,21 @@ func (n *layer) listSortedAllObjects(ctx context.Context, p allObjectParams) ([]
...
+	n.cache.Put(cacheKey, restObjects)
+
+	return objects[:maxKeys]
 }
 
-func trimAfterObjectID(id string, objects []*ObjectInfo) []*ObjectInfo {
-	for i, obj := range objects {
-		if obj.ID().String() == id {
-			return objects[i:]
+func sortObjects(allObjects []*ObjectInfo) (prefixes []string, objects []*ObjectInfo) {
```
KirillovDenis, line 366: A bit confusing name. Maybe is it better to use the word "filter" or "sort"?

api/layer/object_cache.go
masterSplinter01: Fixed

api/layer/object.go
masterSplinter01: I meaned sorting objects into separate lists (objects and prefixes), but triage sounds better, renamed.

api/layer/object.go
```diff
@@ -222,33 +223,37 @@ func (n *layer) ListObjectsV1(ctx context.Context, p *ListObjectsParamsV1) (*Lis
...
+			Bucket:     bkt,
+			Prefix:     p.Prefix,
+			Delimiter:  p.Delimiter,
+			StartAfter: p.Marker,
+		})
+		if err != nil {
+			return nil, err
+		}
+		if p.Marker != "" {
+			allObjects = trimAfterObjectName(p.Marker, allObjects)
```
KirillovDenis, line 245: We don't need to trim because `listSortedAllObjects` did it.

masterSplinter01: Hm, really, removed this

api/layer/object_cache.go
```diff
@@ -74,6 +86,31 @@ func (l *listObjectsCache) Put(key string, objects []*ObjectInfo) {
...
+
+func trimAfterObjectName(startAfter string, objects []*ObjectInfo) []*ObjectInfo {
+	if len(startAfter) != 0 && len(objects) != 0 && objects[0].Name <= startAfter {
+		for i := range objects {
+			if objects[i].Name > startAfter {
+				return objects[i:]
+			}
+		}
+	}
+	return objects
```
KirillovDenis, line 106: Why we return full list rather than empty if all key before `startAfter`?
For example if we have list `[]string{"a", "b"}` and pass `startAfter=c` I expect the result will be `[]string{}`

masterSplinter01: You're right, writing unit tests...

KirillovDenis: I think we can simplify this function to:
```
func trimAfterObjectName(startAfter string, objects []*ObjectInfo) []*ObjectInfo {
	for i := range objects {
		if objects[i].Name > startAfter {
			return objects[i:]
		}
	}
	return []*ObjectInfo{}
}
```

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	t.Run("existing and second-to-last id", func(t *testing.T) {
+		actual := trimAfterObjectID(ids[len(ids)-2].String(), objects)
+		require.Equal(t, objects[len(ids)-1:], actual)
+	})
+
+	t.Run("non-existing id", func(t *testing.T) {
+		actual := trimAfterObjectID("z", objects)
+		require.Empty(t, actual)
+		require.NotNil(t, actual)
```
KirillovDenis, line 94: I suppose we can change:
```
require.Empty(t, actual)
require.NotNil(t, actual)
```
to 
```
require.Equal(t, emptyRes, actual)
```
if define `emptyRes` like `emptyRes := []*ObjectInfo{}`

api/layer/object_cache.go
masterSplinter01: Yep, I've made this changes but have not pushed yet. I will re-request review when I finish.

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,170 @@
...
+	})
+
+	t.Run("nil objects", func(t *testing.T) {
+		actual := trimAfterObjectName(names[0], nil)
+		require.Nil(t, actual)
+	})
+
+	t.Run("empty startafter", func(t *testing.T) {
+		actual := trimAfterObjectName("", objects)
+		require.Equal(t, actual, objects)
```
KirillovDenis, line 73: The second argument must be `expected` (not actual)

api/layer/object_cache.go
```diff
@@ -74,6 +86,29 @@ func (l *listObjectsCache) Put(key string, objects []*ObjectInfo) {
...
+	for i := range objects {
+		if objects[i].Name > startAfter {
+			return objects[i:]
+		}
+	}
+	return nil
+}
+
+func trimAfterObjectID(id string, objects []*ObjectInfo) []*ObjectInfo {
+	for i, obj := range objects[:len(objects)-1] {
```
KirillovDenis, line 108: Should we iterate over all objects? We do neofs `objectSearch` If continuation token is the last object id but I suppose we can return empty list to client.

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	t.Run("GetByContinuationToken for invalid token", func(t *testing.T) {
+		cache := newListObjectsCache()
+		token := randID(t).String()
+		cache.Put(key, objects)
+		actual := cache.GetByContinuationToken(token, key)
+		require.Nil(t, actual)
+	})
+
+	t.Run("GetByMarker for valid marker", func(t *testing.T) {
```
KirillovDenis, line 156: This test failed because objects must be sorted by name to use marker.

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,170 @@
...
+		cache := newListObjectsCache()
+		cache.Put(key, objects)
+		actual := cache.GetByMarker(objects[0].id.String(), key)
+		require.Equal(t, objects[1:], actual)
+	})
+
+	t.Run("GetByMarker for invalid token", func(t *testing.T) {
+		cache := newListObjectsCache()
+		marker := "asdf"
```
KirillovDenis, line 165: Marker should be `zzz` for example because we have no guarantees that this key (`asdf`) will not be between existing object names

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,170 @@
...
+		id := randID(t)
+		objects = append(objects, &ObjectInfo{id: id, Name: id.String()})
+	}
+
+	t.Run("lifetime", func(t *testing.T) {
+		cache := newListObjectsCache()
+		cache.Put(key, objects)
+
+		time.Sleep(defaultCacheLifetime - 10)
+		_, ok := cache.caches[key]
```
KirillovDenis, line 133: We should not use cache directly. Better to use, for example, `cache.GetByContinuationToken(objects[0].ID().String(), key)`.

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,170 @@
...
+	for i := 0; i < cacheSize; i++ {
+		id := randID(t)
+		objects = append(objects, &ObjectInfo{id: id, Name: id.String()})
+	}
+
+	t.Run("lifetime", func(t *testing.T) {
+		cache := newListObjectsCache()
+		cache.Put(key, objects)
+
+		time.Sleep(defaultCacheLifetime - 10)
```
KirillovDenis, line 132: I suppose we should use `defaultCacheLifetime - 10 * time.Second`

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,170 @@
...
+		key       = "key"
+	)
+
+	for i := 0; i < cacheSize; i++ {
+		id := randID(t)
+		objects = append(objects, &ObjectInfo{id: id, Name: id.String()})
+	}
+
+	t.Run("lifetime", func(t *testing.T) {
+		cache := newListObjectsCache()
```
KirillovDenis, line 129: Maybe add an optional parameter to define the cache lifetime? 60 seconds of waiting for a test is too long for me.

api/layer/object_cache_test.go
masterSplinter01: Fixed

api/layer/object_cache.go
masterSplinter01: It  seems like a minor optimization, but added this.

api/layer/object_cache_test.go
masterSplinter01: Removed this test, it's senseless

api/layer/object_cache_test.go
masterSplinter01: I'm not sure because here we test only that cache keeps an entry for certain time. Methods GetBy... are tested below.  
At first, if we add GetBy... here,  the test will include another test for GetBy..., what seems redundant a bit.  
At second,  if I add GetBy... here and this test fail, it will not be clear  enough what is broken: cache's timer or GetBy...

api/layer/object_cache_test.go
masterSplinter01: I've fixed it earlier

api/layer/object_cache.go
KirillovDenis: If we want to avoid redundant `search` we should change:
```
func trimAfterObjectID(id string, objects []*ObjectInfo) []*ObjectInfo {
	if len(objects) != 0 && objects[len(objects)-1].id.String() == id {
		return nil
	}
       ...
}
``` 
to 
```
func trimAfterObjectID(id string, objects []*ObjectInfo) []*ObjectInfo {
	if len(objects) != 0 && objects[len(objects)-1].id.String() == id {
		return []*ObjectInfo{}
	}
       ...
}
``` 
because there is check `if allObjects == nil ` in `ListObjectsV2`

api/layer/object_cache_test.go
KirillovDenis: Otherwise tests run with a race detector will find a race.

api/layer/object_cache_test.go
masterSplinter01: We talked about this with @alexvanin, but we decided to leave cache lifetime hardcoded because there is no such need to add this as a parameter. 

But you're right this test takes too long. Maybe I'd be better to separate this test somehow?

KirillovDenis: We can extract this functionality to separate function and test it. For example:
```
func (l *listObjectsCache) scheduleClearing(key string, d time.Duration) {
	time.AfterFunc(d, func() {
		l.mtx.Lock()
		delete(l.caches, key)
		l.mtx.Unlock()
	})
}
```

api/layer/object.go
```diff
@@ -363,22 +353,47 @@ func (n *layer) listSortedAllObjects(ctx context.Context, p allObjectParams) ([]
...
+			return objects[i+1:]
+		}
+	}
+
+	return nil
+}
+
+func (n *layer) trimAndCacheRestObjects(cacheKey string, maxKeys int, objects []*ObjectInfo) []*ObjectInfo {
+	restObjects := objects[maxKeys-1:]
+	n.cache.Put(cacheKey, restObjects)
```
alexvanin, line 384: I am not sure about cashing trimmed result. Consider bucket `XXX` with 100 keys. There are two consecutive requests from the same user:

1. `ListObjectV1` of bucket `XXX` with `MaxKeys` limit 10
2. `ListObjectV2` of bucket `XXX` with `MaxKeys` limit 100

First request will cache remaining 90 keys after the trim.
Second request will return 90 keys from the cache, instead of 100, which is incorrect.

Am I missing something?

api/layer/object_cache_test.go
alexvanin: > but we decided to leave cache lifetime hardcoded because there is no such need to add this as a parameter.

60 seconds can be hardcoded in terms of "not configurable from the S3 Gateway configuration". However the program component itself may be configured. Let's do that and make these tests pass faster.

api/layer/object_cache.go
masterSplinter01: The idea was to try update data if we don't receive any from cache, but in case of ContinuationToken it's poinless maybe, fixed.

api/layer/object.go
masterSplinter01: I thought we search for a cache entry when a user sets ContinuationToken/Marker in a request, so the user should get a response from cache if he received the token from the previous response. That's why cache doesn't need to keep objects before ContinuationToken/Marker (it makes search by the token a little faster).
 
But I didn't thought about different versions compatibility. I have some suspicions this implementation will not work in this case correctly,  because, yes, you're right,  ListObjectV1 will put a "broken" (i mean filtered with old markers)  cache, which is useless for another version. 

There is two possible ways: 
1. Somehow separate entries in a map of cache for different versions
2. Create a common cache with unfiltered objects for all versions, get back workers and remove afterfuncs, because with afterfunc the cache entry lives only for 1 minute after creating of the entry and we can't prolong existence, but it'd be better to add a way to update timer for cache entry, if we want to read this cache from different ListObjects versions.

api/layer/object_cache.go
masterSplinter01: Okay, fixed

api/layer/object_cache_test.go
masterSplinter01: Fixed

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"crypto/sha256"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/nspcc-dev/neofs-api-go/pkg/object"
+	"github.com/stretchr/testify/require"
+)
+
+var testingCacheLifetime = 15 * time.Second
```
KirillovDenis, line 14: I would reduce the wait time to 5 seconds.

api/layer/object_cache_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	sort.Slice(objects, func(i, j int) bool {
+		return objects[i].Name < objects[j].Name
+	})
+
+	t.Run("lifetime", func(t *testing.T) {
+		cache := newListObjectsCache(testingCacheLifetime)
+		cache.Put(key, objects)
+
+		time.Sleep(cache.cacheLifetime / 2)
```
KirillovDenis, line 139: I think we can drop `sleep` function if we use something like this:
```
...
condition := func() bool {
	return cache.GetByContinuationToken(objects[0].ID().String(), key) == nil
}
require.Never(t, condition, cache.cacheLifetime-time.Second, time.Second)
require.Eventually(t, condition, cache.cacheLifetime+time.Second, time.Second)
...
```

masterSplinter01: Thanks, replaced

KirillovDenis:
I just noticed that `defaultCacheLifetime` is not a constant. I think it should be.

api/layer/object.go
alexvanin: > 1. Somehow separate entries in a map of cache for different versions

In worst case scenario we will have two different caches with the same data, sounds bad.

> 2. Create a common cache with unfiltered objects for all versions,

I would go for it.

>  get back workers and remove afterfuncs, because with afterfunc the cache entry lives only for 1 minute after creating of the entry and we can't prolong existence, but it'd be better to add a way to update timer for cache entry, if we want to read this cache from different ListObjects versions.

Cache invalidation is one of the hardest problems in CS. With proposed prolongation there is a chance we will never invalidate the cache at all. I propose to keep things as they are right now and go back to the cache prolongation issues when we will be tuning performance of the S3 gate.

api/layer/object_cache.go
```diff
@@ -1,79 +1,96 @@
...
 
-// ObjectsListV2Cache provides interface for cache of ListObjectsV2 in a layer struct.
+// ObjectsListCache provides interface for cache of ListObjectsV2 in a layer struct.
 type (
-	ObjectsListV2Cache interface {
-		Get(token string, key string) []*ObjectInfo
-		Put(key string, objects []*ObjectInfo)
+	ObjectsListCache interface {
+		Get(key cacheOptions) []*ObjectInfo
+		Put(p cacheOptions, objects []*ObjectInfo)
```
KirillovDenis, line 30: Why `key` replaced by `p`?

api/layer/object_cache.go
```diff
@@ -1,79 +1,96 @@
...
 	}
 }
 
-func (l *listObjectsCache) Get(token, key string) []*ObjectInfo {
+func (l *listObjectsCache) Get(key cacheOptions) []*ObjectInfo {
 	l.mtx.RLock()
 	defer l.mtx.RUnlock()
 	if val, ok := l.caches[key]; ok {
-		return trimAfterObjectID(token, val.list)
+		return val.list
```
KirillovDenis, line 63: Why we refused `trimAfterObjectID`?

KirillovDenis:
Lint error:
```
api/layer/object_cache_test.go:160  gofmt  File is not `gofmt`-ed with `-s`
```

api/layer/object_cache.go
masterSplinter01: At first, earlier cache was connected closely with token/marker.  In the previous version we read cache only if we received token/marker from the request. 
Now we read cache always,  both versions can use the same cache entry with one big unfiltered list of objects.
That's why specific functions for each version become more senseless.

At second, in the previous version we did the following actions:
1. Try to get cache: find the entry, filter by token/marker
2. If we can't find the entry we ask NeoFS for the list and filter by token/marker
We made filtering in both cases.

At third, one common function for both versions allowed us to create a separate function `listAllObjects` and use it in both ListObjects, what make these functions shorter.

That's why I decided to  take filtering by token/marker outside and replace `GetByContinuationToken` and `GetByMarker` by `Get` which returns all of objects.

alexvanin (approved):

api/layer/object_cache.go
masterSplinter01: Renamed

KirillovDenis (approved):

KirillovDenis:
I think you need to squash commits

PR #182, opened 2021-07-29
Add check of continuation token

### LINKED ISSUE #180: Response with an error if we received an empty value of ContinuationToken
Now the gateway doesn't check is ContinuationToken empty or not.  AWS and Minio return an error in such case: 

Minio
```
 $ aws s3api list-objects-v2 --bucket start --endpoint-ur http://127.0.0.1:9000 --continuation-token ""

An error occurred (InvalidArgument) when calling the ListObjectsV2 operation: The continuation token provided is incorrect

$ aws s3api list-objects-v2 --bucket start --endpoint-ur http://127.0.0.1:9000 --continuation-token ''

An error occurred (InvalidArgument) when calling the ListObjectsV2 operation: The continuation token provided is incorrect
```
AWS: 
```
$ aws --endpoint-url https://s3.eu-north-1.amazonaws.com s3api list-objects-v2  --bucket avbexamplebucket --continuation-token ''

An error occurred (InvalidArgument) when calling the ListObjectsV2 operation: The continuation token provided is incorrect
```

Maybe we should implement the same behavior in the gateway.

masterSplinter01:
Closes #180

### DIFF
--- a/api/handler/object_list.go
+++ b/api/handler/object_list.go
@@ -37,9 +37,8 @@ func encodeV1(p *layer.ListObjectsParamsV1, list *layer.ListObjectsInfoV1) *List
 		Prefix:       p.Prefix,
 		MaxKeys:      p.MaxKeys,
 		Delimiter:    p.Delimiter,
-
-		IsTruncated: list.IsTruncated,
-		NextMarker:  list.NextMarker,
+		IsTruncated:  list.IsTruncated,
+		NextMarker:   list.NextMarker,
 	}
 
 	res.CommonPrefixes = fillPrefixes(list.Prefixes, p.Encode)
@@ -71,16 +70,14 @@ func (h *handler) ListObjectsV2Handler(w http.ResponseWriter, r *http.Request) {
 
 func encodeV2(p *layer.ListObjectsParamsV2, list *layer.ListObjectsInfoV2) *ListObjectsV2Response {
 	res := &ListObjectsV2Response{
-		Name:         p.Bucket,
-		EncodingType: p.Encode,
-		Prefix:       s3PathEncode(p.Prefix, p.Encode),
-		KeyCount:     len(list.Objects) + len(list.Prefixes),
-		MaxKeys:      p.MaxKeys,
-		Delimiter:    s3PathEncode(p.Delimiter, p.Encode),
-		StartAfter:   s3PathEncode(p.StartAfter, p.Encode),
-
-		IsTruncated: list.IsTruncated,
-
+		Name:                  p.Bucket,
+		EncodingType:          p.Encode,
+		Prefix:                s3PathEncode(p.Prefix, p.Encode),
+		KeyCount:              len(list.Objects) + len(list.Prefixes),
+		MaxKeys:               p.MaxKeys,
+		Delimiter:             s3PathEncode(p.Delimiter, p.Encode),
+		StartAfter:            s3PathEncode(p.StartAfter, p.Encode),
+		IsTruncated:           list.IsTruncated,
 		ContinuationToken:     p.ContinuationToken,
 		NextContinuationToken: list.NextContinuationToken,
 	}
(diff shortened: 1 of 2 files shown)


api/handler/object_list.go
```diff
@@ -121,32 +121,39 @@ func parseListObjectsArgsV2(r *http.Request) (*layer.ListObjectsParamsV2, error)
 	}
 	res.ListObjectsParamsCommon = *common
 
-	res.ContinuationToken = r.URL.Query().Get("continuation-token")
-	res.StartAfter = r.URL.Query().Get("start-after")
-	res.FetchOwner, _ = strconv.ParseBool(r.URL.Query().Get("fetch-owner"))
+	if val, ok := queryValues["continuation-token"]; ok {
```
KirillovDenis, line 124: Maybe we should use `queryValues[""]` everywhere in this function?

api/handler/object_list.go
```diff
@@ -121,32 +121,39 @@ func parseListObjectsArgsV2(r *http.Request) (*layer.ListObjectsParamsV2, error)
...
 	return &res, nil
 }
 
 func parseListObjectArgs(r *http.Request) (*layer.ListObjectsParamsCommon, error) {
 	var (
-		err error
-		res layer.ListObjectsParamsCommon
+		err         error
+		res         layer.ListObjectsParamsCommon
+		queryValues = r.URL.Query()
```
KirillovDenis, line 140: I'm not sure if we need to use it here unless we are interested in having a key in the queries (The same for `parseListObjectVersionsRequest`).

api/handler/object_list.go
masterSplinter01: I'm not sure I understand you correctly, but where should I use queryValues[""] in the function. I replaced parsing everywhere.

api/handler/object_list.go
masterSplinter01: But in `parseListObjectsArgs` code looks like:
```
...
		queryValues = r.URL.Query()
	)

	if info := api.GetReqInfo(r.Context()); info != nil {
		res.Bucket = info.BucketName
	}

	res.Delimiter = queryValues.Get("delimiter")
	res.Encode = queryValues.Get("encoding-type")
...
```
`api.GetReqInfo` doesn't throw errors and then in any case we will read keys from queryValues. I'm not sure it's really important to move `queryValues = r.URL.Query()` down. 

In `parseListObjectVersionsRequest` code looks like:
```
...	
queryValues = r.URL.Query()
	)

	if queryValues.Get("max-keys") == "" {
		res.MaxKeys = maxObjectList
...
```

We read keys from queryValues right after parsing.

api/handler/object_list.go
KirillovDenis: I mean we use `queryValues` differently: `queryValues.Get("start-after")` or `queryValues["continuation-token"]`. Question was, should we use the same style?

masterSplinter01: No, we shouldn't. 
At least because `r.Url.Query` returns `Values` type which is an alias of  `map[string][]string`. If we want to use queryValues as a map, we will have to append `[0]` everywhere, I mean `queryValues[""][0]`. I'm not sure it looks good.

KirillovDenis (approved):

alexvanin (approved):
LGTM.

I have an idea, maybe you'll like it. Otherwise just hit merge button.
What if we put this code

```go
	if val, ok := queryValues["continuation-token"]; ok {
		if err := object.NewID().Parse(val[0]); err != nil {
			return nil, api.GetAPIError(api.ErrIncorrectContinuationToken)
		}
		res.ContinuationToken = val[0]
```

into separate func, something like
```go
res.ContinuationToken, err = parseContinuationToken(queryValues)
if err != nil {
    return nil, err
}
```

Then you can write some tests to `parseContinuationToken` with correct and incorrect input values, check expected errors.

KirillovDenis (approved):

PR #208, opened 2021-08-11
Common cache with objects

### LINKED ISSUE #206: Implement common object cache
Common object cache is a TTL LRU cache which contains NeoFS objects headers and their addresses. 

Behaviour of the cache in the following requests: 
1. HeadObject requests: we try to find an object in cache, if the cache contains the object, we return the object from cache. Otherwise, If we don't find an object in the cache we make a request to NeoFS and if we received successful response we put an object into the cache. 
2. GetObject: we make a request to NeoFS anyway but if we received sucsess response we put the object's header into the cache.
3. PutObject -- put an object into the cache
4. DeleteObject -- delete an object from the cache (if exists)

We use [gcache](https://github.com/bluele/gcache) for that.

masterSplinter01:
Closes #206


api/cache/object_cache.go
```diff
@@ -0,0 +1,62 @@
...
+
+const (
+	// DefaultObjectsCacheLifetime is a default lifetime of objects in cache.
+	DefaultObjectsCacheLifetime = time.Minute * 5
+	// DefaultObjectsCacheSize is a default maximum number of objects in cache.
+	DefaultObjectsCacheSize = 1e6
+)
+
+type (
+	// ObjectHeadersCache contains cache with objects, lifetime of cache entries and client to establish connection with NeoFS.
```
KirillovDenis, line 25: > client to establish connection with NeoFS

I don't see any client in `ObjectHeadersCache`

api/cache/object_cache.go
```diff
@@ -0,0 +1,62 @@
...
+// New creates an object of ObjectHeadersCache.
+func New(cacheSize int, lifetime time.Duration) *ObjectHeadersCache {
+	gc := gcache.New(cacheSize).LRU().Build()
+
+	return &ObjectHeadersCache{cache: gc, lifetime: lifetime}
+}
+
+// Get returns cached object.
+func (o *ObjectHeadersCache) Get(address *object.Address) *object.Object {
+	entry, err := o.cache.Get(address.String())
```
KirillovDenis, line 41: Can we just use `address` as key?

KirillovDenis:
I think we should add some basic tests for `ObjectHeadersCache`

api/layer/object.go
```diff
@@ -183,6 +183,11 @@ func (n *layer) objectPut(ctx context.Context, p *PutObjectParams) (*ObjectInfo,
 		return nil, err
 	}
 
+	err = n.objCache.Put(addr, *meta)
+	if err != nil {
```
KirillovDenis, line 187: We can just write:
```
if err = n.objCache.Put(addr, *meta); err != nil {
	n.log.Error("couldn't cache an object", zap.Error(err))
}
```

api/cache/object_cache.go
masterSplinter01: The gcache is implemented as a map. Address is a struct that contains cid []byte and oid []byte. If I understand correctly, we can use structs as keys and compare them if their fields are comparable, but slice, map, and function values are not comparable ([spec](https://golang.org/ref/spec#Comparison_operators)

api/cache/object_cache.go
```diff
@@ -0,0 +1,62 @@
...
+	"time"
+
+	"github.com/bluele/gcache"
+	"github.com/nspcc-dev/neofs-api-go/pkg/object"
+)
+
+// ObjectsCache provides interface for lru cache for objects.
+type ObjectsCache interface {
+	Get(address *object.Address) *object.Object
+	Put(address *object.Address, obj object.Object) error
```
realloc, line 13: `object.Object` should already have CID/OID inside. Why do we need to have the same in `address`?

api/cache/object_cache.go
masterSplinter01: Forgot to update the comment.

api/cache/object_cache.go
```diff
@@ -0,0 +1,62 @@
...
+	Delete(address *object.Address) bool
+}
+
+const (
+	// DefaultObjectsCacheLifetime is a default lifetime of objects in cache.
+	DefaultObjectsCacheLifetime = time.Minute * 5
```
realloc, line 19: Can we have config settings for this?

api/cache/object_cache.go
masterSplinter01: Because when we put an object into the cache we already have parsed object address via:
```
addr := object.NewAddress()
	addr.SetObjectID(oid)
	addr.SetContainerID(bkt.CID)
```

I'm not sure we should do the same inside Put function again and don't use the ready-to-use value.

api/cache/object_cache.go
masterSplinter01: Fixed

cmd/s3-gw/app.go
```diff
@@ -207,3 +210,45 @@ func (a *App) Server(ctx context.Context) {
...
 }
+
+func getCacheOptions(v *viper.Viper, l *zap.Logger) *layer.CacheConfig {
+	cacheCfg := layer.CacheConfig{
+		ListObjectsLifetime: layer.DefaultObjectsListCacheLifetime,
+		Size:                cache.DefaultObjectsCacheSize,
+		Lifetime:            cache.DefaultObjectsCacheLifetime,
+	}
+
+	if v.IsSet(cfgListObjectsCacheLifetime) {
```
KirillovDenis, line 221: I suppose we should use `cfgObjectsCacheLifetime`

cmd/s3-gw/app.go
```diff
@@ -207,3 +210,45 @@ func (a *App) Server(ctx context.Context) {
 
 	close(a.webDone)
 }
+
+func getCacheOptions(v *viper.Viper, l *zap.Logger) *layer.CacheConfig {
+	cacheCfg := layer.CacheConfig{
+		ListObjectsLifetime: layer.DefaultObjectsListCacheLifetime,
```
KirillovDenis, line 216: Maybe move default cache config values to the app-settings.go file?

masterSplinter01: I think it would be better to store consts for cache in files with cache implementation...

KirillovDenis (approved):

PR #225, opened 2021-08-23
[#218] Configurable default policy

### LINKED ISSUE #218: Make defaultPolicy configurable
Now defaultPolicy is "REP 3", that's why creation of buckets without policy raises errors in 1 node - 1 gate deploy.

masterSplinter01:
Closes #218

### DIFF
--- a/cmd/s3-gw/app.go
+++ b/cmd/s3-gw/app.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neofs-node/pkg/policy"
 	"github.com/nspcc-dev/neofs-s3-gw/api"
 	"github.com/nspcc-dev/neofs-s3-gw/api/auth"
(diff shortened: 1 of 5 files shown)


KirillovDenis (approved):

cmd/s3-gw/app.go
```diff
@@ -252,3 +255,22 @@ func getCacheOptions(v *viper.Viper, l *zap.Logger) *layer.CacheConfig {
 	}
 	return &cacheCfg
 }
+
+func getHandlerOptions(v *viper.Viper, l *zap.Logger) *handler.Config {
```
alexvanin, line 259: Why you suddenly name it `Options` instead of `Config`?

alexvanin (approved):

cmd/s3-gw/app.go
masterSplinter01: I thought it sounds better: the function reads options for handler and creates a config object. But if it seems non-logical I can rename it.

alexvanin: It's okay, just wonder.

PR #239, opened 2021-08-30
[#213] Add object acl versioning

### LINKED ISSUE #213: Object ACL versioning
[GetObjectAcl](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAcl.html) and [PutObjectAcl](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectAcl.html) support versioning (acl can be applied to specific version). So current (#176) applying eacl rules by object name looks a little bit odd (if we want to support such acl versioning).

I see the following solutions:
* Use `object id` (s3 version) in eacl rules (need some changes in [neofs-api-go](https://github.com/nspcc-dev/neofs-api-go)). New method to add `object id` filter can be introduced (similar to `AddObjectContainerIDFilter`). The con is we still have limitation for eacl size https://github.com/nspcc-dev/neofs-node/issues/731
* Perhaps we can use solution https://github.com/nspcc-dev/neofs-api/issues/163

KirillovDenis:
closes #213

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -12,6 +12,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neofs-api-go/pkg/acl/eacl"
 	"github.com/nspcc-dev/neofs-api-go/pkg/object"
+	"github.com/nspcc-dev/neofs-api-go/v2/acl"
 	"github.com/nspcc-dev/neofs-s3-gw/api"
 	"github.com/nspcc-dev/neofs-s3-gw/api/cache"
 	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
@@ -69,6 +70,7 @@ type bucketPolicy struct {
 	Version   string      `json:"Version"`
 	ID        string      `json:"Id"`
 	Statement []statement `json:"Statement"`
+	Bucket    string      `json:"-"`
 }
 
 type statement struct {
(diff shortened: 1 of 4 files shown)


api/handler/acl.go
```diff
@@ -34,9 +34,10 @@ var actionToOpMap = map[string][]eacl.Operation{
 }
 
 const (
-	arnAwsPrefix     = "arn:aws:s3:::"
-	allUsersWildcard = "*"
-	allUsersGroup    = "http://acs.amazonaws.com/groups/global/AllUsers"
+	arnAwsPrefix      = "arn:aws:s3:::"
+	allUsersWildcard  = "*"
+	allUsersGroup     = "http://acs.amazonaws.com/groups/global/AllUsers"
+	attributeObjectID = "$Object:objectID"
```
alexvanin, line 40: You can use [const](https://github.com/nspcc-dev/neofs-api-go/blob/6d0b3135ff2337be338a8c5cef5449ac3a06b58e/v2/acl/filters.go#L11) from api-go, even though it is major version dependent.

alexvanin (approved):
Are there any tests in `s3-tests` for that?

KirillovDenis:
> Are there any tests in `s3-tests` for that?

There is `s3tests_boto3.functional.test_s3:test_versioned_object_acl` test but we couldn't use it. This test fails on versioning checking since objects are created in the same epoch and their versions are out of order.

api/handler/acl.go
KirillovDenis: Fixed

PR #244, opened 2021-08-31
ListObjects cache

### LINKED ISSUE #231: Invalidate ListObjects cache in a gate when we put a new object via this gate
In case when a user requests `ListObjects`, then `PutObject`, and then `ListObject`, they will get a response for the second ListObjects from gateway cache. And the response will not have contained the object that was put with `PutObject`until the cache will  not remove the entry.

### LINKED ISSUE #236: Make listObjects cache getting info from Objects cache
Now listObjects cache is independent on object's common cache and keeps ObjectInfo. Now we need to make listObjects keeping only objects ids and getting info from common objects cache otherwise from NeeoFS.

### LINKED ISSUE #240: Use MatchCommonPrefix for listing objects in bucket
With neofs-api-go v2.9.0 NeoFS supports new search prefix `MatchCommonPrefix`. In neofs-api-go v1.29.0 we can use `object.MatchCommonPrefix` constant in the search query. It should be adopted in S3 object listing to reduce amount of `object.Head` operations.

masterSplinter01:
Closes #231 
Closes #236 
Closes #240 

Now ListObjectsCache keeps unsorted objectIDs of all versions of objects in container. Also method and delimiter are removed from key and ListObjectsCache keeps an entry for containerID and prefix.
Added invalidation of cache entries when we put an object.

Also: 
ObjectInfo and BucketInfo moved to a separate file in api/ as basic structs. 
ListObjectsCache moved into cache/ dir from layer/. 
Renamed cache/object_cache.go to cache/objects.go.


api/layer/object.go
```diff
@@ -82,6 +83,9 @@ func (n *layer) objectSearch(ctx context.Context, p *findParams) ([]*object.ID,
 			opts.AddFilter(p.attr, filename, object.MatchStringEqual)
 		}
 	}
+	if p.prefix != "" {
+		opts.AddFilter(object.AttributeFileName, p.prefix, object.MatchCommonPrefix)
```
KirillovDenis, line 87: Maybe we should do `QueryUnescape` for prefix similar to filename?

api/cache/objectslist.go
```diff
@@ -0,0 +1,134 @@
...
+	When we get a request from the user we just try to find the suitable and non-expired cache and then we return
+	the list of objects. Otherwise we send the request to NeoFS.
+*/
+
+// ObjectsListCache provides interface for cache of ListObjectsV2 in a layer struct.
+type (
+	ObjectsListCache interface {
+		Get(key ObjectsListKey) []*object.ID
+		Put(key ObjectsListKey, oids []*object.ID)
+		Update(key ObjectsListKey, oids []*object.ID)
```
KirillovDenis, line 28: I'm not sure we need this function. We can always use `Put` I think

api/cache/objectslist.go
```diff
@@ -0,0 +1,134 @@
...
+		Contains(key ObjectsListKey) bool
```
KirillovDenis, line 29: This functions can be not exported

api/layer/object.go
```diff
@@ -478,20 +484,35 @@ func (n *layer) listSortedObjectsFromNeoFS(ctx context.Context, p allObjectParam
...
-	for _, id := range ids {
-		meta, err := n.objectHead(ctx, bkt.CID, id)
-		if err != nil {
-			n.log.Warn("could not fetch object meta", zap.Error(err))
+
+	for i := 0; i < len(ids); i++ {
+		obj := n.objectFromObjectsCacheOrNeoFS(ctx, bkt.CID, ids[i])
+		if obj == nil {
+			// remove an invalid object's address from list in cache
+			ids = append(ids[:i], ids[i+1:]...)
```
KirillovDenis, line 510: Maybe we can do something like:
```
validIds := make([]*object.ID, 0, len(ids)
for _, id  := range ids {
	obj := n.objectFromObjectsCacheOrNeoFS(ctx, bkt.CID, ids[i])
	if obj == nil {
		continue
	}
        validIds := append(validIds, id)
  //...
}

if len(validIds) != len(ids) {
    // ...
}
```
or we want to save memory?

api/layer/versioning.go
```diff
@@ -145,39 +145,31 @@ func (n *layer) GetBucketVersioning(ctx context.Context, bucketName string) (*Bu
...
-		}
+	sortedNames := make([]string, 0, len(versions))
+	for k := range versions {
+		sortedNames = append(sortedNames, k)
+	}
+	sort.Strings(sortedNames)
 
-		// putting to cache a copy of allObjects because allObjects can be modified further
-		n.listsCache.Put(cacheKey, append([]*ObjectInfo(nil), allObjects...))
+	allObjects = make([]*api.ObjectInfo, 0, p.MaxKeys)
```
KirillovDenis, line 170: Allocation already was made on line 150

api/layer/versioning.go
```diff
@@ -145,39 +145,31 @@ func (n *layer) GetBucketVersioning(ctx context.Context, bucketName string) (*Bu
...
+	)
 
 	bkt, err := n.GetBucketInfo(ctx, p.Bucket)
 	if err != nil {
 		return nil, err
 	}
 
-	cacheKey, err := createKey(bkt.CID, listVersionsMethod, p.Prefix, p.Delimiter)
+	versions, err = n.getAllObjectsVersions(ctx, bkt, p.Prefix, p.Delimiter)
 	if err != nil {
```
KirillovDenis, line 160: Can be:
```
if versions, err = n.getAllObjectsVersions(ctx, bkt, p.Prefix, p.Delimiter); err != nil {
	return nil, err
}
```

api/cache/objectslist.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// DefaultObjectsListCacheLifetime is a default lifetime of entries in cache of ListObjects.
+const DefaultObjectsListCacheLifetime = time.Second * 60
+
+type (
+	// ListObjectsCache contains cache for ListObjects and ListObjectVersions.
+	ListObjectsCache struct {
+		cacheLifetime time.Duration
+		caches        map[ObjectsListKey]objectsListEntry
```
KirillovDenis, line 41: Maybe we should use gcache?

api/layer/object.go
KirillovDenis: Or we can do nothing because

api/cache/objectslist.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// CleanCacheEntriesChangedWithPutObject deletes entries containing specified object.
+func (l *ListObjectsCache) CleanCacheEntriesChangedWithPutObject(objectName string, cid *cid.ID) {
+	cidStr := cid.String()
+	l.mtx.RLock()
+	defer l.mtx.RUnlock()
+	for k := range l.caches {
+		if cidStr == k.cid && strings.HasPrefix(objectName, k.Prefix) {
+			delete(l.caches, k)
```
KirillovDenis, line 121: Should we use `Lock()` (maybe instead of `RLock()`, maybe in addition)?

KirillovDenis: Or we can use [gcache](https://github.com/bluele/gcache) and don't worry about explicit locking

api/layer/object.go
masterSplinter01: Changed implementation, now it marks  addresses to obviously invalid objects.

api/cache/objectslist.go
```diff
@@ -0,0 +1,113 @@
...
+func (l *ListObjectsCache) Put(key ObjectsListKey, oids []*object.ID) error {
+	if len(oids) == 0 {
+		return fmt.Errorf("list is empty, cid: %s, prefix: %s", key.cid, key.prefix)
+	}
+
+	return l.cache.SetWithExpire(key, oids, l.lifetime)
+}
+
+// CleanCacheEntriesChangedWithPutObject deletes entries containing specified object.
+func (l *ListObjectsCache) CleanCacheEntriesChangedWithPutObject(objectName string, cid *cid.ID) {
```
alexvanin, line 91: Testcases for this function will be appreciated.

api/layer/object.go
```diff
@@ -478,20 +486,35 @@ func (n *layer) listSortedObjectsFromNeoFS(ctx context.Context, p allObjectParam
...
+	cacheKey := cache.CreateObjectsListCacheKey(bkt.CID, prefix)
+	ids := n.listsCache.Get(cacheKey)
+
+	if ids == nil {
+		ids, err = n.objectSearch(ctx, &findParams{cid: bkt.CID, prefix: prefix})
+		if err != nil {
+			return nil, err
+		}
+		if err := n.listsCache.Put(cacheKey, ids); err != nil {
+			n.log.Error("couldn't cache an object", zap.Error(err))
```
alexvanin, line 501:
```suggestion
			n.log.Error("couldn't cache object list", zap.Error(err))
```

alexvanin (approved):

api/cache/objectslist.go
```diff
@@ -0,0 +1,113 @@
...
+
+	When we put an object into a container we invalidate entries with prefixes that are prefixes of the object's name.
+*/
+
+type (
+	// ObjectsListCache provides interface for cache of ListObjectsV2 in a layer struct.
+	ObjectsListCache interface {
+		Get(key ObjectsListKey) []*object.ID
+		Put(key ObjectsListKey, oids []*object.ID) error
+		CleanCacheEntriesChangedWithPutObject(objectName string, cid *cid.ID)
```
realloc, line 31: The logic for DELETE will be the same. Should we give this function a shorter and more generic name?

api/cache/objectslist.go
```diff
@@ -0,0 +1,113 @@
...
+	// DefaultObjectsListCacheLifetime is a default lifetime of entries in cache of ListObjects.
+	DefaultObjectsListCacheLifetime = time.Second * 60
+	// DefaultObjectsListCacheSize is a default size of cache of ListObjects.
+	DefaultObjectsListCacheSize = 1e5
+)
+
+type (
+	// ListObjectsCache contains cache for ListObjects and ListObjectVersions.
+	ListObjectsCache struct {
+		lifetime time.Duration
```
KirillovDenis, line 45: I suppose we can remove this field if we will create gcache like:
```
gcache.New(cacheSize).LRU().Expiration(lifetime)
```

api/cache/objectslist.go
masterSplinter01: No, I decided not to invalidate lists in DELETE because every time, when we go through a list of object ids, we make a request to the object's cache (as expected, we delete the object's header from cache after DELETE request from user) and after we got nil, go to NeoFS and get "object not found".  And then we delete this oid from the list. 
```
obj := n.objectFromObjectsCacheOrNeoFS(ctx, bkt.CID, ids[i])
		if obj == nil {
			// mark ids[i] as an address of invalid object
			ids[i] = nil
			continue
		}
```

api/layer/object.go
```diff
@@ -478,20 +486,35 @@ func (n *layer) listSortedObjectsFromNeoFS(ctx context.Context, p allObjectParam
...
+
+	for i := 0; i < len(ids); i++ {
+		if ids[i] == nil {
 			continue
 		}
-		if oi := objectInfoFromMeta(bkt, meta, prefix, delimiter); oi != nil {
+		obj := n.objectFromObjectsCacheOrNeoFS(ctx, bkt.CID, ids[i])
+		if obj == nil {
+			// mark ids[i] as an address of invalid object
+			ids[i] = nil
```
KirillovDenis, line 514: If I understand correctly we don't use this mark. So can we remove this line? Or it's assumed we change list in the cache such way (should we prevent concurrent read/write then)?

masterSplinter01: It changes the list but you're right it causes data race. Changed the way to invalidate cache entries.

api/cache/objectslist.go
masterSplinter01: I'm not sure it's necessary. I don't think lifetime takes so much memery but it leaves the opportunity to change lifetime if we need.
Or maybe we should remove this from other caches: buckets, system etc.?

KirillovDenis: It's not about memory. I just think that if we use `gcache` and it has functionality that we need we should use this feature.

KirillovDenis (approved):

PR #245, opened 2021-09-01
[#190] Add POST object

### LINKED ISSUE #190: Support browser-based uploads using POST
I suppose it's quite useful feature https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html and we should implement it. But there may be problems with acl (#49) (that is provided in policy)

KirillovDenis:
To upload file using a form (with authenticated request) first we need make `policy` with some restriction. After than we encode policy using base64 and sign it. Then we have to create html form with  some fields (all fields (except `x-amz-signature`, and some others) in the form also must be in the policy). Among these fields must be `policy` and `x-amx-signature` using that s3-gw can verify request.
Aws example https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html

closes #190 

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -1,18 +1,158 @@
 package handler
 
 import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
 	"encoding/xml"
+	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/nspcc-dev/neofs-api-go/pkg/acl/eacl"
 	"github.com/nspcc-dev/neofs-s3-gw/api"
+	"github.com/nspcc-dev/neofs-s3-gw/api/auth"
 	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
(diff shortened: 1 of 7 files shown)


KirillovDenis:
Ceph test fails because of aws signature v2

api/handler/put.go
```diff
@@ -14,6 +23,118 @@ import (
...
+		if len(v) != 3 {
+			return fmt.Errorf("invalid condition")
+		}
+		p.Matching = v[0].(string)
+		if p.Matching == "content-length-range" {
+			p.Key = strconv.FormatFloat(v[1].(float64), 'f', 0, 32)
+			p.Value = strconv.FormatFloat(v[2].(float64), 'f', 0, 32)
+		} else {
+			p.Key = strings.ToLower(strings.TrimPrefix(v[1].(string), "$"))
+			p.Value = v[2].(string)
```
alexvanin, line 122: Are there any cases when type conversion will lead to panic? There are plenty of conversions, maybe it will be safer to do `p.Value, ok = v[2].(string)` everywhere, not sure.

alexvanin (approved):
LGTM

PR #247, opened 2021-09-02
[#242] Fix versions sorting

### LINKED ISSUE #242: Relsove object version order during the single gate access
> Are there any tests in `s3-tests` for that?

There is `s3tests_boto3.functional.test_s3:test_versioned_object_acl` test but we couldn't use it. This test fails on versioning checking since objects are created in the same epoch and their versions are out of order.

_Originally posted by @KirillovDenis in https://github.com/nspcc-dev/neofs-s3-gw/issues/239#issuecomment-908654926_

---

In theory, it seems, that in the case of single gate access, the gate should be able to resolve object version order by the order of objects in `S3-Versions-Add` list. Let's discuss it.

KirillovDenis:
Test `s3tests_boto3.functional.test_s3:test_versioned_object_acl` test still fails but the reason is different: the notary absent  and  acl user `grant` format inconsistency.

closes #242 

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -419,7 +419,7 @@ func containsACLHeaders(r *http.Request) bool {
 		r.Header.Get(api.AmzGrantFullControl) != "" || r.Header.Get(api.AmzGrantWrite) != ""
 }
 
-func (h *handler) getNewEAclTable(r *http.Request, objInfo *layer.ObjectInfo) (*eacl.Table, error) {
+func (h *handler) getNewEAclTable(r *http.Request, objInfo *api.ObjectInfo) (*eacl.Table, error) {
 	var newEaclTable *eacl.Table
 	objectACL, err := parseACLHeaders(r)
 	if err != nil {
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -249,7 +249,7 @@ func updateCRDT2PSetHeaders(p *PutObjectParams, versions *objectVersions, versio
 	}
 
 	if versioningEnabled {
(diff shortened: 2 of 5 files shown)


KirillovDenis:
Let's consider sorting for more complex case: suppose we want to put new object and we found out 2 conflicted previous objects:
```
object1 :{v: "c", add-list: "a,b", del-list: "a,b"}
object2 :{v: "e", add-list: "a,d", del-list: "a,d"}
```
New file will be `object3 :{v: "f", add-list: "a,b,c,d,e", del-list: "a,b,d"}`

If files were:
```
object1 :{v: "c", add-list: "a,b", del-list: "a"}
object2 :{v: "e", add-list: "a,d", del-list: "a"}
```
The new one will be: `object3 :{v: "f", add-list: "a,b,c,d,e", del-list: "a"}`

If the first case have deleted objects so we should not much worry about deleted versions order (it just has to be deterministic). The second case more complex and maybe we want it to give us `[a,b,d,c,e]` add-list instead of `[a,b,c,d,e]`

alexvanin:
> and maybe we want it to give us [a,b,d,c,e] add-list instead of [a,b,c,d,e]

I agree with you. Hash-distance sorting should be applied to resolve conflicts. In the examples it is clear that `a` version has no conflicts, thus it goes first. Then we have `b \ d` conflict that resolves into `[a, b, d]`. Then we have `c \ e` conflict, which resolves into `[a, b, d, c, e]`.

alexvanin (approved):

PR #252, opened 2021-09-07
Little fixes

masterSplinter01:
Wrong codes, not enough detailed logs

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -551,7 +551,8 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	h.log.Info("bucket is created",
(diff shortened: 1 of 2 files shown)


api/handler/put.go
```diff
@@ -551,7 +551,8 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	h.log.Info("bucket is created",
-		zap.String("container_id", cid.String()))
+		zap.String("container_id", cid.String()),
+		zap.String("bucket name", p.Name))
```
alexvanin, line 555: Do not use spaces in keys of zap's key-value pairs. It is a bit ruining of the ouput in JSON format, so try to avoid them.

PR #256, opened 2021-09-08
WIP internal: add S3 ACP <-> NeoFS eACL conversion

fyrchik:
An example of how these could be done.
ACP representation is taken from minio.

TODO:
- Beautify code
- Write more tests
- Allow wildcard for canonical user IDs

Signed-off-by: Evgenii Stratonikov <evgeniy@nspcc.ru>


internal/eacloverlay/doc.go
```diff
@@ -0,0 +1,10 @@
+// Package eacloverlay implements conversion between
+// NeoFS eACL and S3 access policies.
+// Current limitations:
+// 1. Canonical user identifiers are expected to contain public key in hex.
+// 2. Resource identifiers are expected to contain object id in base 58.
```
KirillovDenis, line 5: Why don't we use object name instead of object id? If I understand correctly we can add filter by `FileName` attribute. This make sense because bucket policy doesn't apply to a specific version of resources (unlike object ACL)

alexvanin:
We can try to rebase this PR because https://github.com/nspcc-dev/neofs-sdk-go/pull/36 was merged.

smallhive:
I'm completely not sure we need such a conversion now. It would bring another abstraction level and I doubt it will be useful for us or the users

PR #270, opened 2021-09-27
Support CORS

### LINKED ISSUE #217: Support CORS
It would be nice to support [cors](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html):
* [PutBucketCors](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketCors.html)
* [DeleteBucketCors](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteBucketCors.html)
* [GetbucketCors](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketCors.html)

We can store this like tagging objects (#196). I suppose this issue is related to #190

masterSplinter01:
Closes #217

### DIFF
--- a/api/handler/cors.go
+++ b/api/handler/cors.go
@@ -262,13 +262,12 @@ func checkCORS(cors *CORSConfiguration) error {
 	for _, r := range cors.CORSRules {
 		for _, m := range r.AllowedMethods {
 			if _, ok := supportedMethods[m]; !ok {
-				return fmt.Errorf("unsupported HTTP method in CORS config %s", m)
+				return errors.GetAPIErrorWithError(errors.ErrCORSUnsupportedMethod, fmt.Errorf("unsupported method is %s", m))
 			}
 		}
 		for _, h := range r.ExposeHeaders {
 			if h == wildcard {
-				return fmt.Errorf("ExposeHeader \"*\" contains wildcard. We currently do not support wildcard " +
-					"for ExposeHeader")
+				return errors.GetAPIError(errors.ErrCORSWildcardExposeHeaders)
 			}
 		}
 	}
--- a/api/handler/cors.go
+++ b/api/handler/cors.go
@@ -162,6 +162,7 @@ func (h *handler) AppendCORSHeaders(w http.ResponseWriter, r *http.Request) {
 				for _, m := range rule.AllowedMethods {
 					if m == r.Method {
 						w.Header().Set(api.AccessControlAllowOrigin, origin)
+						w.Header().Set(api.AccessControlAllowMethods, strings.Join(rule.AllowedMethods, ", "))
 						w.Header().Set(api.AccessControlAllowCredentials, "true")
 						w.Header().Set(api.Vary, api.Origin)
 						return
@@ -178,6 +179,7 @@ func (h *handler) AppendCORSHeaders(w http.ResponseWriter, r *http.Request) {
 						} else {
 							w.Header().Set(api.AccessControlAllowOrigin, o)
 						}
+						w.Header().Set(api.AccessControlAllowMethods, strings.Join(rule.AllowedMethods, ", "))
 						return
 					}
 				}
--- a/api/handler/cors.go
+++ b/api/handler/cors.go
@@ -1,0 +1,298 @@
+package handler
+
+import (
+	"encoding/xml"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/nspcc-dev/neofs-s3-gw/api"
+	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
+	"github.com/nspcc-dev/neofs-s3-gw/api/layer"
+)
+
+type (
+	// CORSConfiguration stores CORS configuration of a request.
+	CORSConfiguration struct {
+		XMLName   xml.Name   `xml:"http://s3.amazonaws.com/doc/2006-03-01/ CORSConfiguration" json:"-"`
+		CORSRules []CORSRule `xml:"CORSRule" json:"CORSRules"`
+	}
+	// CORSRule stores rules for CORS in a bucket.
+	CORSRule struct {
+		ID             string   `xml:"ID,omitempty" json:"ID,omitempty"`
+		AllowedHeaders []string `xml:"AllowedHeader" json:"AllowedHeaders"`
+		AllowedMethods []string `xml:"AllowedMethod" json:"AllowedMethods"`
+		AllowedOrigins []string `xml:"AllowedOrigin" json:"AllowedOrigins"`
+		ExposeHeaders  []string `xml:"ExposeHeader" json:"ExposeHeaders"`
+		MaxAgeSeconds  int      `xml:"MaxAgeSeconds,omitempty" json:"MaxAgeSeconds,omitempty"`
+	}
+)
+
+const (
+	// DefaultMaxAge -- default value of Access-Control-Max-Age if this value is not set in a rule.
+	DefaultMaxAge = 600
+	wildcard      = "*"
+)
+
+var supportedMethods = map[string]struct{}{"GET": {}, "HEAD": {}, "POST": {}, "PUT": {}, "DELETE": {}}
+
+func (h *handler) GetBucketCorsHandler(w http.ResponseWriter, r *http.Request) {
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not get bucket info", reqInfo, err)
+		return
+	}
+
+	if err = checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	info, err := h.obj.GetBucketCORS(r.Context(), bktInfo)
+	if err != nil {
+		h.logAndSendError(w, "could not get cors", reqInfo, err)
+		return
+	}
+
+	api.WriteResponse(w, http.StatusOK, info, api.MimeNone)
+}
+
+func (h *handler) PutBucketCorsHandler(w http.ResponseWriter, r *http.Request) {
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not get bucket info", reqInfo, err)
+		return
+	}
+
+	if err = checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	cors := &CORSConfiguration{}
+	if err := xml.NewDecoder(r.Body).Decode(cors); err != nil {
+		h.logAndSendError(w, "could not parse cors configuration", reqInfo, err)
+		return
+	}
+	if cors.CORSRules == nil {
+		h.logAndSendError(w, "could not parse cors rules", reqInfo, errors.GetAPIError(errors.ErrMalformedXML))
+		return
+	}
+
+	if err = checkCORS(cors); err != nil {
+		h.logAndSendError(w, "invalid cors configuration", reqInfo, err)
+		return
+	}
+
+	xml, err := xml.Marshal(cors)
+	if err != nil {
+		h.logAndSendError(w, "could not encode cors configuration to xml", reqInfo, err)
+		return
+	}
+
+	p := &layer.PutCORSParams{
+		BktInfo:           bktInfo,
+		CORSConfiguration: xml,
+	}
+
+	if err = h.obj.PutBucketCORS(r.Context(), p); err != nil {
+		h.logAndSendError(w, "could not put cors configuration", reqInfo, err)
+		return
+	}
+
+	api.WriteSuccessResponseHeadersOnly(w)
+}
+
+func (h *handler) DeleteBucketCorsHandler(w http.ResponseWriter, r *http.Request) {
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not get bucket info", reqInfo, err)
+		return
+	}
+
+	if err = checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	if err := h.obj.DeleteBucketCORS(r.Context(), bktInfo); err != nil {
+		h.logAndSendError(w, "could not delete cors", reqInfo, err)
+	}
+
+	w.WriteHeader(http.StatusNoContent)
+}
+
+func (h *handler) AppendCORSHeaders(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodOptions {
+		return
+	}
+	origin := r.Header.Get(api.Origin)
+	if origin == "" {
+		return
+	}
+	reqInfo := api.GetReqInfo(r.Context())
+	if reqInfo.BucketName == "" {
+		return
+	}
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		return
+	}
+
+	info, err := h.obj.GetBucketCORS(r.Context(), bktInfo)
+	if err != nil {
+		return
+	}
+	cors := &CORSConfiguration{}
+	if err = xml.Unmarshal(info, cors); err != nil {
+		return
+	}
+	withCredentials := r.Header.Get(api.Authorization) != ""
+
+	for _, rule := range cors.CORSRules {
+		for _, o := range rule.AllowedOrigins {
+			if o == origin {
+				for _, m := range rule.AllowedMethods {
+					if m == r.Method {
+						w.Header().Set(api.AccessControlAllowOrigin, origin)
+						w.Header().Set(api.AccessControlAllowCredentials, "true")
+						w.Header().Set(api.Vary, api.Origin)
+						return
+					}
+				}
+			}
+			if o == wildcard {
+				for _, m := range rule.AllowedMethods {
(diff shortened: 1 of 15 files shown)


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/bluele/gcache v0.0.2
 	github.com/cpuguy83/go-md2man/v2 v2.0.0 // indirect
 	github.com/google/uuid v1.2.0
+	github.com/gorilla/handlers v1.5.1
```
KirillovDenis, line 10: I suppose we don't need this. Can be deleted by `go mod tidy`

api/router.go
```diff
@@ -177,6 +190,7 @@ func Attach(r *mux.Router, domains []string, m MaxClients, h Handler, center aut
 	api := r.PathPrefix(SlashSeparator).Subrouter()
 
 	api.Use(
+
```
KirillovDenis, line 193: No need

api/response.go
```diff
@@ -240,7 +240,7 @@ func getAPIErrorResponse(info *ReqInfo, err error) ErrorResponse {
 		Message:    desc,
 		BucketName: info.BucketName,
 		Key:        info.ObjectName,
-		Resource:   info.URL.Path,
+		Resource:   "ololo",
```
KirillovDenis, line 243: Debug?

api/handler/object_list.go
```diff
@@ -36,6 +36,7 @@ func (h *handler) ListObjectsV1Handler(w http.ResponseWriter, r *http.Request) {
 	if err = api.EncodeToResponse(w, encodeV1(params, list)); err != nil {
 		h.logAndSendError(w, "something went wrong", reqInfo, err)
 	}
+	api.WriteSuccessResponseHeadersOnly(w)
```
KirillovDenis, line 39: I suppose it has no sense. Body has already sent one line 36 `api.EncodeToResponse(w, encodeV1(params, list)`

cmd/s3-gw/app.go
```diff
@@ -291,5 +293,15 @@ func getHandlerOptions(v *viper.Viper, l *zap.Logger) *handler.Config {
...
 
+	if v.IsSet(cfgDefaultMaxAge) {
+		defaultMaxAge = v.GetString(cfgDefaultMaxAge)
+	}
+
+	if _, err = strconv.Atoi(defaultMaxAge); err != nil {
+		l.Fatal("couldn't parse default max age", zap.Error(err))
+	}
+
+	cfg.DefaultMaxAge = defaultMaxAge
```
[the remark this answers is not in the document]
KirillovDenis: We can also change type of `cfg.DefaultMaxAge` to int

api/layer/system_object.go
```diff
@@ -0,0 +1,171 @@
...
+		p := &getParams{
+			Writer: buf,
+			cid:    bktInfo.CID,
+			oid:    versions.getLast().ID,
+		}
+		lastMeta, err = n.objectGet(ctx, p)
+		if err != nil {
+			return nil, err
+		}
+		lastMeta.ToV2().SetPayload(buf.Bytes())
```
KirillovDenis, line 108: It seems we can get object without `PayloadWriter` and payload will be filled in response. But we need define another `objectGet` method in `api/layer/object.go` to get rid of the `WithPayloadWriter(w)` param.

api/layer/versioning.go
```diff
@@ -352,12 +360,12 @@ func contains(list []string, elem string) bool {
...
 
 func (n *layer) getBucketSettings(ctx context.Context, bktInfo *data.BucketInfo) (*BucketSettings, error) {
-	objInfo, err := n.getSystemObject(ctx, bktInfo, bktInfo.SettingsObjectName())
+	obj, err := n.getSystemObject(ctx, bktInfo, bktInfo.SettingsObjectName(), false)
 	if err != nil {
 		return nil, err
 	}
 
-	return objectInfoToBucketSettings(objInfo), nil
+	return objectInfoToBucketSettings(objectInfoFromMeta(bktInfo, obj, "", "")), nil
```
KirillovDenis, line 368:
```suggestion
	return objectInfoToBucketSettings(objInfoFromMeta(bktInfo, obj)), nil
```

KirillovDenis:
Couldn't add comment (this is test).

api/cache/system.go
```diff
@@ -56,3 +56,8 @@ func (o *SystemCache) Put(key string, obj *object.Object) error {
 func (o *SystemCache) Delete(key string) bool {
 	return o.cache.Remove(key)
 }
+
+// SystemObjectKey is key to use in SystemCache.
+func SystemObjectKey(bucket, obj string) string {
```
KirillovDenis, line 61: Why this replaced method from `BucketInfo`?
```
// SystemObjectKey is key to use in SystemCache.
func (b *BucketInfo) SystemObjectKey(obj string) string {
	return b.Name + obj
}
```

api/layer/layer.go
```diff
@@ -396,12 +415,13 @@ func (n *layer) GetBucketTagging(ctx context.Context, bucketName string) (map[st
...
 
-	objInfo, err := n.getSystemObject(ctx, bktInfo, formBucketTagObjectName(bucketName))
+	obj, err := n.getSystemObject(ctx, bktInfo, formBucketTagObjectName(bucketName), false)
+
 	if err != nil && !errors.IsS3Error(err, errors.ErrNoSuchKey) {
 		return nil, err
 	}
 
-	return formTagSet(objInfo), nil
+	return formTagSet(objectInfoFromMeta(bktInfo, obj, "", "")), nil
```
KirillovDenis, line 424:
```suggestion
	return formTagSet(objInfoFromMeta(bktInfo, obj)), nil
```

api/layer/layer.go
```diff
@@ -381,12 +400,12 @@ func (n *layer) GetObjectTagging(ctx context.Context, oi *data.ObjectInfo) (map[
...
 	}
 
-	objInfo, err := n.getSystemObject(ctx, bktInfo, oi.TagsObject())
+	obj, err := n.getSystemObject(ctx, bktInfo, oi.TagsObject(), false)
 	if err != nil && !errors.IsS3Error(err, errors.ErrNoSuchKey) {
 		return nil, err
 	}
 
-	return formTagSet(objInfo), nil
+	return formTagSet(objectInfoFromMeta(bktInfo, obj, "", "")), nil
```
KirillovDenis, line 408:
```suggestion
	return formTagSet(objInfoFromMeta(bktInfo, obj)), nil
```

api/layer/cors.go
```diff
@@ -0,0 +1,40 @@
...
+		metadata: map[string]string{},
+		prefix:   "",
+		payload:  p.CORSConfiguration,
+	}
+
+	if _, err := n.putSystemObject(ctx, s); err != nil {
+		return err
+	}
+
+	return nil
```
KirillovDenis, line 23:
```suggestion
	_, err := n.putSystemObject(ctx, s)
	return err
```

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+	if err = checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	if err := h.obj.DeleteBucketCORS(r.Context(), bktInfo); err != nil {
+		h.logAndSendError(w, "could not delete cors", reqInfo, err)
+	}
+
+	api.WriteSuccessResponseHeadersOnly(w)
```
KirillovDenis, line 138: Delete success status - 204

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+		h.logAndSendError(w, "could not get cors", reqInfo, err)
+		return
+	}
+
+	if _, err = w.Write(info); err != nil {
+		h.logAndSendError(w, "could not encode cors to response", reqInfo, err)
+		return
+	}
+
+	api.WriteSuccessResponseHeadersOnly(w)
```
KirillovDenis, line 64: We should write status code before the body.

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+		return
+	}
+
+	info, err := h.obj.GetBucketCORS(r.Context(), bktInfo)
+	if err != nil || info == nil {
+		h.logAndSendError(w, "could not get cors", reqInfo, err)
+		return
+	}
+
+	if _, err = w.Write(info); err != nil {
```
KirillovDenis, line 59: Why don't you use `func WriteResponse(w http.ResponseWriter, statusCode int, response []byte, mType mimeType)` function from api?

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+		h.logAndSendError(w, "could not encode cors to response", reqInfo, err)
```
KirillovDenis, line 60:
```suggestion
		h.logAndSendError(w, "could not send cors response", reqInfo, err)
```

api/layer/cors.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (n *layer) GetBucketCORS(ctx context.Context, bktInfo *data.BucketInfo) ([]byte, error) {
+	obj, err := n.getSystemObject(ctx, bktInfo, bktInfo.CORSObjectName(), true)
+	if err != nil {
+		return nil, err
+	}
```
KirillovDenis, line 30: We can check err and if it is `errors.ErrNoSuchkey` then return `errors.ErrNoSuchCORSConfiguration`

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+		return
+	}
+
+	if err = checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	info, err := h.obj.GetBucketCORS(r.Context(), bktInfo)
+	if err != nil || info == nil {
```
KirillovDenis, line 54: We can check only err

KirillovDenis:
Did you run `s3tests_boto3.functional.test_s3.test_set_cors` test? It validates basic `get`, `put` and `delete` operations with cors.

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+		return
+	}
+
+	cors := &CORSConfiguration{}
+	if err := xml.NewDecoder(r.Body).Decode(cors); err != nil {
+		h.logAndSendError(w, "could not parse cors configuration", reqInfo, err)
+		return
+	}
+	if cors.CORSRules == nil {
+		h.logAndSendError(w, "could not parse cors rules", reqInfo, errors.GetAPIError(errors.ErrBadRequest))
```
KirillovDenis, line 87: AWS return `The XML you provided was not well-formed or did not validate against our published schema` error in such case. So we should return `errors.ErrMalformedXML` instead of `errors.ErrBadRequest`

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+		return
+	}
+
+	for _, r := range cors.CORSRules {
+		for _, m := range r.AllowedMethods {
+			if _, ok := supportedMethods[m]; !ok {
+				msg := "unsupported HTTP method in CORS config: " + m
+				h.logAndSendError(w, msg, reqInfo, errors.GetAPIError(errors.ErrBadRequest))
```
KirillovDenis, line 95: The AWS error is `An error occurred (InvalidRequest) when calling the PutBucketCors operation (reached max retries: 1): Found unsupported HTTP method in CORS config. Unsupported method is GETT`. So we should return the same one.

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+				return
+			}
+		}
+	}
+
+	xml, err := xml.Marshal(cors)
```
KirillovDenis, line 101: Perhaps we can save body to the buffer early and don't marshal cors here (we don't change it anyway).

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+	if reqInfo.BucketName == "" {
+		return
+	}
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		return
+	}
+
+	info, err := h.obj.GetBucketCORS(r.Context(), bktInfo)
+	if err != nil || info == nil {
```
KirillovDenis, line 159: We can check only `err`

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+		return
+	}
+	cors := &CORSConfiguration{}
+	if err = xml.Unmarshal(info, cors); err != nil {
```
KirillovDenis, line 163: I'm not sure if unmarshaling cors on every request is a good idea. Can we consider saving no-raw bytes in the cache?

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+	withCredentials := r.Header.Get(api.Authorization) != ""
+
+	for _, rule := range cors.CORSRules {
+		for _, o := range rule.AllowedOrigins {
+			if o == origin {
+				for _, m := range rule.AllowedMethods {
+					if m == r.Method {
+						w.Header().Set(api.AccessControlAllowOrigin, o)
+						w.Header().Set(api.AccessControlAllowCredentials, "true")
+						w.Header().Set(api.Vary, "Origin")
```
KirillovDenis, line 176:
```suggestion
						w.Header().Set(api.Vary, api.Origin)
```

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+						}
+						w.Header().Set(api.AccessControlAllowOrigin, o)
+						w.Header().Set(api.AccessControlAllowMethods, strings.Join(rule.AllowedMethods, ", "))
+						if headers != "" {
+							w.Header().Set(api.AccessControlAllowHeaders, strings.Join(rule.AllowedHeaders, ", "))
+						}
+						if rule.ExposeHeaders != nil {
+							w.Header().Set(api.AccessControlExposeHeaders, strings.Join(rule.ExposeHeaders, ", "))
+						}
+						if rule.MaxAgeSeconds != 0 {
```
KirillovDenis, line 241: Should we allow negative values?

api/handler/cors.go
```diff
@@ -0,0 +1,280 @@
...
+							w.Header().Set(api.AccessControlAllowCredentials, "true")
+						}
+						api.WriteSuccessResponseHeadersOnly(w)
+						return
+					}
+				}
+			}
+		}
+	}
+	h.logAndSendError(w, "Forbidden", reqInfo, errors.GetAPIError(errors.ErrAccessDenied))
```
KirillovDenis, line 256: If I understand correctly cors we can should not forbid the request and check allowed methods. It seems we can just return `Access-Control-Allowed-Methods` and other headers. Browsers must handle this correctly. So we can drop the `sliceContains` and `checkSubslice` functions

masterSplinter01: I'm not sure I understeand correctly, but here it says that wee need to check method in preflight.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html

Also, when I tried to make preflight request for a method which is not allowed in the first mathed rule, I received: 
```
<?xml version="1.0" encoding="UTF-8"?>
<Error>
    <Code>AccessForbidden</Code>
    <Message>CORSResponse: This CORS request is not allowed. This is usually because the evalution of Origin, request method / Access-Control-Request-Method or Access-Control-Request-Headers are not whitelisted by the resource's CORS spec.</Message>
    <Method>POST</Method>
    <ResourceType>BUCKET</ResourceType>
    <RequestId>W3ZGMMHZ9FNHX78H</RequestId>
    <HostId>wCqJjotZlC+gInT77Ewoog4MvI0rH3U/RKhc4F720ob3pC6zHUCuteSlug/TFH2RD74iP5vuMvg=</HostId>
</Error>
```

api/handler/object_list.go
masterSplinter01: It's strange but earlier I added this line to fix problems with s3-tests, but now they work without this line. Okay, removed.

api/cache/system.go
masterSplinter01: As I see we use SystemObjectKey only in the context of cache and don't think we will use in other cases. Maybe i'd be better to keep structs related to cache in cache?

api/handler/cors.go
KirillovDenis: It looks like I was wrong

cmd/s3-gw/app.go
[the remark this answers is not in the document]
masterSplinter01: I'm not sure we need in keep it as int. Yep, it'd be looking clearer but the the only usage of the variable is to be set in a header as a string. So I decided to use it as a string to avoid some extra itoa.

api/layer/system_object.go
```diff
@@ -0,0 +1,171 @@
...
+				zap.Stringer("version id", id),
+				zap.String("name", p.objName),
+				zap.Error(err))
+		}
+	}
+
+	return meta, nil
+}
+
+func (n *layer) getSystemObject(ctx context.Context, bktInfo *data.BucketInfo, objName string, withPayload bool) (*object.Object, error) {
```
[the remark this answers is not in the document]
masterSplinter01: I thought these functions will have too much repeating code, but ok, it looks better separated.

api/handler/cors.go
masterSplinter01: This check meant "if value is set" but made the check stricter.

api/layer/util.go
```diff
@@ -71,6 +71,9 @@ func objInfoFromMeta(bkt *data.BucketInfo, meta *object.Object) *data.ObjectInfo
 }
 
 func objectInfoFromMeta(bkt *data.BucketInfo, meta *object.Object, prefix, delimiter string) *data.ObjectInfo {
+	if meta == nil {
```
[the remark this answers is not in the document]
masterSplinter01: No, it's just a prevention of further referencing to nil pointer. I assume the gw will raise panic if we it passes nil to filenameFromObject or call meta.PayloadSize().

cmd/s3-gw/app.go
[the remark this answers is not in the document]
KirillovDenis: > So I decided to use it as a string to avoid some extra itoa.

If cors are stored we will do itoa anyway.

api/handler/cors.go
masterSplinter01: Fixed

api/layer/util.go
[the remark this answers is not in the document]
masterSplinter01: Okay, removed.

api/cache/system.go
masterSplinter01: Moved it into `layer/system_object.go`.

PR #276, opened 2021-10-19
[#271] Add random key for no sign requests

### LINKED ISSUE #271: Can't create a bucket with aws cli with `--no-sign-request`
`--no-sign-request` means we don't attach tokens from authbox to our requests, we use neofs key. 

When I tried to create a bucket I received an error:
```
An error occurred (InternalError) when calling the CreateBucket operation (reached max retries: 2): couldn't get box data from context
``` 
It happens because we try to get `AccessControlPolicy` from authbox but we don't have the box.

KirillovDenis:
Add new random key to the layer on app start. If no `Authorization` header is provided then random key will be used. Session token for such key will be generated for each anonymous request. 

closes #271


KirillovDenis:
> It seems we will have a problem if number of nodes in the pool is more than 1. We generate session token with one connection and when we use it the connection can be different.

The problem should go away because `neofs-sdk-go` was updated

api/layer/layer.go
```diff
@@ -240,37 +252,42 @@ func NewLayer(log *zap.Logger, conns pool.Pool, config *CachesConfig) Client {
...
 	}
 
-	return n.pool.OwnerID()
+	id, _ := authmate.OwnerIDFromNeoFSKey(n.EphemeralKey())
+	return id
 }
 
-// BearerOpt returns client.WithBearer call option with token from context or with nil token.
-func (n *layer) BearerOpt(ctx context.Context) client.CallOption {
+// CallOptions returns []client.CallOption options: client.WithBearer or client.WithKey (if request is anonymous).
```
alexvanin, line 269:
```suggestion
// CallOptions returns []pool.CallOption options: client.WithBearer or client.WithKey (if request is anonymous).
```

alexvanin (commented):
Are integration tests good with theses changes? No broken test?

KirillovDenis:
> Are integration tests good with these changes? No broken test?

Some tests failed. But they also failed on master branch so this PR doesn't cause test problems

masterSplinter01 (approved):

alexvanin (approved):

Issue #187, opened 2021-08-05
Support list parts


KirillovDenis:
After #186 we should implement:
* [ListParts](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html)
* [ListMultipartUploads](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListMultipartUploads.html)

PR #280, opened 2021-11-15
Multipart upload

### LINKED ISSUE #186: Support multipart uploads
According to #150 #183 gateway doesn't support [multipart uploads](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) but it should.
We can start with these methods:
* [CreateMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html)
* [UploadPart](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html)
* [UploadPartCopy](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html)
* [AbortMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_AbortMultipartUpload.html)
* [CompleteMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html)

masterSplinter01:
Closes #186, #187

### DIFF
--- a/api/handler/multipart_upload.go
+++ b/api/handler/multipart_upload.go
@@ -1,0 +1,645 @@
+package handler
+
+import (
+	"bytes"
+	"encoding/json"
+	"encoding/xml"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/nspcc-dev/neofs-s3-gw/api"
+	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
+	"github.com/nspcc-dev/neofs-s3-gw/api/layer"
+	"go.uber.org/zap"
+)
+
+type (
+	InitiateMultipartUploadResponse struct {
+		XMLName  xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ InitiateMultipartUploadResult" json:"-"`
+		Bucket   string   `xml:"Bucket"`
+		Key      string   `xml:"Key"`
+		UploadID string   `xml:"UploadId"`
+	}
+
+	CompleteMultipartUploadResponse struct {
+		XMLName xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ CompleteMultipartUploadResult" json:"-"`
+		Bucket  string   `xml:"Bucket"`
+		Key     string   `xml:"Key"`
+		ETag    string   `xml:"ETag"`
+	}
+
+	ListMultipartUploadsResponse struct {
+		XMLName            xml.Name          `xml:"http://s3.amazonaws.com/doc/2006-03-01/ ListMultipartUploadsResult" json:"-"`
+		Bucket             string            `xml:"Bucket"`
+		CommonPrefixes     []CommonPrefix    `xml:"CommonPrefixes"`
+		Delimiter          string            `xml:"Delimiter,omitempty"`
+		EncodingType       string            `xml:"EncodingType,omitempty"`
+		IsTruncated        bool              `xml:"IsTruncated"`
+		KeyMarker          string            `xml:"KeyMarker"`
+		MaxUploads         int               `xml:"MaxUploads"`
+		NextKeyMarker      string            `xml:"NextKeyMarker,omitempty"`
+		NextUploadIDMarker string            `xml:"NextUploadIdMarker,omitempty"`
+		Prefix             string            `xml:"Prefix"`
+		Uploads            []MultipartUpload `xml:"Upload"`
+		UploadIDMarker     string            `xml:"UploadIdMarker,omitempty"`
+	}
+
+	ListPartsResponse struct {
+		XMLName              xml.Name      `xml:"http://s3.amazonaws.com/doc/2006-03-01/ ListPartsResult" json:"-"`
+		Bucket               string        `xml:"Bucket"`
+		Initiator            Initiator     `xml:"Initiator"`
+		IsTruncated          bool          `xml:"IsTruncated"`
+		Key                  string        `xml:"Key"`
+		MaxParts             int           `xml:"MaxParts,omitempty"`
+		NextPartNumberMarker int           `xml:"NextPartNumberMarker,omitempty"`
+		Owner                Owner         `xml:"Owner"`
+		Parts                []*layer.Part `xml:"Part"`
+		PartNumberMarker     int           `xml:"PartNumberMarker,omitempty"`
+		StorageClass         string        `xml:"StorageClass,omitempty"`
+		UploadID             string        `xml:"UploadId"`
+	}
+
+	MultipartUpload struct {
+		Initiated    string    `xml:"Initiated"`
+		Initiator    Initiator `xml:"Initiator"`
+		Key          string    `xml:"Key"`
+		Owner        Owner     `xml:"Owner"`
+		StorageClass string    `xml:"StorageClass,omitempty"`
+		UploadID     string    `xml:"UploadId"`
+	}
+
+	Initiator struct {
+		ID          string `xml:"ID"`
+		DisplayName string `xml:"DisplayName"`
+	}
+
+	CompleteMultipartUpload struct {
+		XMLName xml.Name               `xml:"http://s3.amazonaws.com/doc/2006-03-01/ CompleteMultipartUpload"`
+		Parts   []*layer.CompletedPart `xml:"Part"`
+	}
+
+	UploadPartCopyResponse struct {
+		ETag         string `xml:"ETag"`
+		LastModified string `xml:"LastModified"`
+	}
+
+	UploadData struct {
+		TagSet map[string]string
+		ACL    *AccessControlPolicy
+	}
+)
+
+const (
+	uploadIDHeaderName   = "uploadId"
+	partNumberHeaderName = "partNumber"
+)
+
+func (h *handler) CreateMultipartUploadHandler(w http.ResponseWriter, r *http.Request) {
+	/* initiation of multipart uploads is implemented via creation of "system" upload part with 0 part number
+	(min value of partNumber of a common part is 1) and holding data: metadata, acl, tagging */
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not get bucket info", reqInfo, err)
+		return
+	}
+
+	if err := checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	var (
+		hasData bool
+		b       []byte
+
+		uploadID   = uuid.New()
+		data       = &UploadData{}
+		additional = []zap.Field{
+			zap.String("uploadID", uploadID.String()),
+			zap.String("Key", reqInfo.ObjectName),
+		}
+		uploadInfo = &layer.UploadInfoParams{
+			UploadID: uploadID.String(),
+			Bkt:      bktInfo,
+			Key:      reqInfo.ObjectName,
+		}
+	)
+
+	if containsACLHeaders(r) {
+		gateKey, err := h.gateKey(r.Context())
+		if err != nil {
+			h.logAndSendError(w, "couldn't get gate key", reqInfo, err)
+			return
+		}
+		data.ACL, err = parseACLHeaders(r.Header, gateKey)
+		if err != nil {
+			h.logAndSendError(w, "could not parse acl", reqInfo, err)
+			return
+		}
+		hasData = true
+	}
+
+	if len(r.Header.Get(api.AmzTagging)) > 0 {
+		data.TagSet, err = parseTaggingHeader(r.Header)
+		if err != nil {
+			h.logAndSendError(w, "could not parse tagging", reqInfo, err, additional...)
+			return
+		}
+		hasData = true
+	}
+
+	metadata := parseMetadata(r)
+	if contentType := r.Header.Get(api.ContentType); len(contentType) > 0 {
+		metadata[api.ContentType] = contentType
+	}
+
+	p := &layer.UploadPartParams{
+		Info:       uploadInfo,
+		PartNumber: 0,
+		Header:     metadata,
+	}
+
+	if hasData {
+		b, err = json.Marshal(data)
+		if err != nil {
+			h.logAndSendError(w, "could not marshal json with acl and/or tagging", reqInfo, err, additional...)
+			return
+		}
+		p.Reader = bytes.NewReader(b)
+	}
+
+	info, err := h.obj.UploadPart(r.Context(), p)
+	if err != nil {
+		h.logAndSendError(w, "could not upload a part", reqInfo, err, additional...)
+		return
+	}
+
+	resp := InitiateMultipartUploadResponse{
+		Bucket:   info.Bucket,
+		Key:      info.Headers[layer.UploadKeyAttributeName],
+		UploadID: info.Headers[layer.UploadIDAttributeName],
+	}
+
+	if err := api.EncodeToResponse(w, resp); err != nil {
+		h.logAndSendError(w, "could not encode InitiateMultipartUploadResponse to response", reqInfo, err, additional...)
+		return
+	}
+}
+
+func (h *handler) UploadPartHandler(w http.ResponseWriter, r *http.Request) {
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not get bucket info", reqInfo, err)
+		return
+	}
+
+	if err := checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+	var (
+		queryValues = r.URL.Query()
+		uploadID    = queryValues.Get(uploadIDHeaderName)
+		additional  = []zap.Field{zap.String("uploadID", uploadID), zap.String("Key", reqInfo.ObjectName)}
+	)
+
+	partNumber, err := strconv.Atoi(queryValues.Get(partNumberHeaderName))
+	if err != nil || partNumber < layer.UploadMinPartNumber || partNumber > layer.UploadMaxPartNumber {
+		h.logAndSendError(w, "invalid part number", reqInfo, errors.GetAPIError(errors.ErrInvalidPartNumber))
+		return
+	}
+
+	p := &layer.UploadPartParams{
+		Info: &layer.UploadInfoParams{
+			UploadID: uploadID,
+			Bkt:      bktInfo,
+			Key:      reqInfo.ObjectName,
+		},
+		PartNumber: partNumber,
+		Size:       r.ContentLength,
+		Reader:     r.Body,
+	}
+
+	info, err := h.obj.UploadPart(r.Context(), p)
+	if err != nil {
+		h.logAndSendError(w, "could not upload a part", reqInfo, err, additional...)
+		return
+	}
+
+	w.Header().Set(api.ETag, info.HashSum)
+	api.WriteSuccessResponseHeadersOnly(w)
+}
+
+func (h *handler) UploadPartCopy(w http.ResponseWriter, r *http.Request) {
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not get bucket info", reqInfo, err)
+		return
+	}
+
+	if err := checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	var (
+		queryValues = reqInfo.URL.Query()
+		uploadID    = queryValues.Get(uploadIDHeaderName)
+		additional  = []zap.Field{zap.String("uploadID", uploadID), zap.String("Key", reqInfo.ObjectName)}
+	)
+
+	partNumber, err := strconv.Atoi(queryValues.Get(partNumberHeaderName))
+	if err != nil || partNumber < layer.UploadMinPartNumber || partNumber > layer.UploadMaxPartNumber {
+		h.logAndSendError(w, "invalid part number", reqInfo, errors.GetAPIError(errors.ErrInvalidPartNumber))
+		return
+	}
+
+	src := r.Header.Get(api.AmzCopySource)
+	srcBucket, srcObject := path2BucketObject(src)
+
+	srcRange, err := parseRange(r.Header.Get(api.AmzCopySourceRange))
+	if err != nil {
+		h.logAndSendError(w, "could not parse copy range", reqInfo,
+			errors.GetAPIError(errors.ErrInvalidCopyPartRange), additional...)
+		return
+	}
+
+	srcInfo, err := h.obj.GetObjectInfo(r.Context(), &layer.HeadObjectParams{
+		Bucket: srcBucket,
+		Object: srcObject,
+	})
+	if err != nil {
+		h.logAndSendError(w, "could not head source object", reqInfo, err, additional...)
+		return
+	}
+
+	args, err := parseCopyObjectArgs(r.Header)
+	if err != nil {
+		h.logAndSendError(w, "could not parse copy object args", reqInfo,
+			errors.GetAPIError(errors.ErrInvalidCopyPartRange), additional...)
+		return
+	}
+
+	if err = checkPreconditions(srcInfo, args.Conditional); err != nil {
+		h.logAndSendError(w, "precondition failed", reqInfo, errors.GetAPIError(errors.ErrPreconditionFailed),
+			additional...)
+		return
+	}
+
+	if err = h.checkBucketOwner(r, srcBucket, r.Header.Get(api.AmzSourceExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "source expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	p := &layer.UploadCopyParams{
+		Info: &layer.UploadInfoParams{
+			UploadID: uploadID,
+			Bkt:      bktInfo,
+			Key:      reqInfo.ObjectName,
+		},
+		SrcObjInfo: srcInfo,
+		PartNumber: partNumber,
+		Range:      srcRange,
+	}
+
+	info, err := h.obj.UploadPartCopy(r.Context(), p)
+	if err != nil {
+		h.logAndSendError(w, "could not upload part copy", reqInfo, err, additional...)
+		return
+	}
+
+	response := UploadPartCopyResponse{
+		ETag:         info.HashSum,
+		LastModified: info.Created.Format(time.RFC3339),
+	}
+
+	if err = api.EncodeToResponse(w, response); err != nil {
+		h.logAndSendError(w, "something went wrong", reqInfo, err)
+	}
+}
+
+func (h *handler) CompleteMultipartUploadHandler(w http.ResponseWriter, r *http.Request) {
+	reqInfo := api.GetReqInfo(r.Context())
+
+	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
+	if err != nil {
+		h.logAndSendError(w, "could not get bucket info", reqInfo, err)
+		return
+	}
+
+	if err := checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	var (
+		uploadID   = r.URL.Query().Get(uploadIDHeaderName)
+		uploadInfo = &layer.UploadInfoParams{
+			UploadID: uploadID,
+			Bkt:      bktInfo,
+			Key:      reqInfo.ObjectName,
+		}
+		additional = []zap.Field{zap.String("uploadID", uploadID), zap.String("Key", reqInfo.ObjectName)}
+	)
+
+	reqBody := new(CompleteMultipartUpload)
+	if err := xml.NewDecoder(r.Body).Decode(reqBody); err != nil {
+		h.logAndSendError(w, "could not read complete multipart upload xml", reqInfo,
+			errors.GetAPIError(errors.ErrMalformedXML), additional...)
+		return
+	}
+
+	initPart, err := h.obj.GetUploadInitInfo(r.Context(), uploadInfo)
+	if err != nil {
+		h.logAndSendError(w, "could not get multipart upload info", reqInfo, err, additional...)
+		return
+	}
+
+	c := &layer.CompleteMultipartParams{
+		Info:  uploadInfo,
+		Parts: reqBody.Parts,
+	}
+	objInfo, err := h.obj.CompleteMultipartUpload(r.Context(), c)
+	if err != nil {
+		h.logAndSendError(w, "could not complete multipart upload", reqInfo, err, additional...)
+		return
+	}
+
+	if initPart.Size > 0 {
+		initPartPayload := bytes.NewBuffer(make([]byte, 0, initPart.Size))
+		p := &layer.GetObjectParams{
+			ObjectInfo: initPart,
+			Writer:     initPartPayload,
+		}
+		if err = h.obj.GetObject(r.Context(), p); err != nil {
+			h.logAndSendError(w, "could not get multipart upload acl and/or tagging", reqInfo, err, additional...)
+			return
+		}
+
+		uploadData := &UploadData{}
+		if err = json.Unmarshal(initPartPayload.Bytes(), uploadData); err != nil {
+			h.logAndSendError(w, "could not unmarshal multipart upload acl and/or tagging", reqInfo, err, additional...)
+			return
+		}
+
+		if len(uploadData.TagSet) != 0 {
+			t := &layer.PutTaggingParams{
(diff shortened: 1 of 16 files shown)


api/router.go
```diff
@@ -18,11 +18,7 @@ type (
 	Handler interface {
 		HeadObjectHandler(http.ResponseWriter, *http.Request)
 		CopyObjectPartHandler(http.ResponseWriter, *http.Request)
-		PutObjectPartHandler(http.ResponseWriter, *http.Request)
-		ListObjectPartsHandler(http.ResponseWriter, *http.Request)
-		CompleteMultipartUploadHandler(http.ResponseWriter, *http.Request)
-		NewMultipartUploadHandler(http.ResponseWriter, *http.Request)
-		AbortMultipartUploadHandler(http.ResponseWriter, *http.Request)
+
```
KirillovDenis, line 21: Redundant line

api/layer/system_object.go
```diff
@@ -44,7 +44,11 @@ func (n *layer) headSystemObject(ctx context.Context, bkt *data.BucketInfo, objN
 }
 
 func (n *layer) deleteSystemObject(ctx context.Context, bktInfo *data.BucketInfo, name string) error {
-	ids, err := n.objectSearch(ctx, &findParams{cid: bktInfo.CID, attr: objectSystemAttributeName, val: name})
+	f := &findParams{
+		filters: []*filter{{attr: objectSystemAttributeName, val: name}},
+		cid:     nil,
```
KirillovDenis, line 49: I suppose we need use `bktInfo.CID` instead of `nil`

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,549 @@
...
+		UploadID: uploadID.String(),
+		Bkt:      bktInfo,
+		Key:      reqInfo.ObjectName,
+	}
+
+	p := &layer.UploadPartParams{
+		Info:       uploadInfo,
+		PartNumber: 0,
+		Header:     metadata,
+		Reader:     r.Body,
```
KirillovDenis, line 123: `CreateMultipartUpload` should not have the body

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,549 @@
...
+		UploadID: uploadID.String(),
+	}
+
+	if err := api.EncodeToResponse(w, resp); err != nil {
+		h.logAndSendError(w, "could not encode InitiateMultipartUploadResponse to response", reqInfo, err, additional...)
+		return
+	}
+}
+
+func (h *handler) UploadPartHandler(w http.ResponseWriter, r *http.Request) {
```
KirillovDenis, line 177: We should handle headers: `x-amz-copy-source` and others to support [UploadPartCopy](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html)

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,549 @@
...
+	}
+
+	p := &layer.UploadPartParams{
+		Info: &layer.UploadInfoParams{
+			UploadID: queryValues.Get(uploadIdHeaderName),
+			Bkt:      bktInfo,
+			Key:      reqInfo.ObjectName,
+		},
+		PartNumber: partNum,
+		Size:       r.ContentLength,
```
KirillovDenis, line 206: We should check part size according to [limits](https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html)

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,549 @@
...
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	queryValues := reqInfo.URL.Query()
+	uploadID := queryValues.Get(uploadIdHeaderName)
+
+	maxParts := 1000
+	if queryValues.Get("max-parts") != "" {
+		if maxParts, err = strconv.Atoi(queryValues.Get("max-keys")); err != nil || maxParts < 0 {
```
KirillovDenis, line 393: Shouldn't be "max-parts" instead of "max-keys"?

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,549 @@
...
+		return
+	}
+
+	queryValues := reqInfo.URL.Query()
+
+	var maxUploads = 1000
+	delimiter := queryValues.Get("delimiter")
+	prefix := queryValues.Get("prefix")
+	if queryValues.Get("max-uploads") != "" {
+		if maxUploads, err = strconv.Atoi(queryValues.Get("max-keys")); err != nil || maxUploads < 0 {
```
KirillovDenis, line 351: Shouldn't be "max-uploads" instead of "max-keys"?

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,549 @@
...
+		Delimiter:      delimiter,
+		EncodingType:   queryValues.Get("encoding-type"),
+		KeyMarker:      queryValues.Get("key-marker"),
+		MaxUploads:     maxUploads,
+		Prefix:         prefix,
+		UploadIdMarker: queryValues.Get("upload-id-marker"),
+	}
+
+	list, err := h.obj.ListMultipartUploads(r.Context(), p)
+
```
KirillovDenis, line 368: Redundant line

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,549 @@
...
+			UploadID: uploadID,
+			Bkt:      bktInfo,
+			Key:      reqInfo.ObjectName,
+		},
+		MaxParts:         maxParts,
+		PartNumberMarker: partNumberMarker,
+	}
+
+	list, err := h.obj.ListParts(r.Context(), p)
+
```
KirillovDenis, line 418: Redundant line

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+
+	if len(objects) < len(p.Parts) {
+		return nil, errors.GetAPIError(errors.ErrInvalidPart)
+	}
+
+	parts := make([]*data.ObjectInfo, 0, len(p.Parts))
+
+	for i, part := range p.Parts {
+		info := objects[part.PartNumber]
```
KirillovDenis, line 148: As I understand this can cause panic because `part.PartNumber` can be greater than `len(objects)`. https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html. For example we can upload parts with number 4, 5, 15 and then complete multipart upload. So we have 3 object but the part numbers are greater than 3.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,459 @@
...
+	for i, obj := range objects {
+		if obj.Key > key {
+			return objects[i:]
+		}
+	}
+
+	return []*UploadInfo{}
+}
+
+func uploadInfoFromMeta(bkt *data.BucketInfo, meta *object.Object, prefix, delimiter string) *UploadInfo {
```
KirillovDenis, line 425: Unused `bkt *data.BucketInfo`

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,459 @@
...
+		})
+		if err != nil {
+			n.log.Error("could not download a part of multipart upload",
+				zap.String("uploadID", p.Info.UploadID),
+				zap.String("part number", part.Headers[UploadPartNumberAttributeName]),
+				zap.Error(err))
+			return nil, err
+		}
+	}
+	pw.Close()
```
KirillovDenis, line 197: I suppose we should handle the error

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,459 @@
...
+		if attrs[UploadKeyAttributeName] == key {
+			num, err := strconv.Atoi(attrs[UploadPartNumberAttributeName])
+			if err != nil {
+				n.log.Warn("could not cast part number",
+					zap.Stringer("object id", id),
+					zap.Stringer("bucket id", bktInfo.CID),
+					zap.Error(err))
+				continue
+			}
+			res[num] = objInfoFromMeta(bktInfo, meta)
```
KirillovDenis, line 382: We cannot rely on the part number from attribute because we can have only one part with number 17 for example. It will cause panic.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,468 @@
...
+		uploadCompleted = true
+		done <- true
+	}(done)
+	for _, part := range parts {
+		_, err := n.objectGetWithPayloadWriter(ctx, &getParams{
+			Writer: pw,
+			cid:    p.Info.Bkt.CID,
+			oid:    part.ID,
+		})
+		if err != nil {
```
KirillovDenis, line 194: Shouldn't we close `pw` on error?

api/layer/multipart_upload.go
masterSplinter01: I'm not sure, because, if I understand correctly,  `Close()` always returns `nil`: 
```
func (w *PipeWriter) Close() error {
	return w.CloseWithError(nil)
}

func (w *PipeWriter) CloseWithError(err error) error {
	return w.p.CloseWrite(err)
}

func (p *pipe) CloseWrite(err error) error {
	if err == nil {
		err = EOF
	}
	p.werr.Store(err)
	p.once.Do(func() { close(p.done) })
	return nil
}
```

KirillovDenis:
Did you run [s3-compatibility tests](https://github.com/nspcc-dev/s3-tests)?

api/handler/multipart_upload.go
KirillovDenis: Or use `handler.CopyObjectPartHandler` (from `api/handler/unimplemented.go`)

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+const (
+	UploadIDAttributeName         = "S3-Upload-Id"
+	UploadPartNumberAttributeName = "S3-Upload-Part-Number"
+	UploadKeyAttributeName        = "S3-Upload-Key"
+	UploadHasTags                 = "S3-Upload-Has-Tags"
+	UploadHasACL                  = "S3-Upload-Has-ACL"
+	UploadPartKeyPrefix           = ".upload-"
+
+	uploadMinSize = 5 * 1048576       // 5MB
+	uploadMaxSize = 5 * 1099511627776 // 5TB
```
KirillovDenis, line 27: 5TB is max object size not max part size https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+
+func (n *layer) UploadPart(ctx context.Context, p *UploadPartParams) (*data.ObjectInfo, error) {
+	if p.PartNumber != 0 {
+		_, err := n.GetUploadInitInfo(ctx, p.Info)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if p.Size > uploadMaxSize {
```
KirillovDenis, line 109: Shouldn't we also check `uploadMinSize`?

api/layer/multipart_upload.go
KirillovDenis: Then let's explicitly ignore the error.
```suggestion
	_ = pw.Close()
```

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
+	"go.uber.org/zap"
+)
+
+const (
+	UploadIDAttributeName         = "S3-Upload-Id"
+	UploadPartNumberAttributeName = "S3-Upload-Part-Number"
+	UploadKeyAttributeName        = "S3-Upload-Key"
+	UploadHasTags                 = "S3-Upload-Has-Tags"
+	UploadHasACL                  = "S3-Upload-Has-ACL"
```
KirillovDenis, line 23: Unused

KirillovDenis:
Could you fix the linter issues?

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,536 @@
...
+		XMLName xml.Name               `xml:"http://s3.amazonaws.com/doc/2006-03-01/ CompleteMultipartUpload"`
+		Parts   []*layer.CompletedPart `xml:"Part"`
+	}
+)
+
+const (
+	uploadIdHeaderName   = "uploadId"
+	partNumberHeaderName = "partNumber"
+	maxSizeUploadsList   = 1000
+	maxSizePartsList     = 1000
```
KirillovDenis, line 87: Maybe we should define all s3 related constants in one place? For example all list related operations have the same default max items value 1000. At least I would prefer to hold [multipart upload limits](https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html) together

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,536 @@
...
+			return
+		}
+	}
+	if newEaclTable != nil {
+		p := &layer.PutBucketACLParams{
+			Name: reqInfo.BucketName,
+			EACL: newEaclTable,
+		}
+
+		if err = h.obj.PutBucketACL(r.Context(), p); err != nil {
```
KirillovDenis, line 139: I think we can store acl and tags in the 0 part. So we can make checks before uploading and don't create redundant objects.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+	res.Parts = parts
+
+	return &res, nil
+}
+
+func (n *layer) GetUploadInitInfo(ctx context.Context, p *UploadInfoParams) (*data.ObjectInfo, error) {
+	filters := []*filter{
+		{attr: UploadIDAttributeName, val: p.UploadID},
+		{attr: UploadPartNumberAttributeName, val: "0"},
+		{attr: UploadKeyAttributeName, val: p.Key},
```
KirillovDenis, line 365: Can we use only one filter by name?

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+		cid:     p.Bkt.CID,
+	}
+	ids, err := n.objectSearch(ctx, f)
+	if err != nil {
+		if errors.IsS3Error(err, errors.ErrNoSuchKey) {
+			return nil, errors.GetAPIError(errors.ErrNoSuchUpload)
+		}
+		return nil, err
+	}
+	if ids == nil {
```
KirillovDenis, line 378: It will be safer this way:
```suggestion
	if len(ids) == 0 {
```

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,536 @@
...
+
+	if err := checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	queryValues := r.URL.Query()
+
+	partNum, err := strconv.Atoi(queryValues.Get(partNumberHeaderName))
+	if err != nil || partNum == 0 || partNum > uploadMaxPartNumber {
```
KirillovDenis, line 188:
```suggestion
	if err != nil || partNum <= 0 || partNum > uploadMaxPartNumber {
```

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+}
+
+func (n *layer) CompleteMultipartUpload(ctx context.Context, p *CompleteMultipartParams) (*data.ObjectInfo, error) {
+	var (
+		obj             *data.ObjectInfo
+		uploadCompleted = false
+	)
+
+	for i := 1; i < len(p.Parts); i++ {
+		if p.Parts[i].PartNumber < p.Parts[i-1].PartNumber {
```
KirillovDenis, line 139:
```suggestion
		if p.Parts[i].PartNumber <= p.Parts[i-1].PartNumber {
```

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+	return objInfoFromMeta(p.Bkt, meta), nil
+
+}
+
+func (n *layer) getUploadParts(ctx context.Context, uploadID, key string, bktInfo *data.BucketInfo) (map[int]*data.ObjectInfo, error) {
+	f := &findParams{
+		filters: []*filter{
+			{attr: UploadIDAttributeName, val: uploadID},
+			{attr: UploadKeyAttributeName, val: key}},
```
KirillovDenis, line 393: Can we you one filter by matching name prefix?

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+	delete(initMetadata, attrVersionsIgnore)
+
+	go func(done chan bool) {
+		obj, err = n.objectPut(ctx, p.Info.Bkt, &PutObjectParams{
+			Bucket: p.Info.Bkt.Name,
+			Object: p.Info.Key,
+			Reader: pr,
+			Header: initMetadata,
+		})
+		if err != nil {
```
KirillovDenis, line 187: I suppose we should log the error

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	return res, nil
+}
+
+func createUploadPartName(uploadId, key string, partNumber int) string {
+	return UploadPartKeyPrefix + uploadId + "-" + key + "-" + strconv.Itoa(partNumber)
+}
+
+func trimAfterUploadIDAndKey(key, id string, uploads []*UploadInfo) []*UploadInfo {
```
KirillovDenis, line 432: It would be nice to have tests for this function

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+	})
+
+	if p.KeyMarker != "" {
+		if p.UploadIdMarker != "" {
+			uploads = trimAfterUploadIDAndKey(p.KeyMarker, p.UploadIdMarker, uploads)
+		}
+		uploads = trimAfterUploadKey(p.KeyMarker, uploads)
+	}
+
+	if p.MaxUploads != 0 && len(uploads) > p.MaxUploads {
```
KirillovDenis, line 279: We already check `p.MaxUploads` is greater than 0 at line 234

api/layer/multipart_upload.go
masterSplinter01: We don't have limit `uploadMinSize` for the last part. When we just upload a part we don't know if the part the last or not.

api/layer/object.go
```diff
@@ -22,10 +22,14 @@ import (
 
 type (
 	findParams struct {
-		attr   string
-		val    string
-		cid    *cid.ID
-		prefix string
+		filters []*filter
```
alexvanin, line 25: Is it required to be a slice of pointers?

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,536 @@
...
+
+	if err := checkOwner(bktInfo, r.Header.Get(api.AmzExpectedBucketOwner)); err != nil {
+		h.logAndSendError(w, "expected owner doesn't match", reqInfo, err)
+		return
+	}
+
+	metadata := parseMetadata(r)
+	uploadID := uuid.New()
+
+	additional := []zap.Field{zap.String("uploadID", uploadID.String()), zap.String("Key", reqInfo.ObjectName)}
```
alexvanin, line 108:
```suggestion
	additional := []zap.Field{zap.String("uploadID", uploadID.String()), zap.String("key", reqInfo.ObjectName)}
```

And down there again.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+	})
+
+	if p.PartNumberMarker != 0 {
+		for i, part := range parts {
+			if part.PartNumber > p.PartNumberMarker {
+				parts = parts[i:]
+				break
+			}
+		}
+		return &res, nil
```
KirillovDenis, line 347: I'm not sure if we need the `return` statement here

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,536 @@
...
+			h.logAndSendError(w, "could not put bucket acl", reqInfo, err, additional...)
+			if err = h.obj.AbortMultipartUpload(r.Context(), uploadInfo); err != nil {
+				h.log.Error("could not delete multipart upload init file", additional...)
+			}
+			return
+		}
+	}
+
+	tagSet, err := parseTaggingHeader(r.Header)
+	if tagSet != nil {
```
alexvanin, line 149: I would expect
```suggestion
	if err == nil {
```

or even 

```go
if err != nil {
    h.logAndSendError(...)
    h.obj.AbortMultipardUpload(...)
}
h.obj.PutObjectTagging(...)
```

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,536 @@
...
+		UploadID: uploadID,
+		Bkt:      bktInfo,
+		Key:      reqInfo.ObjectName,
+	}
+
+	info, err := h.obj.GetUploadInitInfo(r.Context(), p)
+	if err != nil {
+		h.logAndSendError(w, "could not get upload to abort", reqInfo, err, additional...)
+		return
+	}
```
KirillovDenis, line 464: I think we should try abort multipart uploads even if the 0 part doesn't exist. It could be deleted by some reason while other parts still exist.

api/layer/multipart_upload.go
alexvanin: `UploadHasTags` too.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,494 @@
...
+
+	if !strings.HasPrefix(key, prefix) {
+		return nil
+	}
+
+	if val, ok := userHeaders[object.AttributeTimestamp]; !ok {
+		// ignore empty value
+	} else if dt, err := strconv.ParseInt(val, 10, 64); err == nil {
+		creation = time.Unix(dt, 0)
+	}
```
KirillovDenis, line 476: We can remove empty branch
```suggestion
	if val, ok := userHeaders[object.AttributeTimestamp]; ok {
		if dt, err := strconv.ParseInt(val, 10, 64); err == nil {
			creation = time.Unix(dt, 0)
		}
	}
```

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,497 @@
...
+
+	res.Parts = parts
+
+	return &res, nil
+}
+
+func (n *layer) GetUploadInitInfo(ctx context.Context, p *UploadInfoParams) (*data.ObjectInfo, error) {
+	prefix := createUploadPartName(p.UploadID, p.Key, 0)
+	f := &findParams{
+		prefix: prefix,
```
KirillovDenis, line 373: Why we use prefix? It seems we can use `objectSearchByName` function

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,626 @@
...
+		hasData = true
+	}
+
+	if len(r.Header.Get(api.AmzTagging)) > 0 {
+		data.TagSet, err = parseTaggingHeader(r.Header)
+		if err != nil {
+			h.logAndSendError(w, "could not parse tagging", reqInfo, err, additional...)
+			if err = h.obj.AbortMultipartUpload(r.Context(), uploadInfo); err != nil {
+				h.log.Error("could not delete multipart upload init file", additional...)
+			}
```
KirillovDenis, line 152: It seems we should remove this lines (we haven't started uploading yet)

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,626 @@
...
+			h.logAndSendError(w, "could not marshal json with acl and/or tagging", reqInfo, err, additional...)
+			return
+		}
+		p.Reader = bytes.NewReader(b)
+	}
+
+	info, err := h.obj.UploadPart(r.Context(), p)
+	if err != nil {
+		h.logAndSendError(w, "could not upload a part", reqInfo, err, additional...)
+	}
```
KirillovDenis, line 181:
```suggestion
		h.logAndSendError(w, "could not upload a part", reqInfo, err, additional...)
		return
	}
```

api/handler/multipart_upload.go
```diff
@@ -0,0 +1,626 @@
...
+	if err != nil {
+		h.logAndSendError(w, "could not upload a part", reqInfo, err, additional...)
+		return
+	}
+
+	w.Header().Set(api.ETag, info.HashSum)
+	api.WriteSuccessResponseHeadersOnly(w)
+}
+
+func (h *handler) UploadPartCopy(w http.ResponseWriter, r *http.Request) {
```
KirillovDenis, line 241: It seems you've missed handling the following headers:
* `x-amz-source-expected-bucket-owner`
* `x-amz-copy-source-if-match`
* `x-amz-copy-source-if-modified-since`
* `x-amz-copy-source-if-none-match`
* `x-amz-copy-source-if-unmodified-since`

They should be handle the similar way as `CopyObjectHandler` does.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,563 @@
...
+		initMetadata = objects[0].Headers
+	} else {
+		initMetadata = make(map[string]string)
+		n.log.Error("could not get init multipart upload in listUploads",
+			zap.Stringer("bucket id", p.Info.Bkt.CID),
+			zap.String("uploadID", p.Info.UploadID),
+			zap.String("uploadKey", p.Info.Key),
+		)
+	}
+	if len(objects[0].ContentType) != 0 {
```
KirillovDenis, line 230: This line can cause `panic` since previous `else` branch doesn't contain `return` statement

api/handler/multipart_upload.go
masterSplinter01: Oh, planned to add this, but forgot. Thanks, fixed.

KirillovDenis: It seems `x-amz-source-expected-bucket-owner` header still isn't handled

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,563 @@
...
+			initMetadata[api.ContentType] = objects[0].ContentType
+		}
+	} else {
+		initMetadata = make(map[string]string)
+		n.log.Error("could not get init multipart upload in listUploads",
+			zap.Stringer("bucket id", p.Info.Bkt.CID),
+			zap.String("uploadID", p.Info.UploadID),
+			zap.String("uploadKey", p.Info.Key),
+		)
+	}
```
KirillovDenis, line 232: Shouldn't we return error here? Why we allow to form the result object if initial part is absent?

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,563 @@
...
+func (n *layer) GetUploadInitInfo(ctx context.Context, p *UploadInfoParams) (*data.ObjectInfo, error) {
+	ids, err := n.objectSearchByName(ctx, p.Bkt.CID, createUploadPartName(p.UploadID, p.Key, 0))
+	if err != nil {
+		return nil, err
+	}
+	if len(ids) == 0 {
+		return nil, errors.GetAPIError(errors.ErrNoSuchUpload)
+	}
+
+	meta, err := n.objectHead(ctx, p.Bkt.CID, ids[0])
```
KirillovDenis, line 443: I'm not sure we can return the first found object. The error seems to be more appropriate if we found more than one initial part.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,563 @@
...
+	}
+
+	uniqDirs := make(map[string]struct{})
+
+	for _, ov := range uploads {
+		if ov.IsDir {
+			if _, ok := uniqDirs[ov.Key]; !ok {
+				uniqDirs[ov.Key] = struct{}{}
+				result.Prefixes = append(result.Prefixes, ov.Key)
+			}
```
KirillovDenis, line 362: I suppose we should check directory uniqueness before truncate up to `MaxUploads`

KirillovDenis:
Could you update `docs/s3_test_results.md` and `docs/aws_s3_compat.md`?

api/layer/multipart_upload.go
masterSplinter01: I'm not sure we must return error here. If we lost an initial part but we have the rest parts, shouldn't we to try to complete upload anyway?

api/layer/multipart_upload.go
masterSplinter01: The case when we found two objects with the same upload id and name is odd, but okay

masterSplinter01:
I planned to do it after approve of PR.

KirillovDenis:
> I planned to update the tables after approve of PR.

Why? I think it would be nice to see updated test results before approve

masterSplinter01:
I thought it'd be sensibly if I update tables with results after finishing fixing all remarks and getting the latest version of functionality. But okay, updating the tables.

docs/aws_s3_compat.md
```diff
@@ -56,13 +56,13 @@ Should be supported soon.
...
 |    | Method                  | Comments                                       |
 |----|-------------------------|------------------------------------------------|
-| 🔵 | AbortMultipartUpload    |                                                |
-| 🔵 | CompleteMultipartUpload |                                                |
-| 🔵 | CreateMultipartUpload   | InitiateMultipartUpload and NewMultipartUpload |
-| 🔵 | ListMultipartUploads    |                                                |
-| 🔵 | ListParts               |                                                |
-| 🔵 | UploadPart              | PutObjectPart                                  |
-| 🔵 | UploadPartCopy          | CopyObjectPart                                 |
+| 🟢 | AbortMultipartUpload    |                                                |
```
realloc, line 59: 🎉

docs/aws_s3_compat.md
```diff
@@ -56,13 +56,13 @@ Should be supported soon.
 
 |    | Method                  | Comments                                       |
```
KirillovDenis, line 57: It's about beautifying but still.
```suggestion
|    | Method                  | Comments |
```

docs/s3_test_results.md
```diff
@@ -336,32 +336,33 @@ Compatibility:  0/29/30
...
+| 15 | s3tests_boto3.functional.test_s3.test_abort_multipart_upload                     | ok    | ok     |
+| 16 | s3tests_boto3.functional.test_s3.test_abort_multipart_upload_not_found           | ok    | ok     |
+| 17 | s3tests_boto3.functional.test_s3.test_list_multipart_upload                      | ok    | ok     |
+| 18 | s3tests_boto3.functional.test_s3.test_multipart_upload_missing_part              | ok    | ok     |
+| 19 | s3tests_boto3.functional.test_s3.test_multipart_upload_incorrect_etag            | ok    | ok     |
+| 20 | s3tests_boto3.functional.test_s3.test_multipart_resend_first_finishes_last       | ERROR | ERROR  |
+| 21 | s3tests_boto3.functional.test_s3.test_atomic_multipart_upload_write              | ok    | ok     |
+| 22 | s3tests_boto3.functional.test_s3.test_multipart_copy_versioned                   | ok    | ok     |
+Comments: in [PR](https://github.com/nspcc-dev/s3-tests/pull/5)
 ## Tagging
```
KirillovDenis, line 366:
```suggestion
Comments: in [PR](https://github.com/nspcc-dev/s3-tests/pull/5)

## Tagging
```

KirillovDenis: I mean new line between comments and tagging not between tests results and comments

docs/aws_s3_compat.md
```diff
@@ -54,15 +54,15 @@ For now there are some limitations:
...
-|    | Method                  | Comments                                       |
-|----|-------------------------|------------------------------------------------|
-| 🔵 | AbortMultipartUpload    |                                                |
-| 🔵 | CompleteMultipartUpload |                                                |
-| 🔵 | CreateMultipartUpload   | InitiateMultipartUpload and NewMultipartUpload |
-| 🔵 | ListMultipartUploads    |                                                |
-| 🔵 | ListParts               |                                                |
-| 🔵 | UploadPart              | PutObjectPart                                  |
-| 🔵 | UploadPartCopy          | CopyObjectPart                                 |
+|    | Method                  | Comments|
```
KirillovDenis, line 57: I mean:
```suggestion
|    | Method                  | Comments |
```

KirillovDenis:
We can `complete-multipart-upload` with empty part list. I'm not sure it's a correct behavior. AWS doesn't allow to do this with error: `An error occurred (MalformedXML) when calling the CompleteMultipartUpload operation (reached max retries: 1): The XML you provided was not well-formed or did not validate against our published schema`.

I used `cli` to upload one part and after that I invoked `complete-multipart-upload` with `--multipart-upload file://mpustruct`. Content of file `mpustruct`:
```
{
  "Parts": [
  ]
}
```
And I get success:
```
$ aws s3api complete-multipart-upload --endpoint http://127.0.0.1:8084  --bucket dkirillov2 --key mult1 --upload-id bd86067e-6d53-4967-8589-7340a79ea98b --multipart-upload file://mpustruct
{
    "Bucket": "dkirillov2",
    "Key": "mult1",
    "ETag": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
}
```

I suppose we should check the length of the part list.

masterSplinter01:
I supposed If someone sends request with empty parts xml, they makes it purposely. And there is no something about it in docs. But okay, will add  the checks.

api/layer/multipart_upload.go
```diff
@@ -0,0 +1,566 @@
...
+	objects, err := n.getUploadParts(ctx, p.Info)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(objects) == 1 {
+		obj, err := n.headLastVersionIfNotDeleted(ctx, p.Info.Bkt, p.Info.Key)
+		if err != nil {
+			return nil, err
+		}
```
KirillovDenis, line 190: I think we should check err and if it's `NoSuchKey` log it and return `ErrInvalidPart` (or somehow change error to return). It's a little confusing when you try to invoke `complete-multipart-upload` and get `NoSuchKey`.
People can think they provide wrong key (at least I thought so) though they just tried to complete upload without any uploaded parts in neofs.
```
$ aws s3api complete-multipart-upload --endpoint http://127.0.0.1:8084  --bucket dkirillov2 --key mult1 --upload-id bd86067e-6d53-4967-8589-7340a79ea98b --multipart-upload file://mpustruct
An error occurred (NoSuchKey) when calling the CompleteMultipartUpload operation (reached max retries: 0): The specified key does not exist.
```

AWS returns `InvalidPart` error as I understand

KirillovDenis (approved):

alexvanin (approved):

PR #284, opened 2021-11-22
[#219] Add container name resolving

### LINKED ISSUE #219: Unauthenticated aws s3 requests
We can only use bucket/containers which is owned by s3-gw itself If unauthenticated requests are used (e.g. `--no-sign-request` param in aws cli). It's a little bit odd since we couldn't work with public bucket that was created with authenticated request. The reason is using `owner id` in `pool.ListContainers` to get any bucket and this owner is differ in un/authenticated request. So we get `NoSuchBucket` error.

KirillovDenis:
This PR change bucket name resolving mechanism from listing owner's containers to using system DNS. But to test this system DNS must be configured correctly.
Should be considered with https://github.com/nspcc-dev/neofs-dev-env/pull/149
 
closes #219

### DIFF
--- a/api/layer/container.go
+++ b/api/layer/container.go
@@ -26,6 +26,8 @@ type (
 	}
 )
 
+const nnsNameAttr = "__NEOFS__NAME"
+
 func (n *layer) containerInfo(ctx context.Context, cid *cid.ID) (*data.BucketInfo, error) {
 	var (
 		err error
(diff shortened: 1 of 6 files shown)


realloc:
Later we could need an option with resolving order to let omit some methods. Like:
``` yaml
resolve_order:
  - dns
  - nns
  - other_option
```

alexvanin (approved):

Issue #287, opened 2021-12-02
Print version number at startup


alexvanin:
The same as https://github.com/nspcc-dev/neofs-http-gw/issues/109

Issue #282, opened 2021-11-18
Cannot set lifetime


aprasolova:
I try to set the lifetime from `help`, but `neofs-authmate` fails to parse it
```
$ neofs-authmate issue-secret --wallet wallet.json --lifetime 1d2h30m --peer 192.168.130.71:8080 --bearer-rules <some_rules> --gate-public-key <some_key> --container-policy '{"rep-3": "REP 3"}'
Incorrect Usage: invalid value "1d2h30m" for flag -lifetime: parse error

NAME:
   neofs-authmate issue-secret - Issue a secret in NeoFS network

USAGE:
   neofs-authmate issue-secret [command options] [arguments...]

OPTIONS:
   --wallet value                   path to the wallet
   --address value                  address of wallet account
   --peer value                     address of a neofs peer to connect to
   --bearer-rules value             rules for bearer token as plain json string
   --session-rules value            rules for session token as plain json string
   --gate-public-key value          public 256r1 key of a gate (use flags repeatedly for multiple gates)
   --container-id value             auth container id to put the secret into
   --container-friendly-name value  friendly name of auth container to put the secret into (default: "auth-container")
   --create-session-token           create session token (default: false)
   --lifetime value                 Lifetime of tokens (e.g. 1d2h30m, it will be ceil rounded to the nearest amount of epoch) (default: 720ns)
   --container-policy value         mapping AWS storage class to NeoFS storage policy as plain json string or path to json file
   --aws-cli-credentials value      path to the aws cli credential file
   --help, -h                       show help (default: false)
   
error: invalid value "1d2h30m" for flag -lifetime: parse error
```

Issue #274, opened 2021-10-06
Make system cache storing objects of different types.


masterSplinter01:
Now system cache stores object.Object, what is not siutable sometimes (for example, we have to do extra marshaling/unmarshaling of payload to a struct and vice versa, that not optimal).

PR #292, opened 2021-12-16
CHANGELOG: Add latest changes of v0.18.0 release

masterSplinter01:
Skipped:
#274 -- internal change, can be considered as a part of cors
#278 -- a part of cors
#282 -- fix of #258 which was added in this release
#287, #186 -- minor changes

Mentioned to #276 instead of #271 because the task changed and the PR gives more information than the issue.

Signed-off-by: Angira Kekteeva <kira@nspcc.ru>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 This document outlines major changes between releases.
 
+## [0.18.0] - 2021-12-16
+
+### Added
+- Support for MultipartUpload (#186, #187) 
+- CORS support (#217)
+- Authmate supports setting of tokens lifetime in a more convenient format (duration) (#258)
+- Generation of a random key for `--no-sign-request` (#276)
+
+### Changed
+- Bucket name resolving mechanism from listing owner's containers to using DNS (#219)
+
+### Removed
+- Deprecated golint, replaced by revive (#272)
+
 ## 0.17.0 (24 Sep 2021)
 With this release we introduce [ceph-based](https://github.com/ceph/s3-tests) S3 compatibility results.
 
@@ -107,3 +121,5 @@ Bugs fixed:
 Please refer to [Github
 releases](https://github.com/nspcc-dev/neofs-s3-gw/releases/) for older
 releases.
+
+[0.18.0]: https://github.com/nspcc-dev/neofs-s3-gw/compare/v0.17.0...v0.18.0


CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 This document outlines major changes between releases.
 
+## 0.18.0 (16 Dec 2021)
+
+### Added
+- Support of MultipartUpload (#186, #187)
```
realloc, line 8: Support for

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 This document outlines major changes between releases.
 
+## 0.18.0 (16 Dec 2021)
+
+### Added
+- Support of MultipartUpload (#186, #187) 
+- CORS support (#217)
+- Authmate supports setting of tokens' lifetime in more convenient format (duration) (#258)
+- Generating of a random key for `--no-sign-request` (#276)
```
realloc, line 11: Generating → Generation

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 This document outlines major changes between releases.
 
+## 0.18.0 (16 Dec 2021)
```
alexvanin, line 5: Use format from keepachangelog.com. Then Sami-chan will be able to parse it for auto release.
```suggestion
## [0.18.0]- 2021-12-16
```

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+
+### Added
+- Support for MultipartUpload (#186, #187) 
+- CORS support (#217)
+- Authmate supports setting of tokens' lifetime in more convenient format (duration) (#258)
+- Generation of a random key for `--no-sign-request` (#276)
+
+### Changed
+- Bucket name resolving mechanism from listing owners' container to using DNS (#219)
```
KirillovDenis, line 14: Maybe something like this would be better:
```suggestion
- Bucket name resolving mechanism from listing owner containers to using DNS (#219)
```

alexvanin (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 This document outlines major changes between releases.
 
+## 0.18.0 (16 Dec 2021)
+
+### Added
+- Support for MultipartUpload (#186, #187) 
+- CORS support (#217)
+- Authmate supports setting of tokens' lifetime in more convenient format (duration) (#258)
```
KirillovDenis, line 10: Maybe something like this would be better:
```suggestion
- Authmate supports setting of tokens lifetime in more convenient format (duration) (#258)
```

masterSplinter01: Why? We can put multiple tokens (bearer and session) and tokens' lifetime means 'lifetime of tokens'.

CHANGELOG.md
masterSplinter01: Typo, but I think it'd be better `owner's containers`

CHANGELOG.md
KirillovDenis: If I understand correctly `tokens lifetime` means the same as `lifetime of tokens`

EliChin: I don't thinks an apostrophe is necessary here. It's kinda hard to explain, but I don't feel like "tokens" own "lifetime". In the singular form, I would say `a token lifetime` without `'s`-identified possession. Thus, we don't need it in the plural form; so I would prefer: `Authmate supports setting of tokens lifetime in **a** more convenient format`. 
Though, if you doubt, you can always choose the `lifetime of tokens` form.
p.s. I also googled some phrases like "password lifetime" or "account lifetime" and saw no "'s".

alexvanin (approved):

PR #294, opened 2021-12-26
[#290] Add timeout to authmate

### LINKED ISSUE #290: An option to limit `neofs-authmate` hanging
It would be nice to have a `neofs-authmate` option which would limit the time the utility runs. I'd propose to add a `--await` option as it is done for `neofs-cli`.

masterSplinter01:
Closes #290 
Added timeout to `issue-secret` and `obtain-secret` commands. Default value is 1 minute, the value can be set with a flag `--timeout 10s/42m/100500h`.

Timeout option is common for both commands and, like `--with-log` and `--debug` must be placed before `issue-secret` and `obtain-secret`.

### DIFF
--- a/cmd/authmate/main.go
+++ b/cmd/authmate/main.go
@@ -49,6 +49,7 @@ var (
 	lifetimeFlag           time.Duration
 	containerPolicies      string
 	awcCliCredFile         string
+	timeoutFlag            time.Duration
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cmd/authmate/main.go
```diff
@@ -276,11 +284,20 @@ It will be ceil rounded to the nearest amount of epoch.`,
 				AwsCliCredentialsFile: awcCliCredFile,
 			}
 
-			if err = agent.IssueSecret(ctx, os.Stdout, issueSecretOptions); err != nil {
-				return cli.Exit(fmt.Sprintf("failed to issue secret: %s", err), 6)
-			}
+			errChan := make(chan error, 1)
+			go func() {
+				errChan <- agent.IssueSecret(ctx, os.Stdout, issueSecretOptions)
+			}()
```
KirillovDenis, line 290: Can we just create new context with timeout `ctx = context.WithTimeout(ctx, timeoutFlag)`?

alexvanin (approved):

KirillovDenis (approved):

PR #303, opened 2022-01-11
[#301] Implement GetBucketLocation

### LINKED ISSUE #301: GetBucketLocation should return policy name used at bucket creation
[GetBucketLocation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html) should return the policy name used at bucket creation time with `LocationConstraint` parameter. At creation time it has to be recorded as a container attribute and used later. To comply the S3 specification, it must be from the list of the valid AWS regions, but that's up to the user to set it in a compatible way or not.

KirillovDenis:
closes #301 

### DIFF
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -17,11 +17,12 @@ const (
 type (
 	// BucketInfo stores basic bucket data.
 	BucketInfo struct {
-		Name     string
-		CID      *cid.ID
-		Owner    *owner.ID
-		Created  time.Time
-		BasicACL uint32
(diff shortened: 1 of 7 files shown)


realloc (commented):
Can we also update test run results?

realloc (approved):
LGTM. Maybe, later we should use the container attributes array directly.

alexvanin (approved):

alexvanin:
Rebase, please.

PR #317, opened 2022-01-17
[#306] Handle empty post policy

### LINKED ISSUE #306: Incorrect empty policy handling
Cannot handle empty post policy. It seems we just forget use appropriate flag

So we need use it (in `checkPostPolicy` function):
```
policy := &postPolicy{empty: true}
if policyStr := auth.MultipartFormValue(r, "policy"); policyStr != "" {
	// ...
	policy.empty = false
}
```

KirillovDenis:
Add handling empty post policy
If this PR will be merged after https://github.com/nspcc-dev/neofs-sdk-go/pull/122 (and sdk version will be updated in s3-gw) we can update test results for:
* `test_s3.test_post_object_anonymous_request`
* `test_s3.test_post_object_set_success_code`

closes #306 

### DIFF
--- a/api/handler/put_test.go
+++ b/api/handler/put_test.go
@@ -2,9 +2,12 @@ package handler
 
 import (
 	"encoding/json"
+	"mime/multipart"
+	"net/http"
 	"testing"
 	"time"
 
+	"github.com/nspcc-dev/neofs-s3-gw/api"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


api/handler/put_test.go
```diff
@@ -86,3 +89,18 @@ func TestCustomJSONMarshal(t *testing.T) {
 
 	require.Equal(t, expectedPolicy, policy)
 }
+
```
alexvanin, line 92: Should we also replace `policy := &postPolicy{}` to `policy :=  &postPolicy{empty: true}` in line 86, even though it doesn't affect the test?

alexvanin (approved):

api/handler/put_test.go
KirillovDenis: I'm not sure because not empty policy is tested there.

realloc (approved):

Issue #320, opened 2022-01-17
Bucket not empty


KirillovDenis:
During tests run sometimes cleanup fails because of `An error occurred (BucketNotEmpty) when calling the DeleteBucket operation: The bucket you tried to delete is not empty`.
I think we currently consider presence some system objects in container as `BucketeNotEmpty` (we need check this) but we should not.

Can reproduce on `test_s3.test_versioning_bucket_multipart_upload_return_version_id` test

masterSplinter01:
The bug raises in cases when we try to delete the bucket with disabled or suspended versioning. 

The test makes a request `ListObjectVersions`, gets a list of objects with not null `VersionId == ObjectId`,  puts the `VersionId` in the next request `DeleteMultipleObjects`, and the gateway doesn't delete the objects because of receiving of incorrect `VersionId` (bucket versioning state is disabled/suspended). 

So, I think we need to make `ListObjectVersions` returning null `VersionId` for objects created in buckets with disabled/suspended versioning.

KirillovDenis:
It seems this issue related to #319

masterSplinter01:
Closed via #322

masterSplinter01:
Closed via #322

PR #322, opened 2022-01-18
[#319] Fix null versions

### LINKED ISSUE #319: Support "null" version deletion
We should support `null` version deletion when versioning is enabled 

Check the following tests:
* `test_s3.test_versioning_obj_plain_null_version_removal`
* `test_s3.test_versioning_obj_plain_null_version_overwrite`

KirillovDenis:
closes #319 

### DIFF
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -68,6 +68,15 @@ func (b *BucketInfo) CORSObjectName() string { return bktCORSConfigurationObject
 // Version returns object version from ObjectInfo.
 func (o *ObjectInfo) Version() string { return o.ID.String() }
 
+// NullableVersion returns object version from ObjectInfo.
+// Return "null" if "S3-Versions-unversioned" header present.
+func (o *ObjectInfo) NullableVersion() string {
+	if _, ok := o.Headers["S3-Versions-unversioned"]; ok {
+		return "null"
+	}
+	return o.Version()
+}
+
 // NiceName returns object name for cache.
 func (o *ObjectInfo) NiceName() string { return o.Bucket + "/" + o.Name }
 
--- a/docs/s3_test_results.md
+++ b/docs/s3_test_results.md
@@ -383,17 +383,17 @@ Compatibility: 9/6/8 out of 11
 
 ## Versioning
 
-Compatibility: 16/19/24 out of 26
+Compatibility: 22/19/24 out of 26
 
 |    | Test                                                                                        | s3-gw | minio | aws s3 |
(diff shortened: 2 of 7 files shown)


api/data/info.go
```diff
@@ -68,6 +68,15 @@ func (b *BucketInfo) CORSObjectName() string { return bktCORSConfigurationObject
 // Version returns object version from ObjectInfo.
 func (o *ObjectInfo) Version() string { return o.ID.String() }
 
+// NullableVersion returns object version from ObjectInfo.
+// Return "null" if "S3-Versions-unversioned" header present.
+func (o *ObjectInfo) NullableVersion() string {
```
masterSplinter01, line 73: I think, the `Nullable` prefix seems redundant a bit. 

As I see `Version` function is used only in `NullableVersion` and `TagsObject`. Maybe it'd be more clear to write something like this: 
```
// Version returns object version from ObjectInfo.
// Return "null" if "S3-Versions-unversioned" header present.
func (o *ObjectInfo) Version() string {
	if _, ok := o.Headers["S3-Versions-unversioned"]; ok {
		return "null"
	}
	return o.o.ID.String()
}
...
func (o *ObjectInfo) TagsObject() string { return ".tagset." + o.Name + "." + o.ID.String() }
```

KirillovDenis: In most cases we want to treat version as object id (for storing in headers and so on). `Version()` function is used not only in `NullableVesion` and `TagsObject` (we use it outside `data` package). But currently `NullableVersion` is used only once (to encode `list-object-versions` response) so we can define similar function there and drop it from this package

alexvanin: I think it is okay to keep it there, though. You may want to specify in the comment, that `NullableVersion` is required for listing object version.

alexvanin (approved):

docs/s3_test_results.md
```diff
@@ -383,17 +383,17 @@ Compatibility: 9/6/8 out of 11
 
 ## Versioning
 
-Compatibility: 16/19/24 out of 26
```
masterSplinter01, line 386: As I see this PR fixes the test `test_versioning_bucket_multipart_upload_return_version_id` and deletion of buckets (#320) too, maybe we should update test result here?

KirillovDenis: This test was fixed early https://github.com/nspcc-dev/neofs-s3-gw/pull/321

masterSplinter01: Okay, but anyway the PR closes #320 too.

PR #325, opened 2022-01-20
[#324] Make sure the default container name is unique

### LINKED ISSUE #324: Authmate default container name
When we run `authmate issue-secret` without specific container id, new one will be created. It we don't specify `--container-friendly-name` key the default name `auth-container` will be used. After #299 container friendly name became system name. But because of container name has to be unique we will get the following error (if someone already create container with default `auth-container` name):

`failed to issue secret: status: code = 1024 message = could not handle request: could not execute Put request: could not invoke method (putNamed): chain/client: contract execution finished with state FAULT; exception: at instruction 3509 (THROW): unhandled exception: "name is already taken"`

So we should use default container name that has format `auth-container-{random}` or `auth-container-{timestamp}`.

KirillovDenis:
closes #324 

### DIFF
--- a/authmate/authmate.go
+++ b/authmate/authmate.go
@@ -110,12 +110,19 @@ func (a *Agent) checkContainer(ctx context.Context, cid *cid.ID, friendlyName st
 		return nil, fmt.Errorf("failed to build placement policy: %w", err)
 	}
 
-	cnr := container.New(
+	cnrOptions := []container.Option{
 		container.WithPolicy(pp),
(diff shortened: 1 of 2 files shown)


masterSplinter01 (approved):

realloc (changes requested):
I'd suggest avoiding setting friendly names if the user didn't set it explicitly. That would cost more GAS and require more operations than accessing via CID later.

realloc (approved):
LGTM, didn't run

masterSplinter01 (approved):

PR #328, opened 2022-01-21
[#308] Correct status code

### LINKED ISSUE #308: Use 403 code instead of 500
Since errors kind of `SEARCH is denied by extended ACL check` means the key has not access to perform some operation we should use 403 error instead of 500

Check the following tests:
* `test_s3.test_object_copy_not_owned_bucket`
* `test_s3.test_object_anon_put`
* `test_s3.test_100_continue`

KirillovDenis:
As I understand currently we cannot know type of neofs error so in this PR `strings.Contais(...)` is used.
This works on neofs-nodes that have [this fix](https://github.com/nspcc-dev/neofs-node/issues/1098)

Test `test_s3.test_100_continue` now has problem with bucket ACL (new rules are appended to the end of eACL table so they have no effect because above them there are denial rules)

closes #308

### DIFF
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -307,7 +307,8 @@ func (n *layer) SessionOpt(ctx context.Context) pool.CallOption {
 // Get NeoFS Object by refs.Address (should be used by auth.Center).
 func (n *layer) Get(ctx context.Context, address *object.Address) (*object.Object, error) {
 	ops := new(client.GetObjectParams).WithAddress(address)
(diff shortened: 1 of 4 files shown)


alexvanin:
We will `string.Contains` after https://github.com/nspcc-dev/neofs-sdk-go/issues/125 one day.

alexvanin:
Also do not hesitate to create issues on API to replace string parsing errors to status code parsing. In the SDK there will be separate error structure that could be identified with `error.As`.

alexvanin (approved):

PR #334, opened 2022-01-26
[#329] Multiple container session tokens in auth box

### LINKED ISSUE #329: Support different container session token types in authbox
> I think it would be nice to be able to use the same credential for creating and deleting bucket. Maybe we should support it.

Agree with you. Session token intentionally doesn't have wildcard for the verbs, so the only choice there is to keep two or even three tokens for `PUT`, `DELETE` and `SETEACL` [operations](https://github.com/nspcc-dev/neofs-api/blob/d9d71ccd662dd9681f0fab51e647ea4864510225/session/types.proto#L50-L62).

_Originally posted by @alexvanin in https://github.com/nspcc-dev/neofs-s3-gw/issues/309#issuecomment-1018482341_

KirillovDenis:
Update protobuf structs to support multiple session tokens for container operations and use appropriate tokens for PUT, DELETE, SetEACL operations

closes #329

### DIFF
--- a/docs/authmate.md
+++ b/docs/authmate.md
@@ -128,13 +128,15 @@ it will be auto-generated with values:
 }
 ```
 
-Rules for a session token can be set via param `session-rules` (json-string and file path allowed), the default value is:
+Rules for session tokens can be set via param `session-rules` (json-string and file path allowed), the default value is:
 ```
-{
+[
+  {
     "verb": "PUT",
     "wildcard": true,
     "containerID": null
-}
+  }
+]
 ```
 
 If `session-rules` are set, but `create-session-token` is not, no session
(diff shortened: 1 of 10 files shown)


docs/authmate.md
```diff
@@ -128,13 +128,15 @@ it will be auto-generated with values:
 }
 ```
 
-Rules for a session token can be set via param `session-rules` (json-string and file path allowed), the default value is:
+Rules for a session tokens can be set via param `session-rules` (json-string and file path allowed), the default value is:
```
masterSplinter01, line 131: I suppose, indefinite article should be removed before plural nouns

masterSplinter01 (approved):

alexvanin (approved):
I would rather specify all three verbs in the default session tokens, but we can discuss it in separate issue.

PR #339, opened 2022-01-31
[#338] Remove redundant flag for session tokens in authmate

### LINKED ISSUE #338: Change behaviour of authmate for session tokens.
For creating and putting session tokens we have two params: `--create-session-token` and `--session-rules`. 

When only the first parameter is set it means "create session tokens with default rules", 
when both of them are set it means "create session token with custom rules",
when only the second parameter is set, authmate prints `Warning: rules for session token were set but --create-session flag wasn't, so session token was not created` and creates a box with bearer token but without session tokes. It seems annoying, because most likely the user wanted to create session tokens but forgot to write the flag `--create-session-token`. 

Also two params seem redundant, it'd be enough to have only `--create-session-token` with string value (' ' means default value) and drop `--session-rules`.

masterSplinter01:
Closes #338

### DIFF
--- a/cmd/authmate/main.go
+++ b/cmd/authmate/main.go
@@ -36,7 +36,6 @@ var (
 	accountAddressFlag     string
 	peerAddressFlag        string
 	eaclRulesFlag          string
-	contextRulesFlag       string
 	gateWalletPathFlag     string
 	gateAccountAddressFlag string
 	accessKeyIDFlag        string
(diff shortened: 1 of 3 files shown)


cmd/authmate/main.go
```diff
@@ -45,7 +44,7 @@ var (
 	gatesPublicKeysFlag    cli.StringSlice
 	logEnabledFlag         bool
 	logDebugEnabledFlag    bool
-	sessionTokenFlag       bool
+	sessionTokenFlag       string
```
realloc, line 47: It's not a flag anymore.

PR #345, opened 2022-02-06
[#343] Replace default policy in authmate

### LINKED ISSUE #343: Default storage policy for auth container
It seems that we use [default](https://github.com/nspcc-dev/neofs-s3-gw/blob/204835ace35cabb508de470fcbaeec43e19e4a3a/authmate/authmate.go#L108) container policy for auth container.

Default policy is 
```
REP 1 
CBF 1 
SELECT 2 FROM *
```

I suggest to change it to the policy with at least two replicas. Otherwise it doesn't make much sense in case of storage failures.
```
REP 2 
CBF 3 
SELECT 2 * FROM *
```

masterSplinter01:
Closes #343

### DIFF
--- a/authmate/authmate.go
+++ b/authmate/authmate.go
@@ -327,13 +327,13 @@ func buildPlacementPolicy(placementRules string) (*netmap.PlacementPolicy, error
 	}
 
 	/*
-		REP 1 IN X 			  // place one copy of object
-		CBF 1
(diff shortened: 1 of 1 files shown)


realloc (approved):
Maybe, REP 3 CBF 2 would work better.

KirillovDenis (approved):

PR #346, opened 2022-02-08
Upgrade NeoFS SDK Go library

### LINKED ISSUE #349: Refactor dependency from `pool.Pool`
In current implementation all packages which need communication with NeoFS depend on `pool.Pool`. Because of this, there are difficulties with its redefinition for tests. I propose to define NeoFS network interfaces locally in packages and implement them through `pool.Pool` only on app-side.

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/135
* closes #349


creds/tokens/credentials.go
```diff
@@ -77,37 +62,48 @@ func (c *cred) GetBox(ctx context.Context, address *object.Address) (*accessbox.
...
-		ctx,
-		ops,
-	)
-	if err != nil {
-		return nil, err
+	// read payload
+	var data []byte
+
+	if sz := res.Header.PayloadSize(); sz > 0 {
+		if sz <= 4096 { // configure?
```
alexvanin, line 85: It looks like minor optimization. Suggest to `io.ReadAll()` in all cases. 

By the way, why not `data = make([]byte, sz)` in all cases as well?

cmd/s3-gw/app.go
```diff
@@ -106,11 +106,11 @@ func newApp(ctx context.Context, l *zap.Logger, v *viper.Viper) *App {
...
-		Key:                     &key.PrivateKey,
-		NodeConnectionTimeout:   conTimeout,
-		NodeRequestTimeout:      reqTimeout,
-		ClientRebalanceInterval: reBalance,
-		SessionExpirationEpoch:  math.MaxUint64,
+		Key:                       &key.PrivateKey,
+		NodeConnectionTimeout:     conTimeout,
+		NodeRequestTimeout:        reqTimeout,
+		ClientRebalanceInterval:   reBalance,
+		SessionExpirationDuration: math.MaxUint64,
```
alexvanin, line 113: See https://github.com/nspcc-dev/neofs-http-gw/pull/126#discussion_r815780802

cmd/authmate/main.go
```diff
@@ -421,10 +421,10 @@ func createSDKClient(ctx context.Context, log *zap.Logger, key *ecdsa.PrivateKey
...
 
 	opts := &pool.BuilderOptions{
-		Key:                    key,
-		NodeConnectionTimeout:  poolConnectTimeout,
-		NodeRequestTimeout:     poolRequestTimeout,
-		SessionExpirationEpoch: math.MaxUint32,
+		Key:                       key,
+		NodeConnectionTimeout:     poolConnectTimeout,
+		NodeRequestTimeout:        poolRequestTimeout,
+		SessionExpirationDuration: math.MaxUint32,
```
alexvanin, line 427: See https://github.com/nspcc-dev/neofs-http-gw/pull/126#discussion_r815780802

alexvanin:
Test of object versioning are very important, so we need to solve https://github.com/nspcc-dev/neofs-s3-gw/issues/349 first.

creds/tokens/credentials.go
cthulhu-rider: We limit buffer size when we copy payload to some `io.Writer`.

internal/neofs/neofs.go
```diff
@@ -0,0 +1,615 @@
...
+			return nil, fmt.Errorf("read object list: %w", err)
+		}
+	}
+
+	return buf[:read], nil
+}
+
+//
+// // ReadObjectPayload implements tokens.NeoFS interface method.
+// func (x *NeoFS) ReadObjectPayload(ctx context.Context, addr address.Address) ([]byte, error) {
```
KirillovDenis, line 385: Should be dropped

KirillovDenis:
I think this PR should update sdk up to version that contains fix for https://github.com/nspcc-dev/neofs-sdk-go/issues/149. Otherwise, some s3-compatibility tests fail

api/layer/layer.go
```diff
@@ -294,27 +485,34 @@ func IsAuthenticatedRequest(ctx context.Context) bool {
...
 	return []pool.CallOption{pool.WithKey(&n.anonKey.Key.PrivateKey)}
 }
 
-// Get NeoFS Object by refs.Address (should be used by auth.Center).
-func (n *layer) Get(ctx context.Context, address *object.Address) (*object.Object, error) {
-	ops := new(client.GetObjectParams).WithAddress(address)
-	obj, err := n.pool.GetObject(ctx, ops, n.CallOptions(ctx)...)
-	return obj, n.transformNeofsError(ctx, err)
+// Get NeoFS Object by address (should be used by auth.Center).
+func (n *layer) Get(ctx context.Context, addr *address.Address) (*object.Object, error) {
```
KirillovDenis, line 514: It seems this method is unused

api/layer/object.go
```diff
@@ -71,70 +66,101 @@ type (
...
+	return n.transformNeofsError(ctx, err)
 }
 
 // objectGet returns an object with payload in the object.
-func (n *layer) objectGet(ctx context.Context, cid *cid.ID, oid *object.ID) (*object.Object, error) {
-	ops := new(client.GetObjectParams).WithAddress(newAddress(cid, oid))
-	obj, err := n.pool.GetObject(ctx, ops, n.CallOptions(ctx)...)
-	return obj, n.transformNeofsError(ctx, err)
-}
+func (n *layer) objectGet(ctx context.Context, addr *address.Address) (*object.Object, error) {
```
KirillovDenis, line 148: It's expected this method to return `object.Object` with not empty payload field but as I can see we return only headers

api/layer/system_object.go
```diff
@@ -68,53 +66,41 @@ func (n *layer) putSystemObjectIntoNeoFS(ctx context.Context, p *PutSystemObject
...
 		return nil, n.transformNeofsError(ctx, err)
 	}
 
-	meta, err := n.objectHead(ctx, p.BktInfo.CID, oid)
+	meta, err := n.objectHead(ctx, p.BktInfo.CID, id)
 	if err != nil {
 		return nil, err
 	}
 
+	idsToDeleteArr := updateCRDT2PSetHeaders(p.Metadata, versions, false) // false means "last write wins"
```
KirillovDenis, line 102: This line must be before forming attributes because `updateCRDT2PSetHeaders` method update `p.Metadata`

api/layer/object.go
```diff
@@ -71,70 +66,101 @@ type (
...
-// objectGetWithPayloadWriter and write it into provided io.Reader.
-func (n *layer) objectGetWithPayloadWriter(ctx context.Context, p *getParams) (*object.Object, error) {
-	// prepare length/offset writer
-	w := newWriter(p.Writer, p.offset, p.length)
-	ops := new(client.GetObjectParams).WithAddress(newAddress(p.cid, p.oid)).WithPayloadWriter(w)
-	obj, err := n.pool.GetObject(ctx, ops, n.CallOptions(ctx)...)
-	return obj, n.transformNeofsError(ctx, err)
+// writes payload part of the NeoFS object to the provided io.Writer.
+// Zero range corresponds to full payload (panics if only offset is set).
+func (n *layer) objectWritePayload(ctx context.Context, p getParams) error {
```
KirillovDenis, line 119: I think this name can be a little confused. It would be nice to mention that this is `read` operation. But it's a matter of taste

api/layer/multipart_upload.go
```diff
@@ -260,12 +260,15 @@ func (n *layer) CompleteMultipartUpload(ctx context.Context, p *CompleteMultipar
...
+
 	for _, part := range parts {
-		_, err := n.objectGetWithPayloadWriter(ctx, &getParams{
-			Writer: pw,
-			cid:    p.Info.Bkt.CID,
-			oid:    part.ID,
-		})
+		prmGet.oid = part.ID
+
+		err = n.objectWritePayload(ctx, prmGet)
```
KirillovDenis, line 271: We can use object reader and drop `io.Pipe`

cmd/s3-gw/app.go
```diff
@@ -157,26 +158,30 @@ func newApp(ctx context.Context, l *zap.Logger, v *viper.Viper) *App {
 		NotificationController: nc,
 	}
 
+	var n neofs.NeoFS
+	n.SetConnectionPool(conns)
+
```
KirillovDenis, line 163: Can we use `neofs.NeoFS` from line 123?

api/layer/layer.go
KirillovDenis: Yes

api/layer/object.go
cthulhu-rider: `NeoFS.ReadObject` is used correctly, but implementation doesn't follow the requirement. Fixed.

api/layer/multipart_upload.go
```diff
@@ -169,6 +171,45 @@ func (n *layer) UploadPartCopy(ctx context.Context, p *UploadCopyParams) (*data.
...
+
+	curReader io.Reader
+
+	parts []*data.ObjectInfo
+}
+
+func (x *multiObjectReader) Read(p []byte) (n int, err error) {
+	if x.curReader != nil {
+		n, err = x.curReader.Read(p)
+		if err == nil || err != nil && !stderrors.Is(err, io.EOF) {
```
KirillovDenis, line 190: Condition can be simplified to `if err == nil || !stderrors.Is(err, io.EOF)` or even to `if !stderrors.Is(err, io.EOF)` (though checking to nil is a little optimization)

cmd/authmate/main.go
```diff
@@ -414,17 +414,24 @@ func obtainSecret() *cli.Command {
...
+		NodeRequestTimeout:    poolRequestTimeout,
 	}
-	return pb.Build(ctx, opts)
+	p, err := pb.Build(ctx, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	var neoFS neofs.AuthmateNeoFS
+	neoFS.SetConnectionPool(p)
```
KirillovDenis, line 434: It cause panic because neoFS is nil

KirillovDenis:
> I think this PR should update sdk up to version that contains fix for [nspcc-dev/neofs-sdk-go#149](https://github.com/nspcc-dev/neofs-sdk-go/issues/149). Otherwise, some s3-compatibility tests fail

But such updating leads to `could not verify body: e68c9f917f9e3016df5c324bceaf6a93af0791e6f461ba96be3da02f2dfae19c : c80f94640140cdbbc8b76cec1f0727cacbf4c8508214a5a0b58ab501a1acd5a7: invalid signature` errors. 
It seems we have to adapt solution from https://github.com/nspcc-dev/neofs-http-gw/pull/126

cmd/authmate/main.go
cthulhu-rider: Fixed.

KirillovDenis:
Rebase, please

KirillovDenis (approved):

PR #350, opened 2022-02-14
[#347] Add permission check of session tokens

### LINKED ISSUE #347: The gateway hangs when a user tries to create a bucket with session token without SetEACL permission
When a user with session token with only PUT permission tries to create a bucket, the GW does the following steps: 
1. puts container successfully
2. tries to put eacl, but receives the error:
```
could not handle request: could not execute GetEACL request: extended ACL table is not set for this container
```
and returns 500 Internal Error to the user. 

Then AWS CLI retries to create the bucket again and receives `name is already taken` error because containers with that name is already created (see p. 1)

masterSplinter01:
Closes #347

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -549,7 +549,13 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
 	boxData, err := layer.GetBoxData(r.Context())
 	if err == nil {
 		policies = boxData.Policies
-		p.SessionToken = boxData.Gate.SessionTokenForPut()
+		p.SessionTokenForPut = boxData.Gate.SessionTokenForPut()
+		p.SessionTokenForSetEACL = boxData.Gate.SessionTokenForSetEACL()
+	}
+
+	if p.SessionTokenForPut == nil || p.SessionTokenForSetEACL == nil {
+		h.logAndSendError(w, "couldn't find session token for put or seteacl", reqInfo, errors.GetAPIError(errors.ErrAccessDenied))
+		return
 	}
 
 	if createParams.LocationConstraint != "" {
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
+	"github.com/nspcc-dev/neofs-sdk-go/session"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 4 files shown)


api/layer/container.go
```diff
@@ -169,16 +169,11 @@ func (n *layer) createContainer(ctx context.Context, p *CreateBucketParams) (*ci
 	return bktInfo.CID, nil
 }
 
-func (n *layer) setContainerEACLTable(ctx context.Context, cid *cid.ID, table *eacl.Table) error {
+func (n *layer) setContainerEACLTable(ctx context.Context, cid *cid.ID, table *eacl.Table, sessionToken *session.Token) error {
 	table.SetCID(cid)
+	table.SetSessionToken(sessionToken)
```
KirillovDenis, line 174: It seems we can accept only `*eacl.Table` that already has session token inside

api/layer/layer.go
```diff
@@ -351,7 +352,13 @@ func (n *layer) PutBucketACL(ctx context.Context, param *PutBucketACLParams) err
 		return err
 	}
 
-	return n.setContainerEACLTable(ctx, inf.CID, param.EACL)
+	var sessionToken *session.Token
+	boxData, err := GetBoxData(ctx)
+	if err == nil {
+		sessionToken = boxData.Gate.SessionTokenForSetEACL()
+	}
```
KirillovDenis, line 359: Can we move this to the handlers level (retrieve session token in functions like `CreateBucketHandler`)?

api/layer/container.go
masterSplinter01: I decided to require session token as a separated argument to avoid forgetting to attach the token to `eacl.Table`  before calling of `setContainerEACLTable` in case of possible further use. 
But it's just a matter of taste.

api/layer/layer.go
masterSplinter01: I'm not sure, because the function PutBucketACL is called in `PostObject`, `PutObjectHandler` and  `updateBucketACL` which, in turn, we call in `PutBucketACLHandler`, `PutObjectACLHandler`, `PutBucketPolicyHandler` and `PutObjectACLHandler`.  I think moving of this code will produce a lot of copy-paste. 

Also, is it nice, that we use `Client` interface function in our "inner" function `updateBucketACL`?

KirillovDenis: > I think moving this code to the handlers level will produce a lot of copy-paste.

 We can define the separate function to this. The reason why I consider such moving is checking session token before we we start action (e.g. put object) that require SETEAСL token.

> Also, is it nice, that we use Client interface function in our "inner" function updateBucketACL?

Actually, it's not "inner" function but common part in some handlers (`PutBucketPolicyHandler`, `CompleteMultipartUploadHandler` etc.)

api/handler/put.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
+	"github.com/nspcc-dev/neofs-sdk-go/session"
```
alexvanin, line 9: GitHub package definitions are down below.

PR #351, opened 2022-02-14
[#348] Remove prefix to errors in authmate

### LINKED ISSUE #348: Slightly not correct message of authmate if jthere is an error in session rules json
If input json with session token rules is not correct, e.g:
```
[
    {
      "verb": "PUT",
      "wildcard": true,
      "containerID": null
    },
    {
      "verb": "SETEACL",
      "wildcard": kek,
      "containerID": null
    }
] 
```
Authmate prints: 
```
failed to issue secret: failed to build bearer token: failed to build context for session token: failed to unmarshal rules for session token: invalid character 'k' looking for beginning of value
```
I suppose we should delete a part with bearer token from the message

masterSplinter01:
Closes #348

### DIFF
--- a/authmate/authmate.go
+++ b/authmate/authmate.go
@@ -242,7 +242,7 @@ func (a *Agent) IssueSecret(ctx context.Context, w io.Writer, options *IssueSecr
 
 	gatesData, err := createTokens(options, lifetime, cid)
 	if err != nil {
-		return fmt.Errorf("failed to build bearer token: %w", err)
+		return err
 	}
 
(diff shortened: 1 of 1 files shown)


realloc (approved):

KirillovDenis (approved):
I would add error wrapping to [this line](https://github.com/nspcc-dev/neofs-s3-gw/blob/master/authmate/authmate.go#L483) to prevent theoretically confusing errors from [this function](https://github.com/nspcc-dev/neofs-s3-gw/blob/master/authmate/authmate.go#L443)

PR #356, opened 2022-02-17
[#340] Notification handlers

### LINKED ISSUE #340: Implement put / get bucket notification configuration
To enable notifications in the bucket, S3 gateway should support [PutBucketNotificationConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketNotificationConfiguration.html) and [GetBucketNotificationConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketNotificationConfiguration.html).

S3 notification configuration contains:
- TopicConfiguration for SNS
- QueueConfiguration for SQS
- CloudFunctionConfiguration for Lambda

We should support **only QueueConfiguration records**. With integration of NATS, we want to use decentralized queue, therefore SQS configuration will be suitable. Return error if  `PubBucketNotificationConfiguration` request contains SNS or Lambda configurations.

Store configuration in the system object in the specified bucket. Return this configuration on `GetBucketNotificationConfiguration` request.

masterSplinter01:
Closes #340

### DIFF
--- a/api/data/notifications.go
+++ b/api/data/notifications.go
@@ -1,0 +1,69 @@
+package data
+
+type (
+	NotificationConfiguration struct {
+		QueueConfigurations []QueueConfiguration `xml:"QueueConfiguration" json:"QueueConfigurations"`
+		// Not supported topics
+		TopicConfigurations          []TopicConfiguration          `xml:"TopicConfiguration" json:"TopicConfigurations"`
+		LambdaFunctionConfigurations []LambdaFunctionConfiguration `xml:"CloudFunctionConfiguration" json:"CloudFunctionConfigurations"`
+	}
+
+	QueueConfiguration struct {
+		ID       string   `xml:"Id" json:"Id"`
+		QueueArn string   `xml:"Queue" json:"Queue"`
+		Events   []string `xml:"Event" json:"Events"`
+		Filter   Filter   `xml:"Filter" json:"Filter"`
+	}
+
+	Filter struct {
+		Key Key `xml:"S3Key" json:"S3Key"`
+	}
+
+	Key struct {
+		FilterRules []FilterRule `xml:"FilterRule" json:"FilterRules"`
+	}
+
+	FilterRule struct {
+		Name  string `xml:"Name" json:"Name"`
+		Value string `xml:"Value" json:"Value"`
+	}
+
+	// TopicConfiguration and LambdaFunctionConfiguration -- we don't support these configurations
+	// but we need them to detect in notification configurations in requests.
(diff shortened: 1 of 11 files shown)


api/layer/layer.go
```diff
@@ -257,13 +262,14 @@ func DefaultCachesConfigs() *CachesConfig {
 
 // NewLayer creates instance of layer. It checks credentials
 // and establishes gRPC connection with node.
-func NewLayer(log *zap.Logger, conns pool.Pool, config *Config) Client {
+func NewLayer(log *zap.Logger, conns pool.Pool, config *Config, ncontroller *notifications.Controller) Client {
```
KirillovDenis, line 265: I think we can add `*notifications.Controller` to config param

api/layer/notifications.go
```diff
@@ -0,0 +1,115 @@
...
+		conf      = &data.NotificationConfiguration{}
+		completed bool
+		err       error
+	)
+
+	if err = xml.NewDecoder(tee).Decode(conf); err != nil {
+		return errors.GetAPIError(errors.ErrMalformedXML)
+	}
+
+	if conf != nil {
```
KirillovDenis, line 38: Condition is always true

api/data/notifications.go
```diff
@@ -0,0 +1,163 @@
...
+			return errors.GetAPIError(errors.ErrEventNotification)
+		}
+	}
+	return nil
+}
+
+func (n *NotificationConfiguration) CheckAndComplete() (bool, error) {
+	completed := false
+	if len(n.TopicConfigurations) != 0 {
+		for i, t := range n.TopicConfigurations {
```
KirillovDenis, line 125: Since we [support only](https://github.com/nspcc-dev/neofs-s3-gw/issues/340#issue-1120508149) `QueueConfiguration` we should return error.

api/layer/notifications.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return conf, nil
+}
+
+func (n *layer) checkAndCompleteNotificationConfiguration(c *data.NotificationConfiguration) (completed bool, err error) {
+	if c.TopicConfigurations != nil || c.LambdaFunctionConfigurations != nil {
+		return completed, errors.GetAPIError(errors.ErrNotificationTopicNotSupported)
+	}
+	if c != nil {
```
KirillovDenis, line 120: It's an odd condition because three line above we already access to `c`

KirillovDenis (approved):

cthulhu-rider (approved):

PR #363, opened 2022-02-28
[#195] object locking

### LINKED ISSUE #195: Support object locking
Should we support [object locking](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html)?

There are some problems:
* Where will the flag (that enable this functionality when bucket creating) be stored?
* Where will the object lock configuration be stored?

The similar problems must be solved in #122 so it issue should be done later anyway.

KirillovDenis:
The following handlers are implemented:
* GetObjectLegalHold       
* GetObjectLockConfiguration 
* GetObjectRetention     
* PutObjectLegalHold      
* PutObjectLockConfiguration 
* PutObjectRetention 

closes #195


api/notifications/controller.go
```diff
@@ -52,5 +63,56 @@ func NewController(p *Options) (*Controller, error) {
...
+	for _, stream := range c.handlers {
+		go func(stream Stream) {
+			for {
+				select {
+				case msg := <-stream.ch:
+					fmt.Printf("got message: %s\n", msg.Data)
+					if err := stream.h.HandleMessage(ctx, msg); err != nil {
+						fmt.Printf("could not handle message: %s", err)
+					} else if err = msg.Ack(); err != nil {
+						fmt.Printf("could not ACK message: %s", err)
```
alexvanin, line 110: Those prints look like debug log output. Do we keep them?

api/layer/layer.go
```diff
@@ -472,6 +280,23 @@ func (n *layer) EphemeralKey() *keys.PublicKey {
 	return n.anonKey.Key.PublicKey()
 }
 
+func (n *layer) Initialize(ctx context.Context, c Notificator) error {
```
alexvanin, line 283: Was it necessary to provide `Notificator` directly into `Initialize()` instead of keeping it in config? 
It seems that in the end you still do `n.ncontroller = c`.

Is it possible to call `Initialize()` twice during runtime?

alexvanin (approved):

api/layer/layer.go
KirillovDenis: Actually this function isn't necessary for current PR (it was written when I thought we have to handle ticks for locking objects). But I would like to preserve this function since another [PR](https://github.com/nspcc-dev/neofs-s3-gw/pull/371) is based on it.

The function `Initialize()` should be called once (at least until we support adding/removing handlers for ticks in runtime). That's why I provide `Notificator` directly into this function - I wanted to avoid introducing new flag in `layer` structure that allows to know if the layer is initialized.

masterSplinter01 (approved):

PR #370, opened 2022-03-03
[#366] Add wait for container removed

### LINKED ISSUE #366: Bucket removal can behave incorrectly
In current implementation bucket removal produces container removal https://github.com/nspcc-dev/neofs-s3-gw/blob/44548212858db1d13a27c9b89cfe55041cde1e00/api/layer/container.go#L241.
This operation is not synchronous. I didn't catch any implementation requirements in docs, so may be we need to make sync operation and await for the actual removal (like in bucket creation).

KirillovDenis:
closes #366

### DIFF
--- a/api/layer/container.go
+++ b/api/layer/container.go
@@ -163,8 +163,8 @@ func (n *layer) createContainer(ctx context.Context, p *CreateBucketParams) (*ci
 	return bktInfo.CID, nil
 }
 
-func (n *layer) setContainerEACLTable(ctx context.Context, cid *cid.ID, table *eacl.Table) error {
-	table.SetCID(cid)
+func (n *layer) setContainerEACLTable(ctx context.Context, idCnr *cid.ID, table *eacl.Table) error {
+	table.SetCID(idCnr)
 
 	boxData, err := GetBoxData(ctx)
 	if err == nil {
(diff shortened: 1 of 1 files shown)


api/layer/container.go
```diff
@@ -207,6 +207,43 @@ func (n *layer) waitEACLPresence(ctx context.Context, cid *cid.ID, table *eacl.T
...
+		return err
+	}
+
+	return n.waitForContainerRemoved(ctx, cid, defaultWaitParams())
+}
+
+func (n *layer) waitForContainerRemoved(ctx context.Context, cid *cid.ID, params *waitParams) error {
+	return waitFor(ctx, params, func(ctx context.Context) bool {
+		_, err := n.pool.GetContainer(ctx, cid)
+		if err != nil && strings.Contains(err.Error(), "container not found") {
```
cthulhu-rider, line 239: Lets add `TODO: (neofs-s3-gw#367) handle NeoFS API status error`?

Also we can just `return` this statement w/o `if` block.

api/layer/container.go
```diff
@@ -207,6 +207,43 @@ func (n *layer) waitEACLPresence(ctx context.Context, cid *cid.ID, table *eacl.T
...
+	}
+
+	if err = n.pool.DeleteContainer(ctx, cid, pool.WithSession(sessionToken)); err != nil {
+		return err
+	}
+
+	return n.waitForContainerRemoved(ctx, cid, defaultWaitParams())
+}
+
+func (n *layer) waitForContainerRemoved(ctx context.Context, cid *cid.ID, params *waitParams) error {
```
cthulhu-rider, line 236: Minor suggestion: rename `cid` parameter in order to no collide with package name (in `deleteContainer` too).

api/layer/container.go
```diff
@@ -207,6 +207,43 @@ func (n *layer) waitEACLPresence(ctx context.Context, cid *cid.ID, table *eacl.T
...
+	return waitFor(ctx, params, func(ctx context.Context) bool {
+		_, err := n.pool.GetContainer(ctx, cid)
+		if err != nil && strings.Contains(err.Error(), "container not found") {
+			return true
+		}
+		return false
+	})
+}
+
+func waitFor(ctx context.Context, params *waitParams, fn func(context.Context) bool) error {
```
cthulhu-rider, line 246: Lets clarify in docs that it `waits for fn's true return`?

api/layer/container.go
KirillovDenis: Currently, it doesn't work. Do we need use neofs-node version other than v0.27.5?

cthulhu-rider: I know, meant just to add the TODO comment.

PR #371, opened 2022-03-09
[#192] Support expiration lifecycle

### LINKED ISSUE #192: Support Expiration Lifecycle
Should we support lifecycle? https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html
I suppose there are some restriction (if we want to support it):
* Support only `Expiration actions` (available `Expiration actions` and `Transition actions`) because s3-gw can not handle storage classes (or it will be?)
* `Expiration actions` will only work in the context of the gateway, and the gateway must store the user's credentials

KirillovDenis:
must be closed after #363 

closes #192

### DIFF
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -57,8 +57,9 @@ type (
 
 	// BucketSettings stores settings such as versioning.
 	BucketSettings struct {
-		Versioning        string                   `json:"versioning"`
-		LockConfiguration *ObjectLockConfiguration `json:"lock_configuration"`
+		Versioning             string                   `json:"versioning"`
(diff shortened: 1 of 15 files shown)


KirillovDenis:
We should rebase and open this PR into `tree-service`

smallhive:
The tree service is going to die and it will be completely replaced. According to comments on the issue some questions are not solved. We may leave this PR as a reference, if we need this functionality.

roman-khimov:
OK, let's close the PR, but leave the issue (which is valid) here.

PR #379, opened 2022-03-21
[#347] Add setEACL session token checks

### LINKED ISSUE #347: The gateway hangs when a user tries to create a bucket with session token without SetEACL permission

masterSplinter01:
Closes #347

### DIFF
--- a/api/handler/multipart_upload.go
+++ b/api/handler/multipart_upload.go
@@ -14,6 +14,7 @@ import (
 	"github.com/nspcc-dev/neofs-s3-gw/api/data"
 	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
 	"github.com/nspcc-dev/neofs-s3-gw/api/layer"
+	"github.com/nspcc-dev/neofs-sdk-go/session"
 	"go.uber.org/zap"
 )
 
@@ -345,6 +346,8 @@ func (h *handler) CompleteMultipartUploadHandler(w http.ResponseWriter, r *http.
 	}
 
 	var (
+		sessionTokenSetEACL *session.Token
+
 		uploadID   = r.URL.Query().Get(uploadIDHeaderName)
 		uploadInfo = &layer.UploadInfoParams{
 			UploadID: uploadID,
@@ -352,6 +355,7 @@ func (h *handler) CompleteMultipartUploadHandler(w http.ResponseWriter, r *http.
 			Key:      reqInfo.ObjectName,
 		}
 		additional = []zap.Field{zap.String("uploadID", uploadID), zap.String("Key", reqInfo.ObjectName)}
+		uploadData = &UploadData{}
 	)
 
 	reqBody := new(CompleteMultipartUpload)
@@ -371,16 +375,6 @@ func (h *handler) CompleteMultipartUploadHandler(w http.ResponseWriter, r *http.
 		return
 	}
 
-	c := &layer.CompleteMultipartParams{
-		Info:  uploadInfo,
-		Parts: reqBody.Parts,
-	}
-	objInfo, err := h.obj.CompleteMultipartUpload(r.Context(), c)
-	if err != nil {
-		h.logAndSendError(w, "could not complete multipart upload", reqInfo, err, additional...)
-		return
-	}
-
 	if initPart.Size > 0 {
 		initPartPayload := bytes.NewBuffer(make([]byte, 0, initPart.Size))
 		p := &layer.GetObjectParams{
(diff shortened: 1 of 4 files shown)


alexvanin (approved):

api/layer/layer.go
```diff
@@ -330,17 +330,19 @@ type (
...
-		Policy             *netmap.PlacementPolicy
-		EACL               *eacl.Table
-		SessionToken       *session.Token
-		LocationConstraint string
+		Name                   string
+		ACL                    uint32
+		Policy                 *netmap.PlacementPolicy
+		EACL                   *eacl.Table
+		SessionTokenForPut     *session.Token
+		SessionTokenForSetEACL *session.Token
```
KirillovDenis, line 338: I would drop `SessionTokenForSetEACL` param and use only `EACL` field

api/layer/layer.go
```diff
@@ -330,17 +330,19 @@ type (
...
+		LocationConstraint     string
 	}
 	// PutBucketACLParams stores put bucket acl request parameters.
 	PutBucketACLParams struct {
-		Name string
-		EACL *eacl.Table
+		Name         string
+		EACL         *eacl.Table
+		SessionToken *session.Token
```
KirillovDenis, line 345: I would drop `SessionToken` param and use only `EACL` field

alexvanin (approved):

api/handler/multipart_upload.go
```diff
@@ -398,31 +392,48 @@ func (h *handler) CompleteMultipartUploadHandler(w http.ResponseWriter, r *http.
...
-			}
-			if err = h.updateBucketACL(r, astObject, bktInfo); err != nil {
-				h.logAndSendError(w, "could not update bucket acl while completing multipart upload", reqInfo, err, additional...)
-				return
-			}
+	c := &layer.CompleteMultipartParams{
+		Info:  uploadInfo,
+		Parts: reqBody.Parts,
+	}
+	objInfo, err := h.obj.CompleteMultipartUpload(r.Context(), c)
```
KirillovDenis, line 407: Why did we move this invocation?

api/handler/util.go
```diff
@@ -80,3 +81,17 @@ func parseRange(s string) (*layer.RangeParams, error) {
...
 	}, nil
 }
+
+func getSessionTokenSetEACL(ctx context.Context) (*session.Token, error) {
+	var sessionToken *session.Token
+	boxData, err := layer.GetBoxData(ctx)
+	if err != nil {
+		return nil, err
+	}
+	sessionToken = boxData.Gate.SessionTokenForSetEACL()
```
KirillovDenis, line 91:
```suggestion
	boxData, err := layer.GetBoxData(ctx)
	if err != nil {
		return nil, err
	}
	sessionToken := boxData.Gate.SessionTokenForSetEACL()
```

api/handler/multipart_upload.go
```diff
@@ -398,31 +392,48 @@ func (h *handler) CompleteMultipartUploadHandler(w http.ResponseWriter, r *http.
...
+			ObjectInfo: objInfo,
+			TagSet:     uploadData.TagSet,
+		}
+		if err = h.obj.PutObjectTagging(r.Context(), t); err != nil {
+			h.logAndSendError(w, "could not put tagging file of completed multipart upload", reqInfo, err, additional...)
+			return
+		}
+	}
+
+	if uploadData.ACL != nil {
```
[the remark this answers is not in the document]
masterSplinter01: Lost while rebase, removed `uploadData` from line 358. Now it works.

api/handler/multipart_upload.go
masterSplinter01: Before we used to make the following steps:
1. Get init part info
2. Complete multipart upload
3. If init part has payload, get init part, unmarshal payload, put tagging and/or `ACL` file

If a user don't have a session token, step 3 fails, but an object is created. Thus, now we the following steps: 
1. Get init part info
2. If init part has payload and the payload contains `ACL`, check presence of `EACL` token.
3. Complete multipart upload
4. Put tagging and/or `ACL` file

PR #380, opened 2022-03-22
cmd/s3-gw: Refactor logger

cthulhu-rider:
* see https://github.com/nspcc-dev/neofs-sdk-go/issues/177

Notable changes:
- logger writes to stderr instead of stdout
- app name and version are always presented and fixed
- all user options except `level` are abolished because they are not yet clear to the one who launches the application process

IMO we should provide config options to the user only if they are perceptible to him, and he understands the regulation of performance.


KirillovDenis:
Please make sure you get rid of any [neofs-sdk-go/logger](https://github.com/nspcc-dev/neofs-s3-gw/blob/master/api/layer/versioning_test.go#L20) encountered

KirillovDenis (approved):

KirillovDenis:
Rebase, please

cmd/s3-gw/main.go
```diff
@@ -2,42 +2,63 @@ package main
...
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("build zap logger instance: %v", err))
 	}
 
-	return l
+	return l.With(
+		zap.String("app_name", "neofs-s3-gw"),
+		zap.String("app_version", version.Version),
+	)
```
KirillovDenis, line 61: Can we log the application name and version once when the application starts?

cthulhu-rider: Yes, of course.

PR #385, opened 2022-03-25
[#384] Add check of Notificator interface value in Initialize func

### LINKED ISSUE #384: S3 GW panics when notifications are not enabled
The app panics in `layer/layer.go:294` when calls `c.Listen(ctx)` because Notificator interface value `c` is `nil`

masterSplinter01:
Closes #384

### DIFF
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -281,10 +281,6 @@ func (n *layer) EphemeralKey() *keys.PublicKey {
 }
 
 func (n *layer) Initialize(ctx context.Context, c Notificator) error {
-	if c == nil {
-		return nil
-	}
-
 	if n.IsNotificationEnabled() {
 		return fmt.Errorf("already initialized")
 	}
--- a/api/notifications/controller.go
+++ b/api/notifications/controller.go
@@ -37,10 +37,6 @@ type Stream struct {
 }
 
 func NewController(p *Options, l *zap.Logger) (*Controller, error) {
-	if p == nil {
-		return nil, nil
-	}
-
 	ncopts := []nats.Option{
 		nats.Timeout(p.Timeout),
 	}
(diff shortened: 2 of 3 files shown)


api/layer/layer.go
```diff
@@ -281,7 +282,7 @@ func (n *layer) EphemeralKey() *keys.PublicKey {
 }
 
 func (n *layer) Initialize(ctx context.Context, c Notificator) error {
-	if c == nil {
+	if c == nil || reflect.ValueOf(c).Kind() != reflect.Struct && reflect.ValueOf(c).IsNil() {
```
alexvanin, line 285: Can we avoid reflection until it is absolutely necessary?
We can cast interface to exact structure and check nil value there, can't we?

KirillovDenis:
Actually, I don't like use `reflect`. Can we do refactor to avoid this.
For example we can use something like (during call `Initialize` method):
```
	if v.GetBool(cfgEnableNATS) {
		nopts := getNotificationsOptions(v, l)
		nc, err := notifications.NewController(nopts, l)
		if err != nil {
			l.Fatal("failed to enable notifications", zap.Error(err))
		}
		if err = obj.Initialize(ctx, nc); err != nil {
			l.Fatal("couldn't initialize layer", zap.Error(err))
		}
	}
```

api/layer/layer.go
KirillovDenis: Then we need know exactly structure type/s. It seems that using an interface as a parameter makes no sense in this case

masterSplinter01: Why we should avoid using of reflect?

KirillovDenis:
We can now drop the following condition in `getNotificationsOptions` function :
```
if enabled := v.GetBool(cfgEnableNATS); !enabled {
	return nil
}
```

Also we can drop the following condition in `NewController` function:
```
if p == nil {
	return nil, nil
}
```

api/layer/layer.go
KirillovDenis: It's like a marker that our code does something unsafe and dirty though we actually don't do this. It also makes the code less readable.

KirillovDenis (approved):

PR #392, opened 2022-04-04
Fixed handlers for failed tests

### LINKED ISSUE #393: Fix tests
### Has been ok before become fail or error:

* test_multi_object_delete -- [PR](https://github.com/nspcc-dev/neofs-s3-gw/pull/392)
* test_multi_objectv2_delete -- [PR](https://github.com/nspcc-dev/neofs-s3-gw/pull/392)
* **s3tests_boto3.functional.test_s3.test_multi_object_delete_key_limit**
* **test_object_anon_put_write_access**
* s3tests_boto3.functional.test_s3.test_bucket_create_naming_good_long_63 ... ERROR -- [PR](https://github.com/nspcc-dev/neofs-contract/pull/238)
* test_bucket_get_location -- not correct test, depends on `api_name` in config
* test_bucket_create_naming_dns_long -- [PR](https://github.com/nspcc-dev/neofs-contract/pull/238)
* s3tests_boto3.functional.test_s3.test_bucket_recreate_overwrite_acl ... FAIL -- comment below
* s3tests_boto3.functional.test_s3.test_bucket_recreate_new_acl ... FAIL -- comment below
* s3tests_boto3.functional.test_s3.test_object_acl_full_control_verify_attributes ... ERROR -- fails on `api/handler/acl.go:762` -- tries to form acl record with `id` based on `user_id` from `.conf` file (default `testid`)
* s3tests_boto3.functional.test_s3.test_bucket_create_naming_good_contains_period ... ERROR --  `could not create bucket {"request_id": "23f60168-4584-4333-8d2a-64b54f8e7abe", "method": "CreateBucket", "bucket_name": "test45sxe5hgw9vg2qwfo8zx2h8w9naaa.111", "object_name": "", "error": "save container via connection pool: status: code = 1024 message = chain/client: contract execution finished with state FAULT; exception: at instruction 1167 (THROW): unhandled exception: \"invalid domain name format\""}` -- s3 gw doesn't allow to create a bucket with a name with a dot '.'
* **s3tests_boto3.functional.test_s3.test_object_copy_not_owned_bucket ... FAIL** 

### Problem because the gw has not unversioned mode:
Tests create a bucket, request versioning status of the bucket and expect to get None, but receive `Suspended` because S3 GW hasn't unversioned mode. If we remove the primary check, the test passes.
* s3tests_boto3.functional.test_s3.test_versioning_bucket_create_suspend ... FAIL 
* s3tests_boto3.functional.test_s3.test_versioning_obj_plain_null_version_removal ... FAIL
* s3tests_boto3.functional.test_s3.test_versioning_obj_plain_null_version_overwrite ... FAIL
* s3tests_boto3.functional.test_s3.test_versioning_obj_plain_null_version_overwrite_suspended ... FAIL

### Concurrency (data race in tests?)
* s3tests_boto3.functional.test_s3.test_versioned_concurrent_object_create_concurrent_remove ... FAIL
* s3tests_boto3.functional.test_s3.test_versioned_concurrent_object_create_and_remove ... FAIL

### Problems with copying in versioned bucket:

* s3tests_boto3.functional.test_s3.test_object_copy_versioned_bucket ...  -- copying in the same bucket with versioning throws ` "error": "read object list: status: code = 1024 message = malformed request"`
* s3tests_boto3.functional.test_s3.test_object_copy_versioned_url_encoding ... ERROR -- copying in the same bucket with versioning throws ` "error": "read object list: status: code = 1024 message = malformed request"`
* s3tests_boto3.functional.test_s3.test_object_copy_versioning_multipart_upload ... ERROR -- the same
* s3tests_boto3.functional.test_s3.test_multipart_copy_versioned ... FAIL
* s3tests_boto3.functional.test_s3.test_versioning_copy_obj_version ... ERROR

### Has been error become fail:

* test_object_set_get_unicode_metadata 
* test_put_object_ifnonmatch_failed 
* test_put_object_ifnonmatch_overwrite_existed_failed
* test_object_raw_get_object_acl 
* test_object_raw_get_x_amz_expires_not_expired 

### Other:

* s3tests_boto3.functional.test_s3.test_list_multipart_upload_owner ... FAIL -- missing in the table: assert `'NbUgTSFvPmsRxmGeWpuuGeJUoRoi6PErcM' != 'testid'` fails

* s3tests_boto3.functional.test_s3.test_versioning_obj_create_overwrite_multipart ... FAIL, creates multipart upload with the same key in versioning bucket multiple times, gets object payload and compares with expected payload. Fails on 2nd multipart upload, receives payload of previous version.

### Has been error/fail become ok :D 

s3tests_boto3.functional.test_s3.test_sse_kms_multipart_upload ... ok -- become from error to ok lol
s3tests_boto3.functional.test_s3.test_sse_kms_multipart_invalid_chunks_1 ... ok
s3tests_boto3.functional.test_s3.test_sse_kms_multipart_invalid_chunks_2 ... ok
s3tests_boto3.functional.test_s3.test_sse_kms_transfer_1b ... ok
s3tests_boto3.functional.test_s3.test_sse_kms_transfer_1kb ... ok
s3tests_boto3.functional.test_s3.test_sse_kms_transfer_1MB ... ok
s3tests_boto3.functional.test_s3.test_sse_kms_transfer_13b ... ok

masterSplinter01:
Closes #393

### DIFF
--- a/api/handler/delete.go
+++ b/api/handler/delete.go
@@ -213,19 +213,7 @@ func (h *handler) DeleteMultipleObjectsHandler(w http.ResponseWriter, r *http.Re
 
 	var errs []error
 	for _, obj := range deletedObjects {
-		if obj.Error == nil {
-			deletedObj := DeletedObject{
-				ObjectIdentifier: ObjectIdentifier{
-					ObjectName: obj.Name,
-					VersionID:  obj.VersionID,
-				},
-				DeleteMarkerVersionID: obj.DeleteMarkVersion,
-			}
-			if deletedObj.DeleteMarkerVersionID != "" {
-				deletedObj.DeleteMarker = true
-			}
-			response.DeletedObjects = append(response.DeletedObjects, deletedObj)
-		} else if !requested.Quiet {
+		if obj.Error != nil {
 			code := "BadRequest"
 			if s3err, ok := obj.Error.(errors.Error); ok {
 				code = s3err.Code
@@ -237,15 +225,26 @@ func (h *handler) DeleteMultipleObjectsHandler(w http.ResponseWriter, r *http.Re
 				VersionID: obj.VersionID,
 			})
 			errs = append(errs, obj.Error)
+		} else if !requested.Quiet {
+			deletedObj := DeletedObject{
+				ObjectIdentifier: ObjectIdentifier{
+					ObjectName: obj.Name,
+					VersionID:  obj.VersionID,
+				},
+				DeleteMarkerVersionID: obj.DeleteMarkVersion,
+			}
+			if deletedObj.DeleteMarkerVersionID != "" {
+				deletedObj.DeleteMarker = true
+			}
+			response.DeletedObjects = append(response.DeletedObjects, deletedObj)
 		}
 	}
-
-	if !requested.Quiet && len(errs) != 0 {
-		additional := []zap.Field{
+	if len(errs) != 0 {
+		fields := []zap.Field{
 			zap.Array("objects", marshaler),
 			zap.Errors("errors", errs),
 		}
-		h.logAndSendError(w, "could not delete objects", reqInfo, nil, additional...)
+		h.log.Error("couldn't delete objects", fields...)
 	}
 
 	if err := api.EncodeToResponse(w, response); err != nil {
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -568,7 +568,6 @@ func (n *layer) deleteObject(ctx context.Context, bkt *data.BucketInfo, settings
 		}
 		last := versions.getLast(FromUnversioned())
 		if last == nil {
-			obj.Error = errors.GetAPIError(errors.ErrInvalidVersion)
 			return obj
 		}
 		p.Header[VersionsDeleteMarkAttr] = last.Version()


api/handler/delete.go
```diff
@@ -213,7 +213,10 @@ func (h *handler) DeleteMultipleObjectsHandler(w http.ResponseWriter, r *http.Re
 
 	var errs []error
 	for _, obj := range deletedObjects {
-		if obj.Error == nil {
+		if obj.Error == nil || (len(obj.VersionID) == 0 && errors.IsS3Error(obj.Error, errors.ErrInvalidVersion)) {
```
masterSplinter01, line 216: When we try to delete not existing objects without specified version:
- aws responses without any error in any mode: quiet or verbose 
- the gate responses with an error `invalid version is specified`
- tests requires no error in response

That's why I decided to consider `errors.ErrInvalidVersion` as noError if `versionID` is not specified in a request.

api/handler/delete.go
```diff
@@ -213,7 +213,10 @@ func (h *handler) DeleteMultipleObjectsHandler(w http.ResponseWriter, r *http.Re
...
+			if requested.Quiet {
+				continue
+			}
```
masterSplinter01, line 219: If quiet mode was chosen in a request we don't add successfull object deletion to our response.

api/handler/delete.go
```diff
@@ -240,14 +243,6 @@ func (h *handler) DeleteMultipleObjectsHandler(w http.ResponseWriter, r *http.Re
...
 	}
 
-	if !requested.Quiet && len(errs) != 0 {
-		additional := []zap.Field{
-			zap.Array("objects", marshaler),
-			zap.Errors("errors", errs),
-		}
-		h.logAndSendError(w, "could not delete objects", reqInfo, nil, additional...)
-	}
-
```
masterSplinter01, line 250: I'm not sure do we really need this codeblock because we anyway response with `DeleteObjectsResponse` struct

api/handler/delete.go
KirillovDenis: Maybe we can return no error instead of `errors.ErrInvalidVersion` from `layer` in this case?

api/handler/delete.go
KirillovDenis: It's need to log errors. Otherwise we will see errors only as response on the client side. But we can simplify condition to `if len(errs) != 0`

api/handler/delete.go
masterSplinter01: Yep, I thought about it, but I think we need to check how changing of logics will influence in other functions which use `deleteObjects` function. Maybe later after fixing other tests.

api/handler/delete.go
```diff
@@ -225,7 +228,7 @@ func (h *handler) DeleteMultipleObjectsHandler(w http.ResponseWriter, r *http.Re
 				deletedObj.DeleteMarker = true
 			}
 			response.DeletedObjects = append(response.DeletedObjects, deletedObj)
-		} else if !requested.Quiet {
+		} else {
```
KirillovDenis, line 231: I suppose  have  the following condition:
```
if obj.Error != nil {
    // fill `response.Errors`
} else if !requested.Quite {
    // fill `response.DeletedObjects`
}
```

masterSplinter01: But we need to consider case when `len(obj.VersionID) == 0 && errors.IsS3Error(obj.Error, errors.ErrInvalidVersion)` as not an error in both modes `quite` and `verbose`. That's why I think, first of all, we should go to branch `obj.Error == nil`.

KirillovDenis: https://github.com/nspcc-dev/neofs-s3-gw/pull/392#discussion_r841702616

KirillovDenis (approved):

alexvanin (approved):

PR #394, opened 2022-04-06
Fix grantee

masterSplinter01:


api/handler/acl.go
```diff
@@ -158,17 +159,31 @@ func (h *handler) gateKey(ctx context.Context) (*keys.PublicKey, error) {
 	return gateKey, nil
 }
 
+func (h *handler) bearerTokenIssuerKey(ctx context.Context) (*keys.PublicKey, error) {
```
KirillovDenis, line 162: I suppose we can use hex encoded key value rather than `*keys.PublicKey`

PR #395, opened 2022-04-06
[#395] Fix grantee in ACL

masterSplinter01:
Signed-off-by: Angira Kekteeva <kira@nspcc.ru>

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"context"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"encoding/hex"
 	"encoding/json"
 	"encoding/xml"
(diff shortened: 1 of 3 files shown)


api/handler/acl.go
```diff
@@ -158,17 +159,31 @@ func (h *handler) gateKey(ctx context.Context) (*keys.PublicKey, error) {
...
+
+	return key, nil
+}
+
 func (h *handler) PutBucketACLHandler(w http.ResponseWriter, r *http.Request) {
 	reqInfo := api.GetReqInfo(r.Context())
-	gateKey, err := h.gateKey(r.Context())
+	key, err := h.bearerTokenIssuerKey(r.Context())
 	if err != nil {
 		h.logAndSendError(w, "couldn't get gate key", reqInfo, err)
```
KirillovDenis, line 180: It's not a gate key anymore

api/handler/acl.go
```diff
@@ -158,17 +159,31 @@ func (h *handler) gateKey(ctx context.Context) (*keys.PublicKey, error) {
 	return gateKey, nil
 }
 
+func (h *handler) bearerTokenIssuerKey(ctx context.Context) (*keys.PublicKey, error) {
```
KirillovDenis, line 162: I suppose we can use hex encoded key value rather than `*keys.PublicKey`. I just don't like explicit `elliptic.P256()` mention.
Though we use not only hex encoded key value but also address of it. But it's used as display name and only for one grantee (there is not consistency). So we can omit it.

Any thoughts, @alexvanin

alexvanin: > I just don't like explicit elliptic.P256() mention.

Accessbox, neofs-node and neo-go widely use this explicit curve declaration, so we are tightly connected with NIST P256 anyway. At least we have nice `key.Address()`

alexvanin (approved):
LGTM, see https://github.com/nspcc-dev/neofs-s3-gw/pull/395/files#r843789199

We can merge in into master and rebase `pilot` with force push. Thoughts?

PR #405, opened 2022-04-21
English Check

EliChin:
Checked for some inaccuracies regarding English.

Signed-off-by: Elizaveta Chichindaeva <elizaveta@nspcc.ru>

### DIFF
--- a/api/handler/api.go
+++ b/api/handler/api.go
@@ -16,14 +16,14 @@ type (
 		cfg *Config
 	}
 
-	// Config contains data which handler need to keep.
+	// Config contains data which handler needs to keep.
 	Config struct {
 		DefaultPolicy *netmap.PlacementPolicy
 		DefaultMaxAge int
 	}
 )
 
-// DefaultPolicy is a default policy of placing container in NeoFS if it's not set at the request.
+// DefaultPolicy is a default policy of placing containers in NeoFS if it's not set at the request.
 const DefaultPolicy = "REP 3"
 
 var _ api.Handler = (*handler)(nil)
--- a/api/handler/delete.go
+++ b/api/handler/delete.go
@@ -15,7 +15,7 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-// DeleteObjectsRequest - xml carrying the object key names which needs to be deleted.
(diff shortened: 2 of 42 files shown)


KirillovDenis:
Please use the commit format `[#XX] ...`

api/handler/delete.go
```diff
@@ -15,21 +15,21 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-// DeleteObjectsRequest - xml carrying the object key names which needs to be deleted.
+// DeleteObjectsRequest -- xml carrying the object key names which nshould be deleted.
```
alexvanin, line 18:
```suggestion
// DeleteObjectsRequest -- xml carrying the object key names which should be deleted.
```

api/handler/locking_test.go
```diff
@@ -493,7 +493,7 @@ func TestObjectLegalHold(t *testing.T) {
 	hc.Handler().GetObjectLegalHoldHandler(w, r)
 	assertLegalHold(t, w, legalHoldOff)
 
-	// to make sure put hold is idempotent operation
+	// to make sure put hold is ans idempotent operation
```
alexvanin, line 496:
```suggestion
	// to make sure put hold is an idempotent operation
```

api/handler/api.go
```diff
@@ -16,14 +16,14 @@ type (
...
-	// Config contains data which handler need to keep.
+	// Config contains data which handler needs to keep.
 	Config struct {
 		DefaultPolicy *netmap.PlacementPolicy
 		DefaultMaxAge int
 	}
 )
 
-// DefaultPolicy is a default policy of placing container in NeoFS if it's not set at the request.
+// DefaultPolicy is a default policy of placing containers in NeoFS if it's not set otherwise at the request.
```
alexvanin, line 26: Is `otherwise` suitable there? Looks a bit odd to me for some reason.

The context is that `Policy` may be omitted in the request. If so, then use `DefaultPolicy`.

EliChin: You are right, it's unnecessary here. For some reason I've put `otherwise` to emphasize that something is set not like in the default.

alexvanin (approved):

masterSplinter01 (approved):

PR #412, opened 2022-04-27
Fix links in docs

masterSplinter01:
Goland processed incorrect links in docs as correct links, but now I detected that links are actually incorrect


alexvanin (approved):

KirillovDenis (approved):
It's a matter of taste but maybe we can use relative links in lower case (e.g `#generation-of-wallet`)

PR #436, opened 2022-05-16
Disable homomorphic hash if possible

### LINKED ISSUE #435: Disable homomorphic encryption in produced containers
Authmate and s3 gateway do not work with storage groups, therefore they can disable homomorphic encryption in all produced NeoFS containers. So I propose to disable homomorphic encryption by default for 100% of cases. I think it is better than having opt-in setting in configs. Thoughts? /cc @realloc @anatoly-bogatyrev 

## How to
In all produced containers set attribute `__NEOFS__DISABLE_HOMOMORPHIC_HASHING` to `true`.

alexvanin:
Closes #435

### DIFF
--- a/internal/neofs/neofs.go
+++ b/internal/neofs/neofs.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
 	objectv2 "github.com/nspcc-dev/neofs-api-go/v2/object"
 	"github.com/nspcc-dev/neofs-s3-gw/api/layer/neofs"
 	"github.com/nspcc-dev/neofs-s3-gw/authmate"
@@ -134,6 +135,16 @@ func (x *NeoFS) CreateContainer(ctx context.Context, prm neofs.PrmContainerCreat
 
 	for _, attr := range prm.AdditionalAttributes {
 		cnrOptions = append(cnrOptions, container.WithAttribute(attr[0], attr[1]))
+	}
+
+	// https://github.com/nspcc-dev/neofs-s3-gw/issues/435
+	// environment without hh disabling feature will ignore this attribute
+	// environment with hh disabling feature will set disabling = true if network config says so
+	if hhDisabled, err := isHomomorphicHashDisabled(ctx, x.pool); err != nil {
+		return nil, err
+	} else if hhDisabled {
+		cnrOptions = append(cnrOptions, container.WithAttribute(
+			"__NEOFS__DISABLE_HOMOMORPHIC_HASHING", "true"))
 	}
 
 	cnr := container.New(cnrOptions...)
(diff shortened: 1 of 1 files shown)


KirillovDenis (approved):

cthulhu-rider (approved):
Code is correct. Pls note that if network configuration does not allow hash disabling, containers are going to be denied.

alexvanin:
> Pls note that if network configuration does not allow hash disabling, containers are going to be denied.

Oups, didn't expect that. I thought that `__NEOFS__DISABLE_HOMOMORPHIC_HASHING=true` will be ignored (as it is ignored in the environment that didn't support disabling at all). I guess this PR is incorrect right now.

cthulhu-rider:
* @alexvanin see https://github.com/nspcc-dev/neofs-node/issues/1300.
Implementation: https://github.com/nspcc-dev/neofs-node/commit/a137c87c5979aa62370dac031bfb65117a0a67cc.

alexvanin:
Bucket creation happens much rarer than object creation, so I decided to check network config on each container creation request. Network configuration value can be cached at the startup but it makes code more sluggish, don't think it worth it now.

internal/neofs/neofs.go
```diff
@@ -136,6 +136,18 @@ func (x *NeoFS) CreateContainer(ctx context.Context, prm neofs.PrmContainerCreat
...
+	// https://github.com/nspcc-dev/neofs-s3-gw/issues/435
+	// environment without hh disabling feature will ignore this attribute
+	// environment with hh disabling feature will set proper value from network config
+	hhDisabledValue := "false"
+	if hhDisabled, err := isHomomorphicHashDisabled(ctx, x.pool); err != nil {
+		return nil, err
+	} else if hhDisabled {
+		hhDisabledValue = "true"
+	}
+	cnrOptions = append(cnrOptions, container.WithAttribute(
```
cthulhu-rider, line 148: Correct, but notice that if there is no requirement to add an attribute on S3-side, we can add it only `if hhDisabled` since missing attribute is equivalent to false.

internal/neofs/neofs.go
```diff
@@ -575,3 +587,25 @@ func (x *AuthmateNeoFS) CreateObject(ctx context.Context, prm tokens.PrmObjectCr
...
+	if err != nil {
+		return false, err
+	}
+
+	var (
+		expectedParamKey   = []byte("HomomorphicHashingDisabled")
+		expectedParamValue = []byte{1}
+	)
+
+	ni.NetworkConfig().IterateParameters(func(p *netmap.NetworkParameter) bool {
```
cthulhu-rider, line 602: You can also see how it's implemented in SDK branch https://github.com/nspcc-dev/neofs-sdk-go/blob/be747ff11edd6c59a47d3b5d2b221e5754ac11e1/container/network.go#L51.
Note that there are still no PR in NeoFS API with attribute format, but I guess we'll require `true` string.

internal/neofs/neofs.go
alexvanin: That's nice!

cthulhu-rider (approved):

PR #443, opened 2022-05-20
[#416] Use tree service to list objects

### LINKED ISSUE #416: Use the tree service for ListObjects[v2]
We can use [tree service](https://neospcc.atlassian.net/wiki/spaces/NEOF/pages/2298314753/Object+Tree+Service) to find objects after #407 will be merged.

We should change the `SEARCH` operation (during `ListObjects` request) to:

1. `GetNodeByPath` with flag (only last version) to get node that corresponds to the prefix
2. `GetSubTree` to get nodes (if delimiter is `/` we can use depth 1 to avoid getting unnecessary nodes)
3. Sort node on every level by id and `OID` attribute to get latest version
4. HEAD the latest objects from NeoFS

**Note:** There is some problem with different delimiters. As I understand when node is added to the tree its filepath delimiter is defined (implicitly) and cannot be changed. So we have to decide what to do in the following cases (if we want to use tree service for list objects with non default delimiters):

Suppose we put two objects (`az/bz/c` and `az/b`) into s3, so we have the following tree struct:
* {"id": 0}
    *  {"id": 1, "FileName": "az"}
        * {"id": 2, "FileName": "b"}
        * {"id": 3, "FileName": "bz"}
            * {"id": 4, "FileName": "c"}
            
So now we want to list object with prefix `az/b` and delimiter `z`. It seems in such case we have to:
1. Split prefix by `/` then use as path `az` (in `GetNodeByPath`)
2. `GetSubTree` with max depth (or sequentially invoke `GetSubTree` until all nodes will have appropriate delimiter `z` in their `FileName`)
3. Construct for every node full path (in our case it will be two node `az/bz/c` and `az/b`), trim appropriate prefix `az/b` and finally split by delilmiter `z` 

/cc @alexvanin @realloc

KirillovDenis:
closes #416

The basic list tests are passed:
```
s3tests_boto3.functional.test_s3.test_bucket_list_delimiter_alt ... ok
s3tests_boto3.functional.test_s3.test_bucket_listv2_delimiter_alt ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_delimiter_prefix_underscore ... ok
s3tests_boto3.functional.test_s3.test_bucket_listv2_delimiter_prefix_underscore ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_delimiter_percentage ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_delimiter_basic ... ok
s3tests_boto3.functional.test_s3.test_bucket_listv2_delimiter_basic ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_delimiter_prefix ... ok
s3tests_boto3.functional.test_s3.test_bucket_listv2_delimiter_prefix ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_empty ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_distinct ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_many ... ok
s3tests_boto3.functional.test_s3.test_bucket_listv2_many ... ok
s3tests_boto3.functional.test_s3.test_bucket_listv2_encoding_basic ... ok
s3tests_boto3.functional.test_s3.test_bucket_list_encoding_basic ... ok
```

### DIFF
--- a/internal/neofs/tree.go
+++ b/internal/neofs/tree.go
@@ -60,8 +60,6 @@ const (
 	maxGetSubTreeDepth = 10 // current limit on storage node side
 )
 
-var emptyOID oid.ID
-
 // NewTreeClient creates instance of TreeClient using provided address and create grpc connection.
 func NewTreeClient(addr string, key *keys.PrivateKey) (*TreeClient, error) {
 	conn, err := grpc.Dial(addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
@@ -308,6 +306,8 @@ func (c *TreeClient) getSubTreeLatestVersions(ctx context.Context, cnrID *cid.ID
 	if err != nil {
 		return nil, err
 	}
+
+	var emptyOID oid.ID
 
 	latestVersions := make(map[string]*TreeNode, len(subTree))
 	for _, node := range subTree {
--- a/internal/neofs/tree.go
+++ b/internal/neofs/tree.go
@@ -56,7 +56,11 @@ const (
 	systemTree = "system"
 
 	separator = "/"
+
+	maxGetSubTreeDepth = 10 // current limit on storage node side
 )
+
+var emptyOID oid.ID
 
 // NewTreeClient creates instance of TreeClient using provided address and create grpc connection.
 func NewTreeClient(addr string, key *keys.PrivateKey) (*TreeClient, error) {
@@ -74,7 +78,13 @@ func NewTreeClient(addr string, key *keys.PrivateKey) (*TreeClient, error) {
 	}, nil
 }
 
-func newTreeNode(nodeInfo *tree.GetNodeByPathResponse_Info) (*TreeNode, error) {
+type NodeResponse interface {
+	GetMeta() []*tree.KeyValue
+	GetNodeId() uint64
+	GetTimestamp() uint64
+}
+
+func newTreeNode(nodeInfo NodeResponse) (*TreeNode, error) {
 	var objID oid.ID
 	meta := make(map[string]string, len(nodeInfo.GetMeta()))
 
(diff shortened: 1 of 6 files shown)


internal/neofs/tree.go
```diff
@@ -242,6 +252,96 @@ func (c *TreeClient) GetLatestVersion(ctx context.Context, cnrID *cid.ID, object
...
+func (c *TreeClient) getSubTreeLatestVersions(ctx context.Context, cnrID *cid.ID, nodeID uint64) ([]oid.ID, error) {
+	subTree, err := c.getSubTree(ctx, cnrID, versionTree, nodeID, maxGetSubTreeDepth)
+	if err != nil {
+		return nil, err
+	}
+
+	latestVersions := make(map[string]*TreeNode, len(subTree))
+	for _, node := range subTree {
+		treeNode, err := newTreeNode(node)
+		if err != nil || treeNode.ObjID.Equals(emptyOID) { // invalid OID attribute
```
masterSplinter01, line 315: I'm not sure we need to keep empty id as a global variable, maybe it'd be better to replace it by `oid.ID{}`?

internal/neofs/tree.go
```diff
@@ -242,6 +252,96 @@ func (c *TreeClient) GetLatestVersion(ctx context.Context, cnrID *cid.ID, object
...
+
+	subTree, err := c.getSubTree(ctx, cnrID, versionTree, rootID, 1)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []oid.ID
+	for _, node := range subTree {
+		if node.GetNodeId() != 0 && hasPrefix(node, tailPrefix) {
+			latestNodes, err := c.getSubTreeLatestVersions(ctx, cnrID, node.GetNodeId())
```
masterSplinter01, line 285: I suppose we can drop `latestNodes` and assign directly to `result`.

KirillovDenis: Could you explain what do you mean? The `result` is slice of `[]oid.ID` but `getSubTreeLatestVersions` returns slice of `*data.NodeVersion`. Besides, we have to append to the `results` rather than replace it (we cannot do this because of the functions returns `error` as second value)

internal/neofs/tree.go
KirillovDenis: It seems we can allocate 32 byte for empty `oid` once. I'll make `emptyOID` local variable.

internal/neofs/tree.go
masterSplinter01: Sorry, you're right

alexvanin (approved):

PR #445, opened 2022-05-20
[#420] List versions using the tree service

### LINKED ISSUE #420: Use tree service for ListObjectVersions
We can use [tree service](https://github.com/nspcc-dev/neofs-node/pull/1324) to find object versions after #407 will be merged.

This task should be done after #416 

1. Split prefix into tokens and remove latest token
2. `GetNodeByPath` with tokens from 1.
3. `GetSubTree` with depth 1 and then filter nodes by `FileName` that starts with latest token
4. `GetSubTree` with math depth
5. Find nodes with both `OID` and `FileName` attributes and construct their path using specified separator.
6. HEAD required object from NeoFS (to avoid interaction with NeoFS consider storing version info (`ETag`, `Owner`, etc) into node attributes

KirillovDenis:
depends on #443 
close #420

### DIFF
--- a/api/cache/objectslist.go
+++ b/api/cache/objectslist.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/bluele/gcache"
+	"github.com/nspcc-dev/neofs-s3-gw/api/data"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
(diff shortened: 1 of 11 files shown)


KirillovDenis:
This should fully work after https://github.com/nspcc-dev/neofs-node/issues/1419

masterSplinter01 (approved):

Issue #446, opened 2022-05-23
Use `EncodeToString` instead of `String` in NeoFS SDK structures


alexvanin:
See https://github.com/nspcc-dev/neofs-node/issues/1408

Types:
- [ ] ~~Address~~ (not supported in SDK yet)
- [ ] Container ID
- [ ] Object ID
- [ ] User ID

PR #449, opened 2022-05-24
Tree tagging

### LINKED ISSUE #415: Use tree service for tags
We can use [tree service](https://github.com/nspcc-dev/neofs-node/pull/1324) for storing object/bucket tags after #407 will be merged.

### GetObjectTagging 
1. `GetNodeByPath` to find out node id (see [Get object](https://github.com/nspcc-dev/neofs-s3-gw/issues/413) for more info)
2. `GetSubTree` with depth 1
3. Find node that contains attribute `IsTag` and use its meta  

### PutObjectTagging 
1. Get object tag node as mentioned above
2. `Move` node to the same parent with attributes `[IsTag, ...]` and updated tags/attributes 

### DeleteObjectTagging
1. Get object tag node as mentioned above
2. `Remove` node from step 1

## Bucket tags
Use the same approach as for bucket settings (Add node to the root with attribute `SystemName: "bucket-tags"` and attributes-tags)

### GetBucketTagging
1. `GetNodeByPath` with `SystemName` (to build path) and with all attributes (maybe it is not supported yet) 
2. Get tags from attributes

### PutBucketTagging
1. `GetNodeByPath` using `SystemName` (to build path) and with all attributes (maybe it is not supported yet) 
2. If node doesn't exist then `Add` to root node with attributes `[SystemName, ...]` and tags/attributes
3. If node exists `Move` node to the same parent with attriubtes [SystemName, ...]` and updated tags/attributes

### DeleteBucketTagging
1. `GetNodeByPath` using `SystemName` (to build path)
2. `Remove` node from step 1

masterSplinter01:
Closes #415


internal/neofs/tree.go
```diff
@@ -274,6 +276,110 @@ func (c *TreeClient) DeleteBucketCORS(ctx context.Context, cnrID *cid.ID) (*oid.
...
+		if err != nil {
+			return nil, nil, err
+		}
+		for _, v := range versions {
+			if v.OID.String() == p.VersionID {
+				version = v
+				break
+			}
+		}
+	}
```
KirillovDenis, line 352: I think it's better to place getting required version in `layer` package. it will be consistent with getting objects and will simplify this function.

internal/neofs/tree.go
```diff
@@ -274,6 +276,110 @@ func (c *TreeClient) DeleteBucketCORS(ctx context.Context, cnrID *cid.ID) (*oid.
...
+
+	subtree, err := c.getSubTree(ctx, p.CnrID, versionTree, version.ID, 1)
+	if err != nil {
+		return version, nil, err
+	}
+
+	if subtree == nil {
+		return version, nil, nil
+	}
```
KirillovDenis, line 365: We can safely drop this condition

api/layer/tagging.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (n *layer) DeleteBucketTagging(ctx context.Context, cnrID *cid.ID) error {
+	n.systemCache.Delete(bucketTaggingCacheKey(cnrID))
+
+	return n.treeService.DeleteBucketTagging(ctx, cnrID)
+}
+
+func objectTaggingCacheKey(p *data.ObjectTaggingInfo) string {
+	return ".tagset." + p.CnrID.String() + "." + p.ObjName + "." + p.VersionID
```
alexvanin, line 124: Please use `CnrID.EncodeToString()` instead of `cnrID.String()` in this file, see https://github.com/nspcc-dev/neofs-s3-gw/issues/446

alexvanin (approved):

api/layer/layer.go
```diff
@@ -612,7 +521,12 @@ func (n *layer) removeVersionIfFound(ctx context.Context, bkt *data.BucketInfo,
 			return deleteMarkVersion, err
 		}
 
-		return deleteMarkVersion, n.DeleteObjectTagging(ctx, bkt, &data.ObjectInfo{ID: &version.OID, Bucket: bkt.Name, Name: obj.Name})
+		p := &data.ObjectTaggingInfo{
+			CnrID:     bkt.CID,
+			ObjName:   obj.Name,
+			VersionID: version.OID.String(),
```
KirillovDenis, line 527:
```suggestion
			VersionID: version.OID.EncodeToString(),
```

api/layer/tagging.go
```diff
@@ -0,0 +1,169 @@
...
+	return ".tagset." + cnrID.EncodeToString()
+}
+
+func (n *layer) getTaggedObjectVersion(ctx context.Context, p *data.ObjectTaggingInfo) (*data.NodeVersion, error) {
+	var (
+		err     error
+		version *data.NodeVersion
+	)
+
+	if p.VersionID == "null" {
```
KirillovDenis, line 137: Use `unversionedObjectVersionID` constant

internal/neofs/tree.go
```diff
@@ -274,6 +286,122 @@ func (c *TreeClient) DeleteBucketCORS(ctx context.Context, cnrID *cid.ID) (*oid.
...
+	return tagNode.Meta, nil
+}
+
+func (c *TreeClient) PutObjectTagging(ctx context.Context, cnrID *cid.ID, objVersion *data.NodeVersion, tagSet map[string]string) error {
+	tagNode, err := c.getObjectTaggingNode(ctx, cnrID, objVersion)
+	if err != nil {
+		return err
+	}
+
+	tagSet[isTagKV] = "true"
```
KirillovDenis, line 310: What if user provide tag with name the same as `isTagKV` constant value? Consider using `aws:` prefix for `isTagKV` (https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/allocation-tag-restrictions.html)

masterSplinter01: You're right, fixed for object tagging. For bucket tagging we don't use `isTagKV`, we keep bucket tags in a system node.

KirillovDenis: > You're right, fixed for object tagging. For bucket tagging we don't use `isTagKV`, we keep bucket tags in a system node.

Yes, you are right. But  user-defined tags can still conflict with `SystemName: bucket-tagging` attribute. So we have to handle this. We can use some prefix for user-defined tags

KirillovDenis (approved):

PR #454, opened 2022-05-25
Update/compatibility

alexvanin:


alexvanin:
`s3tests_boto3.functional.test_s3.test_atomic_dual_write_*`

Dual tests may pass or fail. It depends on the version ID when we upload asynchronously.

alexvanin:
`s3tests_boto3.functional.test_s3.test_bucket_create_naming_good_contains_period` related to https://github.com/nspcc-dev/neofs-contract/issues/245

alexvanin:
Atomic tests are not stable, they sometime pass, sometime fail.
Lock object improvement may be related with environment -- I run tests on single storage node environment.
The same is applied for encryption. I can drop encryption table update, because we didn't anything specific for this, as far as I can say.

KirillovDenis:
Let's add comments to sections which isn't supported yet and will not run those tests:
* Policy and replication
* Lifecycle
* Encryption
* Bucket ACL
* PostObject

Issue #452, opened 2022-05-25
Fix tests


KirillovDenis:
After #451 we should fix tests and drop all unused code (e.g. old versioning approach)

PR #455, opened 2022-05-25
Run GitHub Actions in tree-service branch

alexvanin:


alexvanin:
Tests and linter going to be fixed in https://github.com/nspcc-dev/neofs-s3-gw/issues/452
Coverage is failed due to failed test.

alexvanin:
See an issue with go1.17 test. Let me try different approach.

alexvanin:
I kept `dep` target the same. Updated workflows with extra step of calling `make sync-tree`.

KirillovDenis (approved):

PR #458, opened 2022-05-25
*: Refactor working with NeoFS identities

### LINKED ISSUE #446: Use `EncodeToString` instead of `String` in NeoFS SDK structures

cthulhu-rider:
Pull latest changes from NeoFS SDK Go library. Decrease redundant and
unsafe usage of ID pointers. Use `EncodeToString` method in order to
calculate protocol strings.

Signed-off-by: Leonard Lyubich <leonard@nspcc.ru>

* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/251
* closes #446

### DIFF
--- a/api/cache/names.go
+++ b/api/cache/names.go
@@ -4,7 +4,7 @@ import (
 	"time"
 
 	"github.com/bluele/gcache"
-	"github.com/nspcc-dev/neofs-sdk-go/object/address"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 )
 
 // ObjectsNameCache provides lru cache for objects.
@@ -32,23 +32,23 @@ func NewObjectsNameCache(config *Config) *ObjectsNameCache {
 	return &ObjectsNameCache{cache: gc}
 }
 
-// Get returns a cached object.
-func (o *ObjectsNameCache) Get(key string) *address.Address {
+// Get returns a cached object. Returns nil if value is missing.
+func (o *ObjectsNameCache) Get(key string) *oid.Address {
 	entry, err := o.cache.Get(key)
 	if err != nil {
 		return nil
 	}
 
-	result, ok := entry.(*address.Address)
+	result, ok := entry.(oid.Address)
 	if !ok {
 		return nil
 	}
 
-	return result
+	return &result
 }
 
 // Put puts an object to cache.
-func (o *ObjectsNameCache) Put(key string, address *address.Address) error {
+func (o *ObjectsNameCache) Put(key string, address oid.Address) error {
 	return o.cache.Set(key, address)
 }
 
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -217,8 +217,10 @@ func (h *handler) updateBucketACL(r *http.Request, astChild *ast, bktInfo *data.
 	}
 
 	parentAst := tableToAst(bucketACL.EACL, bktInfo.Name)
+	strCID := bucketACL.Info.CID.EncodeToString()
+
 	for _, resource := range parentAst.Resources {
-		if resource.Bucket == bucketACL.Info.CID.String() {
+		if resource.Bucket == strCID {
 			resource.Bucket = bktInfo.Name
 		}
 	}
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -519,8 +519,10 @@ func (h *handler) getNewEAclTable(r *http.Request, bktInfo *data.BucketInfo, obj
 	}
 
 	parentAst := tableToAst(bacl.EACL, objInfo.Bucket)
+	strCID := bacl.Info.CID.EncodeToString()
+
 	for _, resource := range parentAst.Resources {
-		if resource.Bucket == bacl.Info.CID.String() {
+		if resource.Bucket == strCID {
 			resource.Bucket = objInfo.Bucket
 		}
 	}
--- a/api/layer/versioning.go
+++ b/api/layer/versioning.go
@@ -271,10 +271,11 @@ func (v *objectVersions) getDelHeader() string {
 	return strings.Join(v.delList, ",")
 }
 
-func (v *objectVersions) getVersion(oid *oid.ID) *data.ObjectInfo {
+func (v *objectVersions) getVersion(obj oid.ID) *data.ObjectInfo {
+	strObj := obj.EncodeToString()
 	for _, version := range v.objects {
-		if version.Version() == oid.String() {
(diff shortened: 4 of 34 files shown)


creds/tokens/credentials.go
```diff
@@ -11,16 +11,15 @@ import (
...
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 type (
 	// Credentials is a bearer token get/put interface.
 	Credentials interface {
-		GetBox(context.Context, *address.Address) (*accessbox.Box, error)
-		Put(context.Context, *cid.ID, user.ID, *accessbox.AccessBox, uint64, ...*keys.PublicKey) (*address.Address, error)
+		GetBox(context.Context, oid.Address) (*accessbox.Box, error)
+		Put(context.Context, cid.ID, user.ID, *accessbox.AccessBox, uint64, ...*keys.PublicKey) (*oid.Address, error)
```
KirillovDenis, line 22: Shouldn't we also use `oid.Address` as returned value instead of pointer?

cthulhu-rider: It depends on the approach taken: return `(T, error)` or `(*T, error)`. IMO first is expected to improve performance (but it's compilator-dependent) while latter is easier to read. We have same discussion in neofs-node https://github.com/nspcc-dev/neofs-node/pull/1418#discussion_r880087590.

I suggest you also discuss, if you decide to return by value - let me know and I'll change.

api/cache/names.go
```diff
@@ -32,23 +32,18 @@ func NewObjectsNameCache(config *Config) *ObjectsNameCache {
 	return &ObjectsNameCache{cache: gc}
 }
 
-// Get returns a cached object.
-func (o *ObjectsNameCache) Get(key string) *address.Address {
+// Get returns a cached object. Indicates value presence by second return value.
+func (o *ObjectsNameCache) Get(key string) (v oid.Address, ok bool) {
```
alexvanin, line 36: I guess it is possible to keep signature the same and return `&result` value. All  `Get*` functions in caches return `nil` in case of cache miss, so maybe it's worth to keep it.

creds/tokens/credentials.go
alexvanin: I suggest to postpone decision about this and sync signatures in all places at once later.

internal/neofs/neofs.go
```diff
@@ -606,3 +602,8 @@ func isHomomorphicHashDisabled(ctx context.Context, p *pool.Pool) (res bool, err
 
 	return res, nil
 }
+
+// ObjectAddress represents globally unique object identifier of the object in the NeoFS network.
+//
+// Type is a mediator between oid.Address and internal needs.
+type ObjectAddress oid.Address
```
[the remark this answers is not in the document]
cthulhu-rider: Temp artifact, removed.

api/layer/versioning.go
```diff
@@ -271,10 +271,10 @@ func (v *objectVersions) getDelHeader() string {
 	return strings.Join(v.delList, ",")
 }
 
-func (v *objectVersions) getVersion(oid *oid.ID) *data.ObjectInfo {
+func (v *objectVersions) getVersion(obj oid.ID) *data.ObjectInfo {
 	for _, version := range v.objects {
-		if version.Version() == oid.String() {
-			if contains(v.delList, oid.String()) {
+		if version.Version() == obj.String() {
```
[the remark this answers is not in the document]
cthulhu-rider: Missed some places, I've now taken a closer look.

api/cache/names.go
cthulhu-rider: Tried to get rid of using pointers where it's not mutation-needed. Reverted.

alexvanin (approved):
LGTM

api/cache/names.go
```diff
@@ -32,23 +32,23 @@ func NewObjectsNameCache(config *Config) *ObjectsNameCache {
...
-func (o *ObjectsNameCache) Get(key string) *address.Address {
+// Get returns a cached object. Returns nil if value is missing.
+func (o *ObjectsNameCache) Get(key string) *oid.Address {
 	entry, err := o.cache.Get(key)
 	if err != nil {
 		return nil
 	}
 
-	result, ok := entry.(*address.Address)
+	result, ok := entry.(oid.Address)
```
cthulhu-rider, line 42: I suggest handling the conversion failure, otherwise the cache functionality will be completely broken and the program won't notice it. I like panic, but you can also log. Thoughts? @alexvanin @KirillovDenis

api/handler/acl.go
```diff
@@ -217,8 +217,10 @@ func (h *handler) updateBucketACL(r *http.Request, astChild *ast, bktInfo *data.
 	}
 
 	parentAst := tableToAst(bucketACL.EACL, bktInfo.Name)
+	strCID := bucketACL.Info.CID.EncodeToString()
+
 	for _, resource := range parentAst.Resources {
-		if resource.Bucket == bucketACL.Info.CID.String() {
+		if resource.Bucket == strCID {
```
cthulhu-rider, line 223: Can we break on true? @alexvanin @KirillovDenis

api/handler/put.go
```diff
@@ -519,8 +519,10 @@ func (h *handler) getNewEAclTable(r *http.Request, bktInfo *data.BucketInfo, obj
 	}
 
 	parentAst := tableToAst(bacl.EACL, objInfo.Bucket)
+	strCID := bacl.Info.CID.EncodeToString()
+
 	for _, resource := range parentAst.Resources {
-		if resource.Bucket == bacl.Info.CID.String() {
+		if resource.Bucket == strCID {
```
cthulhu-rider, line 525: Can we break on true? @alexvanin @KirillovDenis

api/cache/names.go
KirillovDenis: I would prefer to log it. But we have to provide a logger in the caches. So I think we can do it in a separate issue.

api/handler/put.go
KirillovDenis: No. The same as previous

api/cache/names.go
cthulhu-rider: #465

KirillovDenis (approved):

alexvanin (approved):

masterSplinter01 (approved):

PR #467, opened 2022-05-27
[#451] Handle lock objects using tree service

### LINKED ISSUE #451: Reconsider storing system objects
After #449  and #448  we should revise how we store the remain "system" objects. For example, cors object is saved using `putSystemObjectIntoNeoFS` (https://github.com/nspcc-dev/neofs-s3-gw/blob/20970466dacfaf0b30a0c852be064ff7f96d474b/api/layer/cors.go#L47) witch add extra headers and add node in tree service. It's redundant because later we create special node with cors (https://github.com/nspcc-dev/neofs-s3-gw/blob/20970466dacfaf0b30a0c852be064ff7f96d474b/api/layer/cors.go#L52). The same for notifications. 

Ideally, we should drop the following methods in [tree interface](https://github.com/nspcc-dev/neofs-s3-gw/blob/20970466dacfaf0b30a0c852be064ff7f96d474b/api/layer/tree_service.go#L41-L43) at all:
* AddSystemVersion
* GetSystemVersion
* RemoveSystemVersion

KirillovDenis:
close #451 

### DIFF
--- a/api/cache/system.go
+++ b/api/cache/system.go
@@ -47,6 +47,21 @@ func (o *SystemCache) GetObject(key string) *data.ObjectInfo {
 	return result
 }
 
+// GetLockInfo returns a cached object.
+func (o *SystemCache) GetLockInfo(key string) *data.LockInfo {
+	entry, err := o.cache.Get(key)
+	if err != nil {
+		return nil
+	}
+
+	result, ok := entry.(*data.LockInfo)
+	if !ok {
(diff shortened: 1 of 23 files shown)


KirillovDenis:
Tests and linter going to be fixed in https://github.com/nspcc-dev/neofs-s3-gw/issues/452

alexvanin (approved):

Issue #438, opened 2022-05-17
Remove `layer/neofs` package


alexvanin:
https://github.com/nspcc-dev/neofs-s3-gw/commit/13080d6d969a8e858a73f7bfac53327bf3b08b4f introduced `layer/neofs` package to avoid cycle dependencies. They all gone now, so we can drop this extra package to keep only one `neofs` pkg in `internal/neofs`.

KirillovDenis:
It seems the cycle dependency in tests.

KirillovDenis:
Consider approach like https://github.com/nspcc-dev/neofs-s3-gw/pull/473#discussion_r885531054

PR #473, opened 2022-05-31
[#452] Fix tests

### LINKED ISSUE #452: Fix tests

KirillovDenis:
There is not any `search` anymore

close #452


api/layer/tree_mock.go
```diff
@@ -1,8 +1,7 @@
-package tree
+package layer
```
alexvanin, line 1: What the idea behind moving it into `layer` package? `neofs_mock.go` is still in `neofstest` package meanwhile.

api/handler/object_list.go
```diff
@@ -281,7 +281,7 @@ func encodeListObjectVersionsToResponse(info *layer.ListObjectVersionsInfo, buck
 				DisplayName: ver.Object.Owner.String(),
 			},
 			Size:      ver.Object.Size,
-			VersionID: ver.Object.NullableVersion(),
+			VersionID: ver.Object.Version(), // todo return "null" version for unversioned
```
alexvanin, line 284: Mention #474

api/layer/tree_mock.go
KirillovDenis: Cyclic dependency. Like https://github.com/nspcc-dev/neofs-s3-gw/issues/438

alexvanin (approved):

PR #485, opened 2022-06-01
Upgrade SDK with latest `bearer` package API

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/259

I don't insist on merging this PR, just opened it for information and advice.

PR #487, opened 2022-06-01
 Remove attach of bearer token when bucket owner is not an issuer of the bearer token  

### LINKED ISSUE #459: Fix `test_object_copy_canned_acl`
This test sets canned ACL to the object. 

```py
def test_object_copy_canned_acl():
    bucket_name = get_new_bucket()
    client = get_client()
    alt_client = get_alt_client()
    client.put_object(Bucket=bucket_name, Key='foo123bar', Body='foo')

    copy_source = {'Bucket': bucket_name, 'Key': 'foo123bar'}
    client.copy_object(Bucket=bucket_name, CopySource=copy_source, Key='bar321foo', ACL='public-read')
    ## check ACL is applied by doing GET from another user
    alt_client.get_object(Bucket=bucket_name, Key='bar321foo')  # <- FAILED HERE

    metadata={'abc': 'def'}
    copy_source = {'Bucket': bucket_name, 'Key': 'bar321foo'}
    client.copy_object(ACL='public-read', Bucket=bucket_name, CopySource=copy_source, Key='foo123bar', Metadata=metadata, MetadataDirective='REPLACE')

    # check ACL is applied by doing GET from another user
    alt_client.get_object(Bucket=bucket_name, Key='foo123bar')
```

ACL is not processed during object copying, therefore alt client fails with 
```
layer/object.go:710     error was transformed   {"request_id": "46f7588e-02a7-47ff-94ff-6d26354983bd", "error": "access denied: access to operation SEARCH is denied by extended ACL check: bearer token is not signed by the container owner"}
handler/util.go:25      could not find object   {"request_id": "46f7588e-02a7-47ff-94ff-6d26354983bd", "method": "GetObject", "bucket_name": "yournamehere-sdcore7ma75qo04d-1", "object_name": "bar321foo", "error": "AccessDenied: 403 => Access Denied."}
```

masterSplinter01:
The `test_object_copy_canned_acl` test still not passing because `main_wallet` creates a private container in tests and `alt_wallet` can't `get-object` from private container, even the object was put with `public-read` acl. 

I tried to make `main-wallet` create `public-read` container and the test began to pass.   

I suggest to fix the test and place it in `test_s3_neofs.py`. 

Closes #459

### DIFF
--- a/api/handler/get.go
+++ b/api/handler/get.go
@@ -168,6 +168,7 @@ func (h *handler) GetObjectHandler(w http.ResponseWriter, r *http.Request) {
 		ObjectInfo: info,
 		Writer:     w,
 		Range:      params,
+		BucketInfo: bktInfo,
 	}
 	if err = h.obj.GetObject(r.Context(), getParams); err != nil {
 		h.logAndSendError(w, "could not get object", reqInfo, err)
--- a/api/handler/get.go
+++ b/api/handler/get.go
@@ -168,7 +168,6 @@ func (h *handler) GetObjectHandler(w http.ResponseWriter, r *http.Request) {
 		ObjectInfo: info,
 		Writer:     w,
 		Range:      params,
-		VersionID:  p.VersionID,
 	}
 	if err = h.obj.GetObject(r.Context(), getParams); err != nil {
 		h.logAndSendError(w, "could not get object", reqInfo, err)
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -77,6 +77,7 @@ type (
 	GetObjectParams struct {
 		Range      *RangeParams
 		ObjectInfo *data.ObjectInfo
+		BucketInfo *data.BucketInfo
 		Writer     io.Writer
 	}
 
@@ -124,6 +125,7 @@ type (
 	// CopyObjectParams stores object copy request parameters.
 	CopyObjectParams struct {
 		SrcObject  *data.ObjectInfo
+		ScrBktInfo *data.BucketInfo
 		DstBktInfo *data.BucketInfo
 		DstObject  string
 		SrcSize    int64
@@ -382,6 +384,7 @@ func (n *layer) GetObject(ctx context.Context, p *GetObjectParams) error {
 	var params getParams
 
 	params.objInfo = p.ObjectInfo
+	params.bktInfo = p.BucketInfo
 
 	if p.Range != nil {
 		if p.Range.Start > p.Range.End {
@@ -517,6 +520,7 @@ func (n *layer) CopyObject(ctx context.Context, p *CopyObjectParams) (*data.Obje
 			ObjectInfo: p.SrcObject,
 			Writer:     pw,
 			Range:      p.Range,
+			BucketInfo: p.ScrBktInfo,
 		})
 
 		if err = pw.CloseWithError(err); err != nil {
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -78,7 +78,6 @@ type (
 		Range      *RangeParams
 		ObjectInfo *data.ObjectInfo
 		Writer     io.Writer
-		VersionID  string
 	}
 
 	// HeadObjectParams stores object head request parameters.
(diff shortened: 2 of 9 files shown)


api/handler/get.go
```diff
@@ -168,7 +168,6 @@ func (h *handler) GetObjectHandler(w http.ResponseWriter, r *http.Request) {
 		ObjectInfo: info,
 		Writer:     w,
 		Range:      params,
-		VersionID:  p.VersionID,
```
masterSplinter01, line 171: It seems like the field is not used, also we have `versionID` in `ObjectInfo` field

api/layer/layer.go
```diff
@@ -323,10 +322,12 @@ func (n *layer) Owner(ctx context.Context) user.ID {
 	return ownerID
 }
 
-func (n *layer) prepareAuthParameters(ctx context.Context, prm *neofs.PrmAuth) {
+func (n *layer) prepareAuthParameters(ctx context.Context, prm *neofs.PrmAuth, bktOwner user.ID) {
 	if bd, ok := ctx.Value(api.BoxData).(*accessbox.Box); ok && bd != nil && bd.Gate != nil {
-		prm.BearerToken = bd.Gate.BearerToken
-		return
+		if issuer, ok := bd.Gate.BearerToken.Issuer(); ok && bktOwner.Equals(issuer) {
```
KirillovDenis, line 327: It seems we can adopt https://github.com/nspcc-dev/neofs-s3-gw/pull/485

api/layer/object.go
```diff
@@ -437,23 +444,22 @@ func (n *layer) headVersion(ctx context.Context, bkt *data.BucketInfo, p *HeadOb
...
-	n.objCache.Delete(newAddress(idCnr, idObj))
+	n.objCache.Delete(newAddress(bktInfo.CID, idObj))
 
 	return n.transformNeofsError(ctx, n.neoFS.DeleteObject(ctx, prm))
 }
 
 // objectPutAndHash prepare auth parameters and invoke neofs.CreateObject.
 // Returns object ID and payload sha256 hash.
 func (n *layer) objectPutAndHash(ctx context.Context, prm neofs.PrmObjectCreate) (*oid.ID, []byte, error) {
-	n.prepareAuthParameters(ctx, &prm.PrmAuth)
```
KirillovDenis, line 456: Maybe we can provide `bktInfo` to this function to preserve invocation of the `n.prepareAuthParameters` here?

api/layer/layer.go
```diff
@@ -380,8 +381,7 @@ func (n *layer) ListBuckets(ctx context.Context) ([]*data.BucketInfo, error) {
 func (n *layer) GetObject(ctx context.Context, p *GetObjectParams) error {
 	var params getParams
 
-	params.oid = p.ObjectInfo.ID
-	params.cid = p.ObjectInfo.CID
+	params.objInfo = p.ObjectInfo
```
KirillovDenis, line 384: I suppose we can keep the old `getParams` fields. And extend `GetObjectParams` by `bktInfo` and pass it (`bktInfo`) as the second parameter to `n.initObjectPayloadReader` function

KirillovDenis:
Please fix tests

api/layer/layer.go
alexvanin: I would like to rebase `tree-service` branch after this PR. Adopt it in separate PR.

alexvanin:
> can't get-object from private container, even the object was put with public-read acl.
>
> I suggest to fix the test and place it in test_s3_neofs.py.

It's more like new S3 issue to fix object ACLs, isn't it?

1. `main_wallet` creates private container -- extended ACL denied for everyone
2. `main_wallet` sets public read acl for the object -- extended ACL has one more additional `allow` rule to this object for target `OTHER`, it is a first rule in the table
3. `alt_wallet` tries to get object -- without bearer token NeoFS node applies container EACL to request => first rule allows to get object.

api/layer/layer.go
masterSplinter01: Removed this with dragging of `bktInfo` through params. Added a separated commit to make it easier to view changes

KirillovDenis (approved):

alexvanin:
> > can't get-object from private container, even the object was put with public-read acl.
> > I suggest to fix the test and place it in test_s3_neofs.py.
> 
> It's more like new S3 issue to fix object ACLs, isn't it?
> 
>     1. `main_wallet` creates private container -- extended ACL with 7 `deny` rules for target `OTHER`
> 
>     2. `main_wallet` sets public read acl for the object -- extended ACL has one more additional `allow` rule to this object for target `OTHER`, it is a first rule in the table
> 
>     3. `alt_wallet` tries to get object -- without bearer token NeoFS node applies container EACL to request => first rule allows to get object.

As @masterSplinter01 mentioned, we got access deny error because of `search` request. Which is fair enough, because we do search in the bucket. `tree-service` branch avoids extra search, so maybe the issue should be gone. Let's investigate that and fix per object public ACL in private containers.

api/layer/layer.go
alexvanin: I messed up with the branches. Both SDK update and this PR were into `master` branch, so we could adopt it. For some reason I thought this PR has `tree-service` target.

I pushed https://github.com/nspcc-dev/neofs-s3-gw/commit/4ed939773b020ee6e305f9123b53e62cdbd6f859 into master branch.

PR #383, opened 2022-03-25
[#382] Refactor neofs related types

KirillovDenis:
closes #382 

Signed-off-by: Denis Kirillov <denis@nspcc.ru>

PR #492, opened 2022-06-03
layer: Don't parameterize basic ACL of created containers

cthulhu-rider:


alexvanin (changes requested):
This PR partially reverts https://github.com/nspcc-dev/neofs-s3-gw/pull/383, returns mediator into application package. Let's not do that until internal package will be revised in https://github.com/nspcc-dev/neofs-s3-gw/issues/494. To avoid parametrization, `NeoFS` and `AuthmateNeoFS` types from `internal/neofs` package can be changed. Both have `CreateContainer` method, and code can be shared inside the package.

Also, looking at the required changes, I am not sure removed parametrization is so worth of having all these changes in master branch.

KirillovDenis (approved):

alexvanin (approved):

PR #495, opened 2022-06-03
Add acl to copy and fix if filters in acl

masterSplinter01:
To get `versionID` in getObjectACL` we call `HeadObject` which should be removed when `tree-service` will be merged into `master`.


KirillovDenis:
I suppose we have to add extra `version-id` handling (in acl filters) rather than replace existing logic. [PutBucketPolicy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-bucket-policies.html) allows to restrict access to object by name (without `version-id`). I think we shouldn't drop this functionality.

alexvanin (approved):

KirillovDenis (approved):

Issue #478, opened 2022-05-31
Do not fetch current epoch at object upload


alexvanin:
#457 uses `neofs.TimeToEpoch` invocation to fetch current epoch during object upload. It is required to compare versions of the same file. With `tree-service` we don't have to do that at all.

KirillovDenis:
Wait for `tree-service` branch includes #457

KirillovDenis:
Done in #497

PR #497, opened 2022-06-03
Remove current epoch call

alexvanin:
Creation epoch was used for versioning. With
tree service, versioning is done on the NeoFS
side in the tree.

Related to #478


KirillovDenis (approved):

alexvanin:
> I thought the CodeQL issues are gone now

Those are new alerts. I will do something with it in this PR.

alexvanin:
I tried running CodeQL queries locally and they do not produce stable output when I change a code. So we are going to sanitise logs again and again.

I suggest to suppress CWE-117 warnings, because they are not relevant for `zap` logger. It uses JSON formatted output, which replaces all `\r` and `\n` symbols. Similar issue was mentioned in [etcd](https://github.com/etcd-io/etcd/issues/13588#issuecomment-1011021435) repository, but there is no progress on this specific issue. 

It is possible to use custom [CodeQL query package](https://github.com/github/codeql-go/tree/main/ql/src) without CWE-117 check, but I don't think it worth time to do and maintain this.

So my proposal is:
1. Disable CodeQL checks in `tree-service` branch. Unit tests and linter should be more than enough for developing.
2. Suppress all CWE-117 warnings in master branch.
3. https://github.com/nspcc-dev/neofs-s3-gw/issues/498 after `tree-service` branch merge.

KirillovDenis (approved):

PR #514, opened 2022-06-07
[#514] Missed version id in tags

KirillovDenis:

### DIFF
--- a/api/layer/tagging.go
+++ b/api/layer/tagging.go
@@ -10,34 +10,36 @@ import (
 	"go.uber.org/zap"
 )
 
-func (n *layer) GetObjectTagging(ctx context.Context, p *ObjectVersion) (map[string]string, error) {
+func (n *layer) GetObjectTagging(ctx context.Context, p *ObjectVersion) (string, map[string]string, error) {
 	var (
 		err  error
 		tags map[string]string
 	)
+
 	tags = n.systemCache.GetTagging(objectTaggingCacheKey(p))
 	if tags != nil {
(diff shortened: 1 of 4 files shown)


api/layer/tagging.go
```diff
@@ -3,48 +3,54 @@ package layer
...
 		tags map[string]string
 	)
-	tags = n.systemCache.GetTagging(objectTaggingCacheKey(p))
+
+	key := objectTaggingCacheKey(p)
+	tags = n.systemCache.GetTagging(key)
 	if tags != nil {
-		return tags, nil
+		split := strings.Split(key, ".")
+		return split[len(split)-1], tags, nil
```
alexvanin, line 24: If `tags != nil` then version is stored in `p.VersionID`, isn't it, thus string splitting is not required, isn't it?

api/layer/tagging.go
```diff
@@ -10,41 +10,44 @@ import (
...
-		return tags, nil
+		return p.VersionID, tags, nil
 	}
 
 	version, err := n.getNodeVersion(ctx, p)
 	if err != nil {
-		return nil, err
+		return "", nil, err
 	}
+	p.VersionID = version.OID.EncodeToString()
```
masterSplinter01, line 28: We don't use `p.Version` after the assignment

KirillovDenis: In line 38, we form key to cache tags

alexvanin (approved):

masterSplinter01 (approved):

PR #526, opened 2022-06-10
[#524] Optimize listing

### LINKED ISSUE #524: Reduce HEAD amount in listing
Currently we make head on every object even the final result doesn't contain it. We should optimize this.
image

KirillovDenis:
close #524 


api/layer/object.go
```diff
@@ -331,124 +334,127 @@ func (n *layer) objectPutAndHash(ctx context.Context, prm neofs.PrmObjectCreate,
...
 		}
-	}
 
-	objects := make([]*data.ObjectInfo, 0, len(objectsMap))
-	for _, obj := range objectsMap {
-		objects = append(objects, obj)
+		if oi := objectInfoFromMeta(bkt, obj, p.Prefix, p.Delimiter); oi != nil {
+			objects = append(objects, oi)
+		}
+		if len(objects) == p.MaxKeys+1 {
```
alexvanin, line 452: Check length before `object.Head` request to save one more RPC.

api/layer/object.go
```diff
@@ -331,124 +334,127 @@ func (n *layer) objectPutAndHash(ctx context.Context, prm neofs.PrmObjectCreate,
...
-	if len(allObjects) == 0 {
-		return &result, nil
-	}
-
-	if p.ContinuationToken != "" {
-		allObjects = trimAfterObjectID(p.ContinuationToken, allObjects)
-	}
-
-	if p.StartAfter != "" {
-		allObjects = trimAfterObjectName(p.StartAfter, allObjects)
```
alexvanin, line 392: `trimAfterObjectID` and `trimAfterObjectName` can be removed, I think. 

Or you can hide some comparasion details in those functions and use them in `getLatestObjectsVersions` to improve readability.

alexvanin (commented):
Overall LGTM, will go more deeply into `internal/neofs` changes when PR will be open.

Issue #529, opened 2022-06-14
Support presigned URLs


alexvanin:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html

URL Limiting is tied with IAM. I guess we can allow presigned URLs for container owners?

PR #530, opened 2022-06-14
[#528] Add java sdk client configuration

### LINKED ISSUE #528: Investigate V4 signatures from S3 Java SDK
It seems that S3 Java SDK does not calculate signatures the way neofs-s3-gw expects. We had similar signature issues with other software (maybe it used Java SDK?)

If there is an option to configure Java client to be compatible with neofs-s3-gw, then describe it in README or docs.

If there is no way to configure Java client, then consider changes in neofs-s3-gw code to be compatible with both awscli and Java client.

KirillovDenis:
close #528 

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,36 @@ $ S3_GW_PEERS_0_ADDRESS=grpcs://192.168.130.72:8080 \
   neofs-s3-gw
 ```
 
+## AWS SDK Compatibility
+
+To match signature of the request you must not include the following headers to `SignedHeaders`:
+* `User-Agent`
+* `X-Amzn-Trace-Id`
+
+### AWS SDK JAVA v1
+Using [aws-sdk-java](https://github.com/aws/aws-sdk-java) you can get the following error:
+```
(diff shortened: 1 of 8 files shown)


alexvanin:
Should we mention GitHub discussion why some fields should be omitted? Are there any related discussion in S3 SDK Go repository?

alexvanin:
> The problem is using different agents with pre-signed URLs

So it affects #529. In this PR I propose to check if it is possible to copy signature verification code in our repository and modify it so we will have functions to verify preisgned URLs and regular URLs. Check if `awscli` will be compatible. https://github.com/nspcc-dev/neofs-s3-gw/issues/528#issuecomment-1156037423

KirillovDenis:
I think it should be rebased on #531

alexvanin:
With `api/signer/v4` package is it possible to connect from default Java SDK Client to S3 Gateway?

KirillovDenis:
> With `api/signer/v4` package is it possible to connect from default Java SDK Client to S3 Gateway?

Yes. It's a main goal to adopt `signer/v4`

alexvanin (approved):

PR #531, opened 2022-06-14
[#529] Add presign URLs support

### LINKED ISSUE #529: Support presigned URLs

KirillovDenis:
close #529 


alexvanin (approved):
@realloc @anatoly-bogatyrev I think it is a good example of a feature for separate integration test. Should we create an issue in testcase repository?

PR #536, opened 2022-06-15
[#533] Fix typos

### LINKED ISSUE #533: Typo in KV
```
	// keys for delete marker nodes.
	isDeleteMarkerKV = "I_d_DeleteMarker"
```

KirillovDenis:
Note: merging this changed can lead to missing some data if s3-gw will be used with old trees (from tree service)

close #533 

### DIFF
--- a/internal/neofs/tree.go
+++ b/internal/neofs/tree.go
@@ -47,12 +47,12 @@ type (
 )
 
 const (
-	versioningEnabledKV = "versioning_enabled"
-	lockConfigurationKV = "lock_configuration"
+	versioningEnabledKV = "VersioningEnabled"
+	lockConfigurationKV = "LockConfiguration"
 	oidKV               = "OID"
 	fileNameKV          = "FileName"
 	isUnversionedKV     = "IsUnversioned"
-	isTagKV             = "isTag"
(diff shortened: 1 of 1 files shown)


masterSplinter01 (approved):

alexvanin:
I think @realloc wanted some transition routine in case of incompatible changes like this one. Let's discuss how we are going to do that now or in the future.

- Don't care and break compatibility, because it is _experimental_ branch
- Add transition tool that will iterate over the trees and change names
  - Tool has to find all containers in the network
  - Tool has to have access for container in order to update tree. 

I can imagine a tool that takes wallet that has been used in authmate, lists all user containers and updates tree structures. So we can update tree structure per user, but not per system. In order to update tree structure per system, tool should have access for all containers. Maybe one of alphabet wallets should be able to do that? Maybe tool has to operate with all alphabet wallet and use multisignature account?

alexvanin:
According to latest discussion with @realloc, we can allow having incompatible changes for next release. Next release is going to merge tree service into master branch (along with neofs-node) and it is going to be feature freeze. So it's better to have all renamings before that.

So please rebase this branch.

/cc @KirillovDenis @masterSplinter01

PR #537, opened 2022-06-15
Upgrade NeoFS SDK Go with changed `netmap` package

cthulhu-rider:
`PlacementPolicy` type now provides methods to work with QL-encoded
policies. System network parameters can be read using dedicated method
without iterating. Applications can work with `PlacementPolicy`
variables directly so there is no need to use pointers.

Signed-off-by: Leonard Lyubich <leonard@nspcc.ru>

* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/271

### DIFF
--- a/authmate/authmate.go
+++ b/authmate/authmate.go
@@ -159,7 +159,7 @@ func (a *Agent) checkContainer(ctx context.Context, opts ContainerOptions, idOwn
 
 	cnrID, err := a.neoFS.CreateContainer(ctx, prm)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("create container in NeoFS: %w", err)
 	}
 
 	return cnrID, nil
--- a/authmate/authmate.go
+++ b/authmate/authmate.go
@@ -5,6 +5,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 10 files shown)


authmate/authmate.go
```diff
@@ -147,16 +147,17 @@ func (a *Agent) checkContainer(ctx context.Context, opts ContainerOptions, idOwn
...
 
-	cnrID, err := a.neoFS.CreateContainer(ctx, PrmContainerCreate{
-		Owner:        idOwner,
-		Policy:       *pp,
-		FriendlyName: opts.FriendlyName,
-	})
+	prm.Owner = idOwner
+	prm.FriendlyName = opts.FriendlyName
+
+	cnrID, err := a.neoFS.CreateContainer(ctx, prm)
```
KirillovDenis, line 160: I think we can write just:
```suggestion
	return a.neoFS.CreateContainer(ctx, prm)
```

KirillovDenis (approved):

authmate/authmate.go
cthulhu-rider: Really, thanks. Don't we want to add context to the error using `fmt.Errorf`? If no, I'll `return` directly.

KirillovDenis: Actually it would be nice to add context to errors across the all project. It seems we need a new issue for that.

KirillovDenis: So you can add context to this case now. Or we'll do it in https://github.com/nspcc-dev/neofs-s3-gw/issues/539

masterSplinter01 (dismissed):

alexvanin (approved):

PR #550, opened 2022-06-23
[#539] Add context to errors

### LINKED ISSUE #539: Add context to errors across the all project
Actually it would be nice to add context to errors across the all project. It seems we need a new issue for that.

_Originally posted by @KirillovDenis in https://github.com/nspcc-dev/neofs-s3-gw/pull/537#discussion_r899018099_

KirillovDenis:
Didn't add context to some error because such errors can already be [appropriate error](https://github.com/nspcc-dev/neofs-s3-gw/blob/8a1fc8ae3f99cff5dc060d969a384e94e109ec1f/api/errors/errors.go#L15) and we have to preserve this.

I see two possible approach to manage this (add context to all errors): 

1. Add the following function and Invoke it [here](https://github.com/nspcc-dev/neofs-s3-gw/blob/8a1fc8ae3f99cff5dc060d969a384e94e109ec1f/api/response.go#L116) (to send appropriate error to user):
```
func fetchS3Error(err error) *errors.Error {
	for err != nil {
		if e, ok := err.(errors.Error); ok {
			return &e
		}
		err = stderrors.Unwrap(err)
	}
	return nil
}
```

2. Add the following function and invoke it when we want to add context:
```
func wrapError(err error, msg string) error {
	if _, ok := err.(errors.Error); ok {
		return err
	}

	return fmt.Errorf("%s: %w", msg, err)
}
```

The first approach is more CPU intensive. The second one can potentially lead to missing some important context.

close #539 

### DIFF
--- a/api/handler/cors.go
+++ b/api/handler/cors.go
@@ -5,6 +5,8 @@ import (
 	"strconv"
 	"strings"
 
+	"go.uber.org/zap"
+
 	"github.com/nspcc-dev/neofs-s3-gw/api"
 	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
 	"github.com/nspcc-dev/neofs-s3-gw/api/layer"
@@ -89,11 +91,13 @@ func (h *handler) AppendCORSHeaders(w http.ResponseWriter, r *http.Request) {
 	}
 	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
 	if err != nil {
+		h.log.Warn("get bucket info", zap.Error(err))
 		return
 	}
 
 	cors, err := h.obj.GetBucketCORS(r.Context(), bktInfo)
 	if err != nil {
+		h.log.Warn("get bucket cors", zap.Error(err))
 		return
 	}
 
--- a/api/layer/container.go
+++ b/api/layer/container.go
@@ -2,6 +2,7 @@ package layer
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"time"
 
@@ -48,7 +49,7 @@ func (n *layer) containerInfo(ctx context.Context, idCnr cid.ID) (*data.BucketIn
 		if client.IsErrContainerNotFound(err) {
 			return nil, errors.GetAPIError(errors.ErrNoSuchBucket)
 		}
-		return nil, err
+		return nil, fmt.Errorf("get neofs container: %w", err)
 	}
 
 	info.Owner = *res.OwnerID()
@@ -151,13 +152,13 @@ func (n *layer) createContainer(ctx context.Context, p *CreateBucketParams) (*da
 		AdditionalAttributes: attributes,
 	})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("create container: %w", err)
 	}
 
 	bktInfo.CID = *idCnr
 
 	if err = n.setContainerEACLTable(ctx, bktInfo.CID, p.EACL, p.SessionEACL); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("set container eacl: %w", err)
 	}
 
 	if err = n.bucketCache.Put(bktInfo); err != nil {
--- a/api/layer/cors.go
+++ b/api/layer/cors.go
@@ -24,7 +24,7 @@ func (n *layer) PutBucketCORS(ctx context.Context, p *PutCORSParams) error {
 	)
 
 	if err := xml.NewDecoder(tee).Decode(cors); err != nil {
-		return err
+		return fmt.Errorf("xml decode cors: %w", err)
 	}
 
 	if cors.CORSRules == nil {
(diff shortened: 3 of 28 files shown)


authmate/authmate.go
```diff
@@ -369,7 +369,7 @@ func buildBearerTokens(key *keys.PrivateKey, table *eacl.Table, lifetime lifetim
 	for _, gateKey := range gatesKeys {
 		tkn, err := buildBearerToken(key, table, lifetime, gateKey)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("builf bearer token: %w", err)
```
alexvanin, line 372:
```suggestion
			return nil, fmt.Errorf("build bearer token: %w", err)
```

authmate/authmate.go
```diff
@@ -211,22 +211,22 @@ func (a *Agent) IssueSecret(ctx context.Context, w io.Writer, options *IssueSecr
...
 	policies, err := preparePolicy(options.ContainerPolicies)
 	if err != nil {
-		return err
+		return fmt.Errorf("prepare policies: %w", err)
 	}
 
 	lifetime.Iat, lifetime.Exp, err = a.neoFS.TimeToEpoch(ctx, time.Now().Add(options.Lifetime))
 	if err != nil {
-		return err
+		return fmt.Errorf("time to epoch fail: %w", err)
```
alexvanin, line 219: Maybe `fetch time to epoch: %w`?

api/response.go
```diff
@@ -192,9 +192,9 @@ func EncodeToResponse(w http.ResponseWriter, response interface{}) error {
 	w.WriteHeader(http.StatusOK)
 
 	if _, err := w.Write(xmlHeader); err != nil {
-		return err
+		return fmt.Errorf("write headers: %w", err)
 	} else if err = xml.NewEncoder(w).Encode(response); err != nil {
-		return err
+		return fmt.Errorf("xml encode response: %w", err)
```
alexvanin, line 197: Verb in the first place looks better, IMO: `encode xml response: %w`.

api/layer/cors.go
```diff
@@ -44,12 +44,11 @@ func (n *layer) PutBucketCORS(ctx context.Context, p *PutCORSParams) error {
...
 
-	_, err := n.putSystemObjectIntoNeoFS(ctx, s)
-	if err != nil {
-		return err
+	if _, err := n.putSystemObjectIntoNeoFS(ctx, s); err != nil {
+		return fmt.Errorf("put system object: %w", err)
 	}
 
-	if err = n.systemCache.PutCORS(systemObjectKey(p.BktInfo, s.ObjName), cors); err != nil {
+	if err := n.systemCache.PutCORS(systemObjectKey(p.BktInfo, s.ObjName), cors); err != nil {
```
alexvanin, line 51: Why shadow `err` there? In `api/layer/object.go` and `api/handler/multipart_upload.go` you did the opposite.

api/notifications/controller.go
```diff
@@ -115,12 +115,12 @@ func NewController(p *Options, l *zap.Logger) (*Controller, error) {
 
 	nc, err := nats.Connect(p.URL, ncopts...)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("nats connect: %w", err)
```
alexvanin, line 118: Verb in the first place looks better.

api/layer/container.go
```diff
@@ -48,7 +49,7 @@ func (n *layer) containerInfo(ctx context.Context, idCnr cid.ID) (*data.BucketIn
 		if client.IsErrContainerNotFound(err) {
 			return nil, errors.GetAPIError(errors.ErrNoSuchBucket)
 		}
-		return nil, err
+		return nil, fmt.Errorf("neofs get container: %w", err)
```
alexvanin, line 52: Verb first.

api/handler/cors.go
```diff
@@ -89,11 +91,13 @@ func (h *handler) AppendCORSHeaders(w http.ResponseWriter, r *http.Request) {
 	}
 	bktInfo, err := h.obj.GetBucketInfo(r.Context(), reqInfo.BucketName)
 	if err != nil {
+		h.log.Warn("failed get bucket info", zap.Error(err))
```
alexvanin, line 94: failed _to_ get, or simple `get bucket info: %w`.

alexvanin (changes requested):
You've added a plenty of WARN logs after gcache get. Cache miss produces [KeyNotFound](https://github.com/bluele/gcache/blob/d8b7e051c564c174fea6ef60d180abf601099015/cache.go#L17). This error is valid and should not be logged.

api/layer/cors.go
KirillovDenis: Because the `err` is new variable now

alexvanin (approved):

masterSplinter01 (approved):

PR #561, opened 2022-06-28
[#490] Optimize GetObjectTaggingAndLock

### LINKED ISSUE #490: Provide NodeVersion to GetObjectTaggingAndLock method
Actually we can go further and save one more request to the tree service (getting `*data.NodeVersion`). Currently we do this twice: in `layer.GetObjectInfo` and in `layer.GetObjectTaggingAndLock`. But this requires significant refactoring.
We can consider this as a separate issue though.

_Originally posted by @KirillovDenis in https://github.com/nspcc-dev/neofs-s3-gw/issues/480#issuecomment-1144596703_

KirillovDenis:
close #490 

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -321,7 +321,7 @@ func (h *handler) PutObjectACLHandler(w http.ResponseWriter, r *http.Request) {
 		VersionID: versionID,
 	}
 
-	objInfo, err := h.obj.GetObjectInfo(r.Context(), p)
+	extendedInfo, err := h.obj.GetObjectInfo(r.Context(), p)
 	if err != nil {
 		h.logAndSendError(w, "could not get object info", reqInfo, err)
 		return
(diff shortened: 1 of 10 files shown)


api/handler/acl.go
```diff
@@ -321,7 +321,7 @@ func (h *handler) PutObjectACLHandler(w http.ResponseWriter, r *http.Request) {
 		VersionID: versionID,
 	}
 
-	objInfo, err := h.obj.GetObjectInfo(r.Context(), p)
+	extInfo, err := h.obj.GetObjectInfo(r.Context(), p)
```
masterSplinter01, line 324: I think it is not obvious enough that `extInfo` means extended info. Maybe it'd be better to name it as `extendedInfo` or `extendedObjectInfo`?

alexvanin (approved):
Fix conflict and merge.

PR #581, opened 2022-07-06
Do not use user role with public keys in eacl target

### LINKED ISSUE #553: Working with the eACL rule target apart from the protocol
I've noticed next statement in ACL code:
https://github.com/nspcc-dev/neofs-s3-gw/blob/1012a3c6a418183bdf75d50b30f0bdc6260977ec/api/handler/acl.go#L820-L824

According to this, handler expects that both fields can be set in the [message](https://github.com/nspcc-dev/neofs-api/blob/10f18e4b69ec79991b1a2b83e212d99203f3ee1a/acl/types.proto#L150-L158).

However, protocol requires (maybe not detailed enough) exactly one field to be set. Am I missing something? What can we do to follow the protocol?

alexvanin:
Blocked by #580 
Closes #553

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -808,6 +808,7 @@ func formRecords(operations []*astOperation, resource *astResource) ([]*eacl.Rec
 				}
 				targetKeys = append(targetKeys, (ecdsa.PublicKey)(*pk))
 			}
+			// Unknown role is used, because it is ignored when keys are set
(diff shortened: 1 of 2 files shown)


cthulhu-rider:
@alexvanin developing interface of target setting is going to force mutually exclusive setting of a role or keys.

KirillovDenis (dismissed):

masterSplinter01 (approved):

KirillovDenis (approved):

PR #583, opened 2022-07-06
 Fix error when file with policy cannot be read

### LINKED ISSUE #558: Print error "File not found" if a .json file doesn't exist
If not existing filepath is set, authmate tries to parse it anyway and prints the following error:
```
$ authmate issue-secret --container-policy police.json
couldn't parse container policy: invalid character 's' looking for beginning of value
```

It looks confusing because that's the error of a filepath not content of `.json`.

masterSplinter01:
I'm not sure is there a convenient way to differentiate invalid plain-string json and non-existing filepath, that's why `authmate` returns the same error for both cases. But text of the error when file cannot be read become clearer.
 
Closes #558

### DIFF
--- a/cmd/s3-authmate/main.go
+++ b/cmd/s3-authmate/main.go
@@ -185,7 +185,7 @@ func issueSecret() *cli.Command {
 			},
 			&cli.StringFlag{
 				Name:        "bearer-rules",
-				Usage:       "rules for bearer token as plain json string",
+				Usage:       "rules for bearer token (filepath or a plain json string are allowed)",
 				Required:    false,
 				Destination: &eaclRulesFlag,
 			},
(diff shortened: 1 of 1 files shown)


cmd/s3-authmate/main.go
```diff
@@ -440,9 +440,16 @@ func parsePolicies(val string) (authmate.ContainerPolicies, error) {
...
-	if err != nil {
+
+	var (
 		data = []byte(val)
+		err  error
+	)
+
+	if !json.Valid(data) {
+		if data, err = os.ReadFile(val); err != nil {
+			return nil, fmt.Errorf("coudln't read json file or its content is invalid")
```
KirillovDenis, line 451: Maybe we should use the following msg: `coudln't read json file or provided json is invalid`.
Since user will see this message when he provided invalid json string

alexvanin (dismissed):
See @KirillovDenis's comment.

KirillovDenis (approved):

PR #599, opened 2022-07-19
[#569] Refactor versioning in tree service

### LINKED ISSUE #569: Refactor storing of versioning setting in the tree service
Now we store versioning settings as two bools:
 https://github.com/nspcc-dev/neofs-s3-gw/blob/master/api/data/info.go#L48
```
BucketSettings struct {
  IsNoneStatus      bool
  VersioningEnabled bool                     `json:"versioning_enabled"`
  LockConfiguration *ObjectLockConfiguration `json:"lock_configuration"`
}
```
1. `IsNoneStatus` = true, `VersioningEnabled` = false -- unversioned (if a tree node has not been created)
2. `IsNoneStatus` = false, `VersioningEnabled` = false -- suspended
3. `IsNoneStatus`= false, `VersioningEnabled` = true -- versioned
4. `IsNoneStatus` = true, `VersioningEnabled` = true -- is not possible

Maybe we can store it as a string `Unversioned`, `Versioned`, `Suspended` or as int (line `enum`: 0 = `Unversioned`, 1 = `Versioned`, 2 = `Suspended`).

It will make code simpler. 

BUT it breaks backward compatibility.

masterSplinter01:
Closes #569

### DIFF
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -14,9 +14,9 @@ const (
 	bktCORSConfigurationObject         = ".s3-cors"
 	bktNotificationConfigurationObject = ".s3-notifications"
 
-	VerUnversioned = "Unversioned"
-	VerEnabled     = "Enabled"
-	VerSuspended   = "Suspended"
+	VersioningUnversioned = "Unversioned"
+	VersioningEnabled     = "Enabled"
+	VersioningSuspended   = "Suspended"
 )
 
 type (
@@ -96,13 +96,13 @@ func (o *ObjectInfo) Address() oid.Address {
 }
 
 func (b BucketSettings) Unversioned() bool {
-	return b.Versioning == VerUnversioned
+	return b.Versioning == VersioningUnversioned
 }
 
 func (b BucketSettings) VersioningEnabled() bool {
-	return b.Versioning == VerEnabled
+	return b.Versioning == VersioningEnabled
 }
 
 func (b BucketSettings) VersioningSuspended() bool {
-	return b.Versioning == VerSuspended
+	return b.Versioning == VersioningSuspended
 }
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -13,6 +13,10 @@ const (
 	bktSettingsObject                  = ".s3-settings"
 	bktCORSConfigurationObject         = ".s3-cors"
 	bktNotificationConfigurationObject = ".s3-notifications"
+
+	VerUnversioned = "Unversioned"
+	VerEnabled     = "Enabled"
+	VerSuspended   = "Suspended"
 )
 
 type (
@@ -45,8 +49,7 @@ type (
 
 	// BucketSettings stores settings such as versioning.
 	BucketSettings struct {
-		IsNoneStatus      bool
-		VersioningEnabled bool                     `json:"versioning_enabled"`
+		Versioning        string                   `json:"versioning"`
 		LockConfiguration *ObjectLockConfiguration `json:"lock_configuration"`
 	}
 
@@ -90,4 +93,16 @@ func (o *ObjectInfo) Address() oid.Address {
 	addr.SetObject(o.ID)
 
 	return addr
+}
+
+func (b BucketSettings) Unversioned() bool {
+	return b.Versioning == VerUnversioned
+}
+
+func (b BucketSettings) VersioningEnabled() bool {
+	return b.Versioning == VerEnabled
+}
+
+func (b BucketSettings) VersioningSuspended() bool {
+	return b.Versioning == VerSuspended
 }
--- a/api/handler/versioning.go
+++ b/api/handler/versioning.go
@@ -33,7 +33,7 @@ func (h *handler) PutBucketVersioningHandler(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	if configuration.Status != data.VerEnabled && configuration.Status != data.VerSuspended {
+	if configuration.Status != data.VersioningEnabled && configuration.Status != data.VersioningSuspended {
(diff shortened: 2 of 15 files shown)


api/handler/versioning.go
```diff
@@ -33,15 +33,18 @@ func (h *handler) PutBucketVersioningHandler(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	settings.VersioningEnabled = configuration.Status == "Enabled"
-	settings.IsNoneStatus = false
+	if configuration.Status != data.VerEnabled && configuration.Status != data.VerSuspended {
+		h.logAndSendError(w, "couldn't get bucket settings", reqInfo, err)
```
KirillovDenis, line 37: Fix log message please

api/layer/layer.go
```diff
@@ -523,7 +523,7 @@ func (n *layer) deleteObject(ctx context.Context, bkt *data.BucketInfo, settings
 			Created: time.Now(),
 			Owner:   n.Owner(ctx),
 		},
-		IsUnversioned: !settings.VersioningEnabled,
+		IsUnversioned: !settings.VersioningEnabled(),
```
KirillovDenis, line 526:
```suggestion
		IsUnversioned: settings.VersioningSuspended(),
```

api/layer/object.go
```diff
@@ -138,10 +138,14 @@ func (n *layer) objectGet(ctx context.Context, bktInfo *data.BucketInfo, objID o
 func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.ObjectInfo, error) {
 	own := n.Owner(ctx)
 
-	versioningEnabled := n.isVersioningEnabled(ctx, p.BktInfo)
+	bktSettings, err := n.GetBucketSettings(ctx, p.BktInfo)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get versioning settings obejct: %w", err)
```
KirillovDenis, line 143: Typo `obejct` -> `object`

api/handler/versioning.go
```diff
@@ -33,15 +33,18 @@ func (h *handler) PutBucketVersioningHandler(w http.ResponseWriter, r *http.Requ
...
+	}
+	settings.Versioning = configuration.Status
 
 	p := &layer.PutSettingsParams{
 		BktInfo:  bktInfo,
 		Settings: settings,
 	}
 
-	if !p.Settings.VersioningEnabled && bktInfo.ObjectLockEnabled {
+	if !p.Settings.VersioningEnabled() && bktInfo.ObjectLockEnabled {
```
KirillovDenis, line 47:
```suggestion
	if p.Settings.VersioningSuspended() && bktInfo.ObjectLockEnabled {
```

api/handler/versioning.go
```diff
@@ -77,13 +80,10 @@ func (h *handler) GetBucketVersioningHandler(w http.ResponseWriter, r *http.Requ
...
+	if settings.Unversioned() {
 		return res
 	}
-	if settings.VersioningEnabled {
-		res.Status = "Enabled"
-	} else {
-		res.Status = "Suspended"
-	}
+	res.Status = settings.Versioning
+
```
KirillovDenis, line 87: We can write:
```
res := &VersioningConfiguration{}
if !settings.Unversioned() {
	res.Status = settings.Versioning
}
return res	
```

internal/neofs/tree.go
```diff
@@ -260,18 +260,17 @@ func newPartInfo(node NodeResponse) (*data.PartInfo, error) {
...
 	settings := &data.BucketSettings{}
-
-	if versioningEnabledValue, ok := node.Get(versioningEnabledKV); ok {
-		if settings.VersioningEnabled, err = strconv.ParseBool(versioningEnabledValue); err != nil {
-			return nil, fmt.Errorf("settings node: invalid versioning: %w", err)
-		}
+	if versioningValue, ok := node.Get(versioningKV); ok {
+		settings.Versioning = versioningValue
+	} else {
+		settings.Versioning = data.VerUnversioned
```
KirillovDenis, line 273: We can write:
```go
settings := &data.BucketSettings{Versioning: data.VerUnversioned}
if versioningValue, ok := node.Get(versioningKV); ok {
	settings.Versioning = versioningValue
}
```

api/data/info.go
```diff
@@ -13,6 +13,10 @@ const (
 	bktSettingsObject                  = ".s3-settings"
 	bktCORSConfigurationObject         = ".s3-cors"
 	bktNotificationConfigurationObject = ".s3-notifications"
+
+	VerUnversioned = "Unversioned"
```
KirillovDenis, line 17: It seems we can use empty string here or drop this constant at all (since `Unversioned` string isn't used as I understand). So when we create new BucketSettings like:
```go 
bktSettings := &BucketSettings{}
```
we get settings with "Unversioned" status if we update the following function:
```go
func (b BucketSettings) Unversioned() bool {
	return len(b.Versioning) == 0
}
```

api/data/info.go
```diff
@@ -13,6 +13,10 @@ const (
...
+	VerEnabled     = "Enabled"
```
KirillovDenis, line 18: Can we use more clear name like `VersioningEnabled`?

api/data/info.go
masterSplinter01: You're right, but constant string `Unversioned` makes code more readable (someone who will read the code will not need know what does empty string mean). I'm not sure we should replace it by empty string or drop it at all.

KirillovDenis (approved):

alexvanin (approved):

PR #622, opened 2022-07-25
[#612] Make listing more robust

### LINKED ISSUE #612: Inconsistent behaviour at listing of one object
With parallel listing there is a bit inconsistency of having correct response. `test_bucket_listv2_both_continuationtoken_startafter` passes 50% of time.

```py
def test_bucket_listv2_both_continuationtoken_startafter():
    key_names = ['bar', 'baz', 'foo', 'quxx']
    bucket_name = _create_objects(keys=key_names)
    client = get_client()

    response1 = client.list_objects_v2(Bucket=bucket_name, StartAfter='bar', MaxKeys=1)
    next_continuation_token = response1['NextContinuationToken']

    response2 = client.list_objects_v2(Bucket=bucket_name, StartAfter='bar', ContinuationToken=next_continuation_token)
    eq(response2['ContinuationToken'], next_continuation_token)
    eq(response2['StartAfter'], 'bar')
    eq(response2['IsTruncated'], False)
    key_names2 = ['foo', 'quxx']
    keys = _get_keys(response2)
    eq(keys, key_names2)
```

response1 should contain `baz` object, however it returns `foo` sometimes. This happens because both `baz` and `foo` are processed in parallel (we trim output after receiving extra object). Sometimes `foo` processed first and returned in response.

We can:
- do not produce pool for 1 object
- limit number of request before sending requests (then we might return less than requested number of objects)

KirillovDenis:
Fix test `test_bucket_listv2_both_continuationtoken_startafter`

close #612

### DIFF
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -471,8 +471,11 @@ func (n *layer) getLatestObjectsVersions(ctx context.Context, p allObjectParams)
 	objects = make([]*data.ObjectInfo, 0, p.MaxKeys)
 
 	for obj := range objOutCh {
-		if len(objects) == p.MaxKeys { // todo reconsider stop condition
-			next = obj
+		// TODO (@kirillovdenis) : #612, #525 reconsider stop condition
+		// currently we handle 3 more items to reduce the likelihood of missing the last object in batch
+		// (potentially we can miss it because of pool of workers)
+		if len(objects) == p.MaxKeys+3 {
+			//next = obj
 			break
 		}
 		objects = append(objects, obj)
@@ -481,6 +484,11 @@ func (n *layer) getLatestObjectsVersions(ctx context.Context, p allObjectParams)
 	sort.Slice(objects, func(i, j int) bool {
 		return objects[i].Name < objects[j].Name
 	})
+
+	if len(objects) > p.MaxKeys {
+		next = objects[p.MaxKeys]
+		objects = objects[:p.MaxKeys]
+	}
 
 	return
 }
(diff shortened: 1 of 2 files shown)


KirillovDenis:
It seems we need to find more appropriate solution. For example we can think about pausing worker routine if it handled more than (1-2-5-10) object than others. Current PR just reduce likelihood of error.

api/layer/object.go
```diff
@@ -471,8 +471,11 @@ func (n *layer) getLatestObjectsVersions(ctx context.Context, p allObjectParams)
 	objects = make([]*data.ObjectInfo, 0, p.MaxKeys)
 
 	for obj := range objOutCh {
-		if len(objects) == p.MaxKeys { // todo reconsider stop condition
-			next = obj
+		// TODO (@kirillovdenis) : #612, #525 reconsider stop condition
```
alexvanin, line 474: Can you link https://github.com/nspcc-dev/neofs-s3-gw/issues/625?

alexvanin (approved):

api/layer/object.go
```diff
@@ -471,8 +471,11 @@ func (n *layer) getLatestObjectsVersions(ctx context.Context, p allObjectParams)
...
+		// currently we handle 3 more items to reduce the likelihood of missing the last object in batch
+		// (potentially we can miss it because of pool of workers)
+		if len(objects) == p.MaxKeys+3 {
+			//next = obj
```
masterSplinter01, line 478: May we delete it?

masterSplinter01 (approved):

alexvanin (approved):

PR #644, opened 2022-08-03
[#643] Fix tagging cache

### LINKED ISSUE #643: Cache for tagging doesn't work if `versionID` is blank
How to reproduce:
0. put breakpoint on `api/layer/tagging.go:19`
1. create a bucket
2. put an object with tags
3. get object tagging -- s3-gw cannot find anything in system cache, because it asks an item with empty `versionID`, but cache stores needed tags by key with filled `versionID` 
5. get object tagging again -- the same as in step 3

masterSplinter01:
Closes #643

### DIFF
--- a/api/auth/signer/v4/v4.go
+++ b/api/auth/signer/v4/v4.go
@@ -3,7 +3,7 @@
 // Provides request signing for request that need to be signed with
 // AWS V4 Signatures.
 //
-// Standalone Signer
+// # Standalone Signer
 //
 // Generally using the signer outside of the SDK should not require any additional
 // logic when using Go v1.5 or higher. The signer does this by taking advantage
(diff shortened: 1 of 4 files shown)


masterSplinter01:
This solution requires make one request to the tree service to find needed tagging in cache. But it seems simple and optimal to me. 
The second solution is to store tags in cache by keys with empty `versionID` -- it will tagging for the latest versions of objects, despite of is versioning enabled or not. But I don't like it because of the following reasons: 
1. we always put tagging node for specific version of object, that's why we will need to check bucket settings and if the specified version latest only to put tags to cache by correct key
2. we will have to attentively invalidate system cache record with tags while each call of `put-object`, `delete-object`, `complete-multipart-upload` => risk to place new bugs increases

KirillovDenis:
Shouldn't we check if `ObjectVersion` contains a valid `VersionID` and then keep using cache before `getNodeVersion` also.

masterSplinter01:
I think we don't need to mention it in changelog because it's an inner and too minor change

masterSplinter01:
I have no idea why lint issues appeared: https://github.com/nspcc-dev/neofs-s3-gw/pull/644/files#diff-c66d78b521142a630cb4bf2b528327203e9f47acb1bb741d73b5409be017f20a

KirillovDenis:
Update your local golangci-lint to  `v1.48.0`:
```shell
$ curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
```

and run in s3-gw dir:
```
$ golangci-lint run --fix
```

KirillovDenis (approved):

PR #646, opened 2022-08-04
[#595] Add SSE with customer key

### LINKED ISSUE #595: Customer encryption in S3 Gateway
Based on the pilot feedback, many application use customer keys to encrypt data in S3. Consider implementing such scheme in S3 Gateway.

Info: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html
Related #193

KirillovDenis:
This PR add Server Side Encryption with Customer key (SSE-C).

There are some notes:
* Encrypt [only payload](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html), not metadata. 
* Store actual object size in header. For multipart upload parts in tree service in meta has actual size rather than encrypted size.
* We cannot decrypt parts payload during `CompleteMultipartUpload` so we store parts as concatenation of encrypted parts rather than encrypted concatenation of parts. It leads to having a quite complicated decrypter in `layer.GetObject` which need to create a new `sio.DecryptReader` for every part in result object.
* Encryption available only when TLS is enabled.

close #595 


api/data/info.go
```diff
@@ -47,6 +48,14 @@ type (
 		Headers     map[string]string
 	}
 
+	// EncryptionInfo store parsed object encryption headers.
```
masterSplinter01, line 51: stores -> stores

api/layer/layer.go
```diff
@@ -259,6 +285,72 @@ func (f MsgHandlerFunc) HandleMessage(ctx context.Context, msg *nats.Msg) error
...
+	}
+}
+
+// Key returns encryption key as slice.
+func (p EncryptionParams) Key() []byte {
+	return p.customerKey[:]
+}
+
+// AESKey returns encryption key.
+func (p EncryptionParams) AESKey() AES256Key {
```
masterSplinter01, line 302: We don't use the function anywhere. Will we use it in the future?

api/layer/layer.go
```diff
@@ -259,6 +285,72 @@ func (f MsgHandlerFunc) HandleMessage(ctx context.Context, msg *nats.Msg) error
...
+func NewEncryptionParams(key AES256Key) EncryptionParams {
+	return EncryptionParams{
+		enabled:     true,
+		customerKey: key,
+	}
+}
+
+// Key returns encryption key as slice.
+func (p EncryptionParams) Key() []byte {
+	return p.customerKey[:]
```
masterSplinter01, line 298: I got why `EncryptionParams` has unexported fields, but can't we store `customerKey` as `[]byte`? 

As for me, type `AES256Key [32]byte` is redundant, and we cast it to a slice in all usages anyway.  And then we can make the fileds exported and drop methods. and `NewEncryptionParams` function.

api/layer/layer.go
```diff
@@ -259,6 +285,72 @@ func (f MsgHandlerFunc) HandleMessage(ctx context.Context, msg *nats.Msg) error
...
+func (p EncryptionParams) AESKey() AES256Key {
+	return p.customerKey
+}
+
+// Enabled returns true if key isn't empty.
+func (p EncryptionParams) Enabled() bool {
+	return p.enabled
+}
+
+// HMAC compute salted HMAC.
```
masterSplinter01, line 311: computes

api/errors/errors.go
```diff
@@ -141,6 +141,7 @@ const (
 	ErrSSEEncryptedObject
 	ErrInvalidEncryptionParameters
 	ErrInvalidSSECustomerAlgorithm
+	ErrInvalidEncryptionAlgorithm
```
masterSplinter01, line 144: Aren't `ErrInvalidSSECustomerAlgorithm` and `ErrInvalidEncryptionAlgorithm` the same?

api/layer/object.go
```diff
@@ -145,6 +147,42 @@ func MimeByFileName(name string) string {
...
+	r, err = sio.EncryptReader(r, sio.Config{MinVersion: sio.Version20, MaxVersion: sio.Version20, Key: key, CipherSuites: []byte{sio.AES_256_GCM}})
+	if err != nil {
+		return nil, 0, fmt.Errorf("couldn't create encrypter: %w", err)
+	}
+
+	return r, encSize, nil
+}
+
+func parseCompletedPartHeader(hdr string) (*Part, error) {
+	// partInfo[0] -- part number, partInfo[1] -- part size, partInfo[2] -- checksum
```
masterSplinter01, line 165: I'm not sure that it looks good that we have two copied code chunks. Can we call `parseCompletedPartHeader`  in `formUploadAttributes` and then cast `layer.Part` to `handler.Part`?

api/errors/errors.go
KirillovDenis: Almost. The new one is actual aws error https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html. I'll drop the old one.

api/layer/encryption/encryption.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	p.enabled = true
+	p.customerKey = make([]byte, aes256KeySize)
+	copy(p.customerKey, key)
+	return p, nil
+}
+
+// Key returns encryption key as slice.
+func (p Params) Key() []byte {
+	return p.customerKey[:]
```
masterSplinter01, line 82: `[:]` seems unnecessary, doesn't it?

api/layer/encryption/encryption.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+const (
+	blockSize     = 1 << 16 // 64KB
+	fullBlockSize = blockSize + 32
+	aes256KeySize = 32
+)
+
+// NewParams creates new params to encrypt with provided key.
+func NewParams(key []byte) (Params, error) {
```
masterSplinter01, line 69: Shouldn't `NewParams` return a pointer to `Params`?

api/layer/encryption/encryption.go
```diff
@@ -0,0 +1,344 @@
...
+	copy(p.customerKey, key)
+	return p, nil
+}
+
+// Key returns encryption key as slice.
+func (p Params) Key() []byte {
+	return p.customerKey[:]
+}
+
+// Enabled returns true if key isn't empty.
```
masterSplinter01, line 85: If we `enabled` depends on whether key is not empty, maybe we can drop `enable` and check in `Enable()` length of key?

api/handler/attributes.go
```diff
@@ -202,23 +214,14 @@ func formUploadAttributes(info *data.ObjectInfo, maxParts, marker int) (*ObjectP
...
-		}
-		num, err := strconv.Atoi(partInfo[0])
-		if err != nil {
-			return nil, err
-		}
-		size, err := strconv.Atoi(partInfo[1])
+		part, err := layer.ParseCompletedPartHeader(p)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("invalid competed part: %w", err)
```
masterSplinter01, line 219: A little typo - competed

api/handler/get.go
```diff
@@ -71,13 +72,25 @@ func overrideResponseHeaders(h http.Header, query url.Values) {
 	}
 }
 
-func writeHeaders(h http.Header, extendedInfo *data.ExtendedObjectInfo, tagSetLength int, isBucketUnversioned bool) {
+func addSSECHeaders(responseHeader http.Header, requestHeader http.Header) {
+	responseHeader.Set(api.AmzServerSideEncryptionCustomerAlgorithm, requestHeader.Get(api.AmzServerSideEncryptionCustomerAlgorithm))
+	responseHeader.Set(api.AmzServerSideEncryptionCustomerKeyMD5, requestHeader.Get(api.AmzServerSideEncryptionCustomerKeyMD5))
+}
+
+func writeHeaders(h http.Header, requestHeader http.Header, extendedInfo *data.ExtendedObjectInfo, tagSetLength int, isBucketUnversioned bool) {
```
masterSplinter01, line 80: It seems like the function has too many params, maybe we should group it into a struct? 

But I’m not insisting, it’s up to you

api/layer/encryption/encryption.go
```diff
@@ -0,0 +1,342 @@
...
+
+	partDataRemain  uint64
+	encPartRangeLen uint64
+
+	seqNumber uint64
+	decLen    uint64
+	skipLen   uint64
+
+	ln  uint64
+	off uint64
```
masterSplinter01, line 58: Maybe ‘offset’ is clearer?

api/layer/layer.go
```diff
@@ -7,8 +7,12 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"strconv"
+	"strings"
 	"time"
 
+	"github.com/nspcc-dev/neofs-s3-gw/api/layer/encryption"
+
```
masterSplinter01, line 15: Empty line

api/layer/util.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/nspcc-dev/neofs-s3-gw/api/layer/encryption"
+
```
masterSplinter01, line 13: Empty line

masterSplinter01 (approved):

PR #632, opened 2022-07-28
[#607] Wallet config

masterSplinter01:
Close #607

PR #647, opened 2022-08-04
Update CHANGELOG

masterSplinter01:
Related to #632

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 This document outlines major changes between releases.
 
+## [Unreleased]
+
+### Added
+- Exposure of pool metrics (#615)
+### Changed
+- Improved wallet configuration via `.yaml` config and environment variables (#607)
+### Removed
+### Fixed
+### Updating from v0.23.0
+1. Make sure configuration of wallet is valid. If you use:
+   1. environment variables: set `S3_GW_WALLET_PATH` instead of `S3_GW_WALLET`, 
+   set `S3_GW_WALLET_ADDRESS` instead of `S3_GW_ADDRESS`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 This document outlines major changes between releases.
 
+## [Unreleased]
+
+### Added
+- Exposure of pool metrics (#615)
+### Changed
+- Improved wallet configuration via `.yaml` config and environment variables (#607)
+### Removed
+### Fixed
```
KirillovDenis, line 12: I think we can add new section when we need it. Also we should add `Updating from v0.23.0` section, see https://github.com/nspcc-dev/neofs-s3-gw/blob/master/CHANGELOG.md#updating-from-v0220

masterSplinter01: I suppose adding of the most used sections in advance will make further appending of new changes a little more friendly. We can remove unused sections when we release.

masterSplinter01: Also core dev team fills changelog in the same way: https://github.com/nspcc-dev/neofs-node/blob/master/CHANGELOG.md

CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
+### Updating from v0.23.0
+1. Make sure configuration of wallet is valid. If you use:
+   1. environment variables: set `S3_GW_WALLET_PATH` instead of `S3_GW_WALLET`, 
+   set `S3_GW_WALLET_ADDRESS` instead of `S3_GW_ADDRESS`
+   2. `.yaml` config: to set wallet path use `wallet.path`,
```
KirillovDenis, line 17:
```suggestion
   2. `.yaml` config: to set wallet path use `wallet.path` instead of `wallet`, 
```

KirillovDenis (approved):

PR #655, opened 2022-08-13
Optimize listing

### LINKED ISSUE #616: Reduce number of request during object listing
It seems we can process name by using `data.NodeVesion` struct before trying to get objectinfo from cache or NeoFS. It can help to reduce unnecessary requests.

_Originally posted by @KirillovDenis in https://github.com/nspcc-dev/neofs-s3-gw/issues/613#issuecomment-1193640901_

### LINKED ISSUE #625: Optimize parallel listing strategy
> Current PR just reduce likelihood of error.

Agree. Let's think about it in separate issue. I thought about combining expectations and dynamic feedback to `nodesGenerator()`
1) User expects 5 items so `nodesGenerator()` generate exactly 5 node versions (I will call it jobs) into the channel
2) In no request's failing, then it is optimal solution
3) If some requests fail we should update strategy by giving feedback to `nodesGenerator()`:
  3.1) Re-request the same object in the same go-routine without feedback
  3.2) Produce feedback so `nodesGenerator()` will produce one more job to do.

This solution might slower than current implementation in case of network fails. However it should be robust and consistent when network is healthy. I think it is more important.

_Originally posted by @alexvanin in https://github.com/nspcc-dev/neofs-s3-gw/issues/622#issuecomment-1195407541_

KirillovDenis:
This PR:
* limit generator with `max-keys` value to avoid get extra objects
* if object couldn't be `HEAD` from NeoFS we form it using `data.NodeVersion` (it doesn't have some fields: `createdAt`, `owner` but the `name`, `size` and `etag` are still valid)
* check if object is directory before request to NeoFS

Close #625, close #616

### DIFF
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -682,19 +682,19 @@ func IsSystemHeader(key string) bool {
 }
 
 func shouldSkip(node *data.NodeVersion, p allObjectParams, existed map[string]struct{}) bool {
-	filepath := node.FilePath
-	if len(p.Delimiter) > 0 {
-		tail := strings.TrimPrefix(filepath, p.Prefix)
-		index := strings.Index(tail, p.Delimiter)
-		if index >= 0 {
-			filepath = p.Prefix + tail[:index+1]
-		}
-	}
-	if _, ok := existed[filepath]; ok {
+	if node.IsDeleteMarker() {
 		return true
 	}
 
-	if filepath <= p.Marker {
+	filePath := node.FilePath
+	if dirName := tryDirectoryName(node, p.Prefix, p.Delimiter); len(dirName) != 0 {
+		filePath = dirName
+	}
+	if _, ok := existed[filePath]; ok {
+		return true
+	}
+
+	if filePath <= p.Marker {
 		return true
 	}
 
@@ -707,7 +707,7 @@ func shouldSkip(node *data.NodeVersion, p allObjectParams, existed map[string]st
 		}
 	}
 
-	existed[filepath] = struct{}{}
+	existed[filePath] = struct{}{}
 	return false
 }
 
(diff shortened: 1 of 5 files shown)


api/layer/object.go
```diff
@@ -682,19 +699,19 @@ func IsSystemHeader(key string) bool {
...
+		return true
+	}
+
+	filePath := node.FilePath
+	if dirName := tryDirectoryName(node, p.Prefix, p.Delimiter); len(dirName) != 0 {
+		filePath = dirName
 	}
-	if _, ok := existed[filepath]; ok {
+	if _, ok := existed[filePath]; ok {
 		return true
```
masterSplinter01, line 711: Could you please provide an example when we go to this line?
I coudn't find a suitable example.

api/layer/object.go
```diff
@@ -596,12 +595,18 @@ func (n *layer) initWorkerPool(ctx context.Context, size int, p allObjectParams,
 				wg.Add(1)
 				err = pool.Submit(func() {
 					defer wg.Done()
-					if oi := n.objectInfoFromObjectsCacheOrNeoFS(ctx, p.Bucket, node, p.Prefix, p.Delimiter); oi != nil {
-						select {
-						case <-ctx.Done():
-						case objCh <- oi:
+					oi := n.objectInfoFromObjectsCacheOrNeoFS(ctx, p.Bucket, node, p.Prefix, p.Delimiter)
+					if oi == nil {
+						// try to get object again
```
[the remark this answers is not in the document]
KirillovDenis: `objectInfoFromObjectsCacheOrNeoFS` can fail `HEAD` request because of network problem. We should try request again before forming partial data.
Also see `3.1` in https://github.com/nspcc-dev/neofs-s3-gw/issues/625

api/layer/object.go
KirillovDenis: Let's say you have objects: `prefix/dir/obj1`, `prefix/dir/obj2`, `prefix/obj3` and you do listing with prefix: `prefix` and delimiter: `/`. In this case after you handle `prefix/dir/obj1` and get "directory" `prefix/dir/` you can skip handling all other objects with prefix `prefix/dir/` (in our case it's `prefix/dir/obj2`).
The final listing result will be:
```json
{
  "Content": [
    {
      "Key": "prefix/obj3"
    }
  ],
  "CommonPrefix": [
    {
      "Prefix": "prefix/dir/"
    }
  ]
}
```

masterSplinter01 (approved):

PR #688, opened 2022-08-29
Fix object head after put

### LINKED ISSUE #682: Fix test_get_obj_head_tagging
This simple test is failing. 
1) Upload object
2) Set two tags
3) Head object and check `x-amz-tagging-count` header == `2`

The reason is hidden in #577. After this PR at object put we cache `ExtendedObjectInfo`, which contains `NodeVersion`.
https://github.com/nspcc-dev/neofs-s3-gw/blob/163038b37d90e9fe17286bc450a573c8e2320107/api/layer/object.go#L289-L294

This `NodeVersion` is missing `ID` field. It isn't set because it is never returned from tree service in the first place.
https://github.com/nspcc-dev/neofs-s3-gw/blob/163038b37d90e9fe17286bc450a573c8e2320107/internal/neofs/tree.go#L1343

The same `NodeVersion` is going to be read from cache and used in head request.
https://github.com/nspcc-dev/neofs-s3-gw/blob/163038b37d90e9fe17286bc450a573c8e2320107/api/handler/head.go#L78

In the end we got here with `objVersion.ID == 0` which is incorrect.
https://github.com/nspcc-dev/neofs-s3-gw/blob/163038b37d90e9fe17286bc450a573c8e2320107/internal/neofs/tree.go#L1034

## Solutions

1) Keep caching only ObjectInfo without NodeVersion.
2) Get latest version from tree service, set it in `NodeVersion` of `ExtendedObjectInfo` **and then** store in cache.
3) Try to set ID in `NodeVersion` with return value of `tree.AddByPath`.

alexvanin:
Close #682 

This PR updates tree service interface to return added node id in `AddVersion` method. 

---

```
# nosetests -v --nologcapture -s s3tests_boto3.functional.test_s3:test_get_obj_head_tagging
Done with cleanup of buckets in tests.
Done with cleanup of buckets in tests.
Done with cleanup of buckets in tests.
s3tests_boto3.functional.test_s3.test_get_obj_head_tagging ... ok
```

### DIFF
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -594,7 +594,7 @@ func (n *layer) deleteObject(ctx context.Context, bkt *data.BucketInfo, settings
 		IsUnversioned: settings.VersioningSuspended(),
 	}
 
-	if obj.Error = n.treeService.AddVersion(ctx, bkt.CID, newVersion); obj.Error != nil {
+	if _, obj.Error = n.treeService.AddVersion(ctx, bkt.CID, newVersion); obj.Error != nil {
 		return obj
 	}
 
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -250,7 +250,7 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Object
 
 	newVersion.OID = id
 	newVersion.ETag = hex.EncodeToString(hash)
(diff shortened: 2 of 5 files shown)


KirillovDenis (approved):
LGTM

PR #689, opened 2022-08-29
[#684] Fix continuation token for directories

### LINKED ISSUE #684: Fix more listing errors
Keep fixing #501 

There are still issues with 
```
s3tests_boto3.functional.test_s3.test_bucket_listv2_delimiter_prefix_underscore
s3tests_boto3.functional.test_s3.test_bucket_listv2_delimiter_prefix 
```

Both tests are working fine with v1 listing. 

```
objects: 'asdf', 'boo/bar', 'boo/baz/xyzzy', 'cquux/thud', 'cquux/bla'
prefix: ''
delim: '/'
expected: 'asdf', 'boo/', 'cquux/'
```

The issue is that second listing returns empty result with `Truncated=False`. I think it is because second item is a dir (`boo/`) and dirs have empty object ID. It can't be used as continuation token.

KirillovDenis:
close #684 

### DIFF
--- a/api/handler/object_list_test.go
+++ b/api/handler/object_list_test.go
@@ -120,6 +120,35 @@ func TestS3BucketListDelimiterBasic(t *testing.T) {
 	require.Equal(t, "quux/", listV1Response.CommonPrefixes[1].Prefix)
 }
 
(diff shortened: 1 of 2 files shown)


alexvanin (approved):
That's some nice tests :+1:

PR #703, opened 2022-09-07
[#657] replace filename with filepath

### LINKED ISSUE #657: Replace `FileName` with `FilePath`
After https://github.com/nspcc-dev/neofs-api/issues/238 and when a new constant `object.FilePath` will be supported in https://github.com/nspcc-dev/neofs-sdk-go/issues/215 we should use the `FilePath` attribute to store object name

KirillovDenis:
**Note:** this is breaking change. We will not be able to work with object that were created with old s3-gw versions.
Though we can check if the `FilePath` exists and if not use `FileName`

close #657


KirillovDenis:
I'm not sure if we really need [this](https://github.com/KirillovDenis/neofs-s3-gw/blob/01da3e123c638de13de470361bd8951f50a46fa4/api/layer/util.go#L142). User can do nothing with object name that is actually `OID` but see such object in listing. On the other hand he can find out something using this OID with `neofs-cli` for example

alexvanin:
> I'm not sure if we really need [this](https://github.com/KirillovDenis/neofs-s3-gw/blob/01da3e123c638de13de470361bd8951f50a46fa4/api/layer/util.go#L142). User can do nothing with object name that is actually `OID` but see such object in listing. On the other hand he can find out something using this OID with `neofs-cli` for example

OID for more detailed investigation looks better than having no info about the object.

--

Don't you want to try neofs tests from s3-tests on this PR?

KirillovDenis:
> Don't you want to try neofs tests from s3-tests on this PR? Also listing tests would be very useful with all those trailing slash changes.

I checked some specific tests  for trailing slashes. But yes, we should run all tests

KirillovDenis:
Tests passed

alexvanin (approved):

PR #718, opened 2022-10-04
Bugfix/713 fix caches

### LINKED ISSUE #713: aws s3api list-objects might list objects in bucket owned by different credentials for a short period of time
Steps to reproduce:

1. Create two different wallets (let's say wallet_user1.json and wallet_user2.json).
2. Issue two sets of s3 keys by neofs-s3-authmate issue-secret - one for wallet_user1.json and another one for wallet_user2.json
3. Configure two aws clients using these two different sets of s3 keys.
4. Create bucket using user1's aws credentials:
`aws --no-verify-ssl s3api create-bucket --bucket acl-test-user1 --endpoint-url http://node1.neofs:8084`
5. List objects in this bucket using user2's aws credentials from another terminal  - an error should be received as an output. This is expected behaviour.
`aws --no-verify-ssl s3api list-objects --bucket acl-test-user1 --endpoint-url http://172.19.17.229:8084
An error occurred (InternalError) when calling the ListObjects operation (reached max retries: 2): failed to get sub tree: rpc error: code = Unknown desc = bearer token must be signed by the container owner`
6. Put any object in this bucket using user1's aws credentials:
`aws --no-verify-ssl s3api put-object --bucket acl-test-user1 --key user1-owner-data --body /tmp/data.txt --endpoint-url http://node1.neofs:8084`
7. List objects in this bucket using user1's aws credentials:
`aws --no-verify-ssl s3api list-objects --bucket acl-test-user1 --endpoint-url http://node1.neofs:8084`
8. Immediately after, list objects in this bucket using user2's aws credentials from another terminal:
`aws --no-verify-ssl s3api list-objects --bucket acl-test-user1 --endpoint-url http://172.19.17.229:8084`
9. This time command will return the list of objects for this owned by user1's bucket, although it was executed from user2's credentials.
10. You can re-run command above multiple times - each time it will return the list of objects.
11. This behaviour will persist for about 10-16 seconds.
12. After this period of time the command will start getting an error from step 5 again, as expected.
13. You can go and repeat all steps starting from step 6 to reproduce the issue.

NOTE:
a) If you skip step 7, command in step 8 will return an error, as it should.
b) If use any other endpoint in command from step 8, it will return an error, too.

Cache?

KirillovDenis:
close #713


masterSplinter01:
I'm not sure we need to fix it in this PR, but I found a bug during check of the PR. 
I have two aws credentials profiles: default and alt. I tried to get versioning  
```
$ aws s3api create-bucket --bucket test --acl private --endpoint-url http://127.0.0.1:8084   
$ aws s3api put-bucket-versioning --versioning-configuration Status=Enabled  --bucket test --endpoint-url http://127.0.0.1:8084 
$ aws s3api get-bucket-versioning  --endpoint-url http://127.0.0.1:8084 --bucket test --profile alt 

Connection was closed before we received a valid response from endpoint URL: "http://127.0.0.1:8084/test?versioning".
```
It happens because of nil pointer reference in `GetBucketVersioningHandler`:
```
...
settings, err := h.obj.GetBucketSettings(r.Context(), bktInfo)
	if err != nil {
		h.log.Warn("couldn't get version settings object: default version settings will be used",
			zap.String("request_id", reqInfo.RequestID),
			zap.String("method", reqInfo.API),
			zap.String("bucket_name", reqInfo.BucketName),
			zap.Error(err))
			// we got nil settings and an error (in my case, access to operation GET is denied), but we continue to go to the next line
	}

	if err = api.EncodeToResponse(w, formVersioningConfiguration(settings)); err != nil { //nil reference
		h.logAndSendError(w, "something went wrong", reqInfo, err)
	}
```

masterSplinter01:
Shouldn't we fix caches for system objects? 
1. created a `private` bucket, put  versioning settings into the bucket 
2. requested bucket settings with `alt` profile -- received nothing because of `access denied`
2. requested bucket setting with profile of owner of a bucket -- got settings
3. repeated  i.1 -- got settings  
```
aws s3api get-bucket-versioning  --endpoint-url http://127.0.0.1:8084 --bucket test --profile alt | jq                                                                                                               
{
  "Status": "Enabled"
}
```

api/layer/tagging.go
```diff
@@ -12,14 +12,11 @@ import (
 )
 
 func (n *layer) GetObjectTagging(ctx context.Context, p *ObjectVersion) (string, map[string]string, error) {
-	var (
-		err  error
-		tags map[string]string
-	)
+	var err error
```
masterSplinter01, line 15: It seems like we can drop this declaration

alexvanin:
@dansingjulia @KirillovDenis Can we check that default 1 minute cache does not break any existing test? I think we have quite hard time with caches in storage node, so just to be sure.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ This document outlines major changes between releases.
 
 ## [Unreleased]
 
+### Fixed
+- Unrestricted access to not owned objects via cache (#713)
```
alexvanin, line 8: Maybe also add `Update from` section and mention new config parameters?

KirillovDenis: I'm not sure we need such section because we don't change config params but add a new one with some default values

dansingjulia (approved):

dansingjulia:
tests passed

CHANGELOG.md
alexvanin: You described breaking config changes. But we want to notify readers that *some* config changes are present in new release. It is important for config manager tools that should adopt configuration of new parameter somehow.  The easiest way for us to notify config manager developers is to mention it in CHANGELOG file and use `config change` label on GitHub.

We can mention it in `Fixed` section near description or add separate record in `Add` section or describe it in `Update from`. Any way is okay.

alexvanin (approved):
Overall LGTM. See comments above.

masterSplinter01 (approved):

alexvanin (approved):

PR #739, opened 2022-11-03
Feature/568 configure region policy mapping

### LINKED ISSUE #568: Common region-policy settings
One can define AWS region mapping to storage policy in authbox, but we also need a way to setup a default region set in config or in some external config source.

With this feature app developers could use a known set of regions without diving into policy creation specifics.

KirillovDenis:
close #568

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -667,12 +667,10 @@ func parseMetadata(r *http.Request) map[string]string {
 }
 
 func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
-	var (
-		reqInfo = api.GetReqInfo(r.Context())
-		p       = layer.CreateBucketParams{
-			Name: reqInfo.BucketName,
-		}
-	)
+	reqInfo := api.GetReqInfo(r.Context())
+	p := &layer.CreateBucketParams{
+		Name: reqInfo.BucketName,
+	}
 
 	if err := checkBucketName(reqInfo.BucketName); err != nil {
 		h.logAndSendError(w, "invalid bucket name", reqInfo, err)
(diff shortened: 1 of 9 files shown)


config/config.yaml
```diff
@@ -110,10 +110,15 @@ nats:
...
-default_policy: REP 3
+# Parameters of NeoFS container placement policy
+placement_policy:
+  # Default policy of placing containers in NeoFS
+  # If a user sends a request `CreateBucket` and doesn't define policy for placing of a container in NeoFS, the S3 Gateway
+  # will put the container with default policy.
+  default: REP 3
+  # Region to placement policy mapping json file.
+  # Path to container policy mapping. The same as '--container-policy' flag for authmate
+  region_mapping: /path/to/container/policy.json
```
alexvanin, line 121: At first I thought about having mapping directly in config (see below), but I think it might provide some issues for JSON encoded policy. Also it is more consistent having the same mapping format for both authmate and gateway.

Anyway, I would like to see @realloc comment about that.

```yaml
placement_policy:
  region_mapping:
    rep3: REP 3
    complex: REP 1 IN X CBF 1 SELECT 1 FROM * AS X
```

api/handler/put.go
```diff
@@ -762,6 +747,27 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (h handler) setPolicy(prm *layer.CreateBucketParams, locationConstraint string, userPolicies []*accessbox.ContainerPolicy) {
+	prm.Policy = h.cfg.Policy.Default
+
+	if locationConstraint == "" {
+		return
+	}
+
+	if policy, ok := h.cfg.Policy.RegionMap[locationConstraint]; ok {
```
alexvanin, line 757: Make a comment here about parsing order. It is important to check default policies from config first, because they are controlled by system administrators and they are more relevant than immutable user policies from accessbox.

alexvanin, line 757: So it is guaranteed that `h.cfg.Policy.RegionMap` map contains at least default policy, right?

config/config.yaml
KirillovDenis: > At first I thought about having mapping directly in config

Me too. But in such case we couldn't set such mapping using env variables. (They cannot be fetched using `viper.GetStringMapString()`)

api/handler/put.go
KirillovDenis: Current order the following (every next override previous on conflict):

* default policy from config
* policies policy from config
* policies from user access box

Is it OK?

KirillovDenis: Not. It contains only policies from `placement_policy.region_mapping` config value

alexvanin: Oh, I've been mistaken in my first comment. I thought that policies from config have priority over access box, so I suggested to highlight it in the comment.

In fact they didn't.

alexvanin:
@KirillovDenis can you rebase and update changelog for merge?

alexvanin (approved):

PR #744, opened 2022-11-09
[#742] Add multiple listeners

### LINKED ISSUE #742: Listen on multiple sockets
Multiple IP addresses might be assigned to the single interface. Gateway should be able to listen on multiple addresses.

KirillovDenis:
close #742 

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -218,7 +218,7 @@ func (h *handler) PutObjectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	encryption, err := h.formEncryptionParams(r.Header)
+	encryptionParams, err := formEncryptionParams(r)
 	if err != nil {
 		h.logAndSendError(w, "invalid sse headers", reqInfo, err)
 		return
@@ -230,7 +230,7 @@ func (h *handler) PutObjectHandler(w http.ResponseWriter, r *http.Request) {
 		Reader:       r.Body,
 		Size:         r.ContentLength,
 		Header:       metadata,
-		Encryption:   encryption,
+		Encryption:   encryptionParams,
 		CopiesNumber: copiesNumber,
 	}
 
@@ -304,7 +304,7 @@ func (h *handler) PutObjectHandler(w http.ResponseWriter, r *http.Request) {
 	if settings.VersioningEnabled() {
 		w.Header().Set(api.AmzVersionID, objInfo.VersionID())
 	}
-	if encryption.Enabled() {
+	if encryptionParams.Enabled() {
 		addSSECHeaders(w.Header(), r.Header)
 	}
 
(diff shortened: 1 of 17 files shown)


docs/configuration.md
```diff
@@ -95,7 +95,7 @@ $ neofs-s3-gw --healthcheck_timeout 15s --connect_timeout 1m --rebalance_interva
 ### Monitoring and metrics
```
masterSplinter01, line 95: I think in `--help` and in line 49:
```
### Listening on address and TLS
You can make the gateway listen on specific address using the `--listen_address` option
```
we should mention that only one address can be set in this way. Or make it similar to `--peers`, but I'm not sure.

api/handler/put.go
```diff
@@ -326,16 +326,16 @@ func getCopiesNumberOrDefault(metadata map[string]string, defaultCopiesNumber ui
 	return uint32(copiesNumber), nil
 }
 
-func (h handler) formEncryptionParams(header http.Header) (enc encryption.Params, err error) {
-	sseCustomerAlgorithm := header.Get(api.AmzServerSideEncryptionCustomerAlgorithm)
-	sseCustomerKey := header.Get(api.AmzServerSideEncryptionCustomerKey)
-	sseCustomerKeyMD5 := header.Get(api.AmzServerSideEncryptionCustomerKeyMD5)
+func (h handler) formEncryptionParams(r *http.Request) (enc encryption.Params, err error) {
```
masterSplinter01, line 329: It seems like `h` is not used inside the function.

cmd/s3-gw/app.go
```diff
@@ -505,6 +432,40 @@ func (a *App) startServices() {
...
+			zap.String("tls cert", serverInfo.TLS.CertFile), zap.String("tls key", serverInfo.TLS.KeyFile))
+		a.servers[i] = newServer(ctx, serverInfo, a.log)
+	}
+}
+
+func (a *App) updateServers() error {
+	serversInfo := fetchServers(a.cfg)
+
+	if len(serversInfo) != len(a.servers) {
+		return fmt.Errorf("invalid servers configuration: length mismatch: old '%d', new '%d", len(a.servers), len(serversInfo))
```
masterSplinter01, line 451: Maybe "amount mismatch" would be clearer than "length mismatch"?

alexvanin (commented):
Rebase, please, and consider https://github.com/nspcc-dev/neofs-http-gw/pull/230#discussion_r1030575642 here as well.

KirillovDenis:
> Rebase, please, and consider [nspcc-dev/neofs-http-gw#230 (comment)](https://github.com/nspcc-dev/neofs-http-gw/pull/230#discussion_r1030575642) here as well.

I'm not sure we really need this. We have to make `cfgTLSCertFile` and `cfgTLSKeyFile` also slice parameters and it doesn't allow us to set several listen addresses with different tls settings (e.g. we want the first address be without tls and the second with tls) using command line args. So I suggest doesn't change the flags here.

alexvanin (approved):

PR #758, opened 2022-12-05
[#549] Move layer logic to handler

### LINKED ISSUE #549: Remove layer package
May be considered obsolete

KirillovDenis:
close #549 


smallhive:
It is a huge refactor. I consider it should solve something, but I can't figure out what. The original issue doesn't contain useful info for the goal. I think we can close PR and the issue

PR #762, opened 2022-12-12
[#754] Add CID and OID to logs

### LINKED ISSUE #754: Add `CID` and `OID` to logs
It would be nice to see in logs not only object name but its oid also. It can help debug issues quicker.
For bucket operations (e.g. `HeadBucket`) it would be nice to log CID. (Currently we just return it in response header)

ir0nbee:
Signed-off-by: Artem Tataurov <a.tataurov@yadro.com>

Close/Closes #754

Most handlers have the logging call on first levels.
Some of them rely on logging calls placed deeper at `getNodeVersion` and `getNodeVersionFromCache`: this is for cases when OID is not available for higher levels. Examples:
* DeleteObjectHandler
* DeleteMultipleObjectsHandler
* AbortMultipartUploadHandler
* PutObjectRetentionHandler
* GetObjectRetentionHandler
* PutObjectLegalHoldHandler
* GetObjectLegalHoldHandler
* ListMultipartUploadsHandler
* ListPartsHandler

### DIFF
--- a/api/handler/multipart_upload.go
+++ b/api/handler/multipart_upload.go
@@ -348,6 +348,13 @@ func (h *handler) UploadPartCopy(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	h.log.Debug("copy details",
+		zap.String("reqId", reqInfo.RequestID),
+		zap.String("bucket", reqInfo.BucketName),
+		zap.Stringer("cid", bktInfo.CID),
+		zap.String("object", reqInfo.ObjectName),
+		zap.Stringer("oid", info.ID))
+
 	response := UploadPartCopyResponse{
 		ETag:         info.HashSum,
 		LastModified: info.Created.UTC().Format(time.RFC3339),
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -482,11 +482,27 @@ func (n *layer) GetObjectInfo(ctx context.Context, p *HeadObjectParams) (*data.O
 
 // GetExtendedObjectInfo returns meta information and corresponding info from the tree service about the object.
 func (n *layer) GetExtendedObjectInfo(ctx context.Context, p *HeadObjectParams) (*data.ExtendedObjectInfo, error) {
+	var objInfo *data.ExtendedObjectInfo
+	var err error
+
 	if len(p.VersionID) == 0 {
-		return n.headLastVersionIfNotDeleted(ctx, p.BktInfo, p.Object)
+		objInfo, err = n.headLastVersionIfNotDeleted(ctx, p.BktInfo, p.Object)
+	} else {
+		objInfo, err = n.headVersion(ctx, p.BktInfo, p)
+	}
+	if err != nil {
+		return nil, err
 	}
 
-	return n.headVersion(ctx, p.BktInfo, p)
+	reqInfo := api.GetReqInfo(ctx)
+	n.log.Debug("get object",
+		zap.String("reqId", reqInfo.RequestID),
+		zap.String("bucket", p.BktInfo.Name),
+		zap.Stringer("cid", p.BktInfo.CID),
+		zap.String("object", objInfo.ObjectInfo.Name),
+		zap.Stringer("oid", objInfo.ObjectInfo.ID))
+
+	return objInfo, nil
 }
 
 // CopyObject from one bucket into another bucket.
@@ -645,7 +661,12 @@ func (n *layer) CreateBucket(ctx context.Context, p *CreateBucketParams) (*data.
 func (n *layer) ResolveBucket(ctx context.Context, name string) (cid.ID, error) {
 	var cnrID cid.ID
 	if err := cnrID.DecodeString(name); err != nil {
-		return n.resolver.Resolve(ctx, name)
+		if cnrID, err = n.resolver.Resolve(ctx, name); err != nil {
+			return cid.ID{}, err
+		}
+
+		reqInfo := api.GetReqInfo(ctx)
+		n.log.Info("resolve bucket", zap.String("reqId", reqInfo.RequestID), zap.String("bucket", name), zap.Stringer("cid", cnrID))
 	}
 
 	return cnrID, nil
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/minio/sio"
+	"github.com/nspcc-dev/neofs-s3-gw/api"
 	"github.com/nspcc-dev/neofs-s3-gw/api/data"
 	"github.com/nspcc-dev/neofs-s3-gw/api/errors"
 	"github.com/nspcc-dev/neofs-s3-gw/api/layer/encryption"
@@ -227,6 +228,13 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 	if err != nil {
 		return nil, err
 	}
+
+	reqInfo := api.GetReqInfo(ctx)
+	n.log.Debug("upload part",
+		zap.String("reqId", reqInfo.RequestID),
+		zap.String("bucket", bktInfo.Name), zap.Stringer("cid", bktInfo.CID),
+		zap.String("multipart upload", p.Info.UploadID),
+		zap.Int("part number", p.PartNumber), zap.String("object", p.Info.Key), zap.Stringer("oid", id))
 
 	partInfo := &data.PartInfo{
 		Key:      p.Info.Key,
@@ -535,6 +543,11 @@ func (n *layer) AbortMultipartUpload(ctx context.Context, p *UploadInfoParams) e
 	}
 
 	for _, info := range parts {
+		reqInfo := api.GetReqInfo(ctx)
+		n.log.Debug("part details",
+			zap.String("reqId", reqInfo.RequestID),
+			zap.String("bucket", p.Bkt.Name), zap.Stringer("cid", p.Bkt.CID),
+			zap.String("object", info.Key), zap.Stringer("oid", info.OID))
 		if err = n.objectDelete(ctx, p.Bkt, info.OID); err != nil {
 			n.log.Warn("couldn't delete part", zap.String("cid", p.Bkt.CID.EncodeToString()),
 				zap.String("oid", info.OID.EncodeToString()), zap.Int("part number", info.Number), zap.Error(err))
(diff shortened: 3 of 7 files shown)


masterSplinter01:
Could you add to the description a link to the issue?
```
Close/Closes #754
```

The `master` branch should be updated.

api/handler/acl.go
```diff
@@ -167,6 +167,8 @@ func (h *handler) GetBucketACLHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	h.log.Debug("target details", zap.String("cid", bktInfo.CID.String()))
```
masterSplinter01, line 170: It seems we log `cid` only after successful `getBucketAndCheckOwner`.

masterSplinter01, line 170: Isn't it too excessive to log every request processing (event successful)?

api/handler/acl.go
```diff
@@ -345,6 +353,10 @@ func (h *handler) PutObjectACLHandler(w http.ResponseWriter, r *http.Request) {
 		return
```
masterSplinter01, line 353: If `getBucketAndCheckOwner` was successful and `GetObjectInfo` wasn't the gw couldn't log at least`oid`.

api/handler/multipart_upload.go
```diff
@@ -296,6 +296,11 @@ func (h *handler) UploadPartCopy(w http.ResponseWriter, r *http.Request) {
 		Object:    srcObject,
 		VersionID: versionID,
 	})
+
+	h.log.Debug("source details",
+		zap.String("cid", srcBktInfo.CID.String()),
+		zap.String("oid", srcInfo.ID.String()))
```
masterSplinter01, line 302: `srcInfo` can be `nil` here

api/handler/versioning.go
```diff
@@ -25,6 +26,8 @@ func (h *handler) PutBucketVersioningHandler(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	h.log.Debug("target details", zap.String("cid", bktInfo.CID.String()))
```
masterSplinter01, line 29: It seems like you call `n.log.Debug("target details",...)` after `h.logAndSendError(...)` everywhere in handler. I think it would be better to append `cid` and `oid` to `additionals` and pass it as a function parameter. And log `cid`, `oid` in `h.logAndSendError`.

api/layer/multipart_upload.go
```diff
@@ -562,6 +564,9 @@ func (n *layer) ListParts(ctx context.Context, p *ListPartsParams) (*ListPartsIn
 	parts := make([]*Part, 0, len(partsInfo))
 
 	for _, partInfo := range partsInfo {
+		n.log.Debug("part details",
+			zap.String("cid", p.Info.Bkt.CID.String()),
+			zap.String("oid", partInfo.OID.String()))
```
masterSplinter01, line 569: The function can't return any err after line 559, so, `cid` and `oid` will be printed only if `ListParts` passes successfully.

api/layer/tagging.go
```diff
@@ -175,6 +176,14 @@ func (n *layer) getNodeVersion(ctx context.Context, objVersion *ObjectVersion) (
 		return nil, errors.GetAPIError(errors.ErrNoSuchKey)
 	}
 
+	if version != nil {
+		n.log.Debug("target details",
+			zap.String("cid", objVersion.BktInfo.CID.String()),
+			zap.String("oid", version.OID.EncodeToString()))
+	} else {
+		n.log.Debug("target details", zap.String("cid", objVersion.BktInfo.CID.String()))
```
masterSplinter01, line 184: I think it'd be better to use `zap.Fields`

api/handler/acl.go
roman-khimov: It should be OK for debug, NeoGo can log every P2P packet in this mode. But I'd rather use `zap.Stringer` here (and everywhere down below) to avoid `.String()` transformation for every request.

ir0nbee: @masterSplinter01 logging `cid` and (if applicable) `oid` on every call is the idea of the issue. However, the `debug` level saves us from doings this all the time.
@roman-khimov do you recommend the Stringer because of the specification? `Stringer constructs a field with the given key and the output of the value's String method. The Stringer's String method is called lazily.`

api/handler/versioning.go
ir0nbee: The idea was to make all methods do the logging as soon as cid/oid identified no matter of handler's result. And the `logAndSendError` gets called in case of error only.

api/layer/multipart_upload.go
ir0nbee: We can log `cid` very early for most handlers. I just don't know if we should the logging.
@alexvanin @KirillovDenis what do you think?

api/handler/acl.go
roman-khimov: To call `h.log.Debug()` we need its parameters, one of them is the result of `zap.String()` that takes two arguments, one of which is the result of `bktInfo.CID.String()` -> `String()` is always called in this case, with some associated allocation/base58 overhead.

Now with `Stringer()` we only have to convert `bktInfo.CID` to interface to get parameters, it's much easier to do that. And then IIUC zap is smart enough to check the message level first and then extract data from fields only if it's needed (and call `String()` in this case).

Granted, all of this probably isn't critical at the moment, but still we better use more optimal APIs when they're available (this inspired nspcc-dev/neo-go#2842 BTW).

api/handler/acl.go
```diff
@@ -167,6 +167,8 @@ func (h *handler) GetBucketACLHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	h.log.Debug("target details", zap.Stringer("cid", bktInfo.CID))
```
KirillovDenis, line 170: Actually we should log the bucket name also, otherwise we can miss mapping because of concurrent logs order.

Also, maybe we should use `zap.String("cid", bktInfo.CID.EncodeToString())`?

api/layer/tagging.go
```diff
@@ -175,6 +176,12 @@ func (n *layer) getNodeVersion(ctx context.Context, objVersion *ObjectVersion) (
 		return nil, errors.GetAPIError(errors.ErrNoSuchKey)
 	}
 
+	if version != nil {
+		n.log.Debug("target details", zap.Stringer("cid", objVersion.BktInfo.CID), zap.Stringer("oid", version.OID))
+	} else {
+		n.log.Debug("target details", zap.Stringer("cid", objVersion.BktInfo.CID))
+	}
```
KirillovDenis, line 183: Maybe we write just:
```
if version != nil && !version.IsDeleteMarker() {
		n.log.Debug("target details", zap.Stringer("cid", objVersion.BktInfo.CID), zap.Stringer("oid", version.OID), zap.String("bucket", objVersion.BktInfo.Name), zap.String("object", objVersion.ObjectName))
}		
```
?

api/handler/acl.go
masterSplinter01: I thought that in most cases printing of `oid` and `cid` is needed when something goes wrong. But ok :)

masterSplinter01: I thought that in most cases printing of oid and cid is needed when something goes wrong. But ok :)

api/layer/multipart_upload.go
KirillovDenis: It seems this is ok. If we couldn't get any parts we cannot print OID.

ir0nbee: The remaining code (code that goes after two possible errors) should not break at the moment - all the work gets done with the data we already have, and it does not involve any operations with side effects.

api/handler/acl.go
ir0nbee: If an error takes place and objInfo is `nil` then the handler will call logAndSendError and return. The log won't happen plus we will have details on the problem.

api/handler/acl.go
```diff
@@ -167,6 +167,8 @@ func (h *handler) GetBucketACLHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	h.log.Debug("target details", zap.String("container", reqInfo.BucketName), zap.Stringer("cid", bktInfo.CID))
```
KirillovDenis, line 170: I would use the `bucket` instead of `container`

api/handler/acl.go
masterSplinter01: Sorry, it was a typo. Fixed:
If getBucketAndCheckOwner was successful and GetObjectInfo wasn't the gw couldn't log at least `cid`.

ir0nbee: 1) the logging is gone from some handlers now: Denis suggested moving it into a couple of internal methods.
2) There are other places where your comment is applicable though. And again, `logAndSendError` saves us there: if we run into errors before the log.Debug gets called then the needed details will be in error logs. However, for cases of valid CID+OID we will have our Debug log line in case of errors that happen further.

api/handler/acl.go
ir0nbee: Ok, will fix that too.

api/layer/layer.go
```diff
@@ -482,11 +482,24 @@ func (n *layer) GetObjectInfo(ctx context.Context, p *HeadObjectParams) (*data.O
...
+	} else {
+		objInfo, err = n.headVersion(ctx, p.BktInfo, p)
+	}
+
+	if err == nil {
+		n.log.Debug("get object",
+			zap.String("container", p.BktInfo.Name),
+			zap.Stringer("cid", p.BktInfo.CID),
+			zap.String("object", objInfo.ObjectInfo.Name),
+			zap.Stringer("oid", objInfo.ObjectInfo.ID))
```
KirillovDenis, line 499:
```suggestion
       if err != nil {
		return nil, err
	}

	n.log.Debug("get object",
		zap.String("bucket", p.BktInfo.Name),
		zap.Stringer("cid", p.BktInfo.CID),
		zap.String("object", objInfo.ObjectInfo.Name),
		zap.Stringer("oid", objInfo.ObjectInfo.ID))

	return objInfo, nil
```

api/handler/acl.go
masterSplinter01: > We already log CID of bucket on creating, and can get it using HeadBucket request

As for me, it doesn't seem convenient

KirillovDenis:
I would prefer to simplify this PR to [this](https://github.com/nspcc-dev/neofs-s3-gw/compare/master...KirillovDenis:neofs-s3-gw:feature/754-add_logging_cid/oid)

api/handler/multipart_upload.go
```diff
@@ -296,6 +296,7 @@ func (h *handler) UploadPartCopy(w http.ResponseWriter, r *http.Request) {
 		Object:    srcObject,
 		VersionID: versionID,
 	})
+
```
KirillovDenis, line 299: Why?

masterSplinter01 (approved):

alexvanin (approved):

PR #765, opened 2022-12-21
[#754] Remove logging from ListParts

### LINKED ISSUE #754: Add `CID` and `OID` to logs

ir0nbee:
Closes #754 

Removed logging from ListParts at `layer/multipart_upload.go` since more detailed logging happens in `getUploadParts` that gets called in ListParts.

Signed-off-by: Artem Tataurov <a.tataurov@yadro.com>

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -730,8 +730,8 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	h.log.Debug("target details", zap.String("reqId", reqInfo.RequestID),
-		zap.String("bucket", reqInfo.BucketName), zap.Stringer("cid", bktInfo.CID))
+	h.log.Info("bucket is created", zap.String("reqId", reqInfo.RequestID),
+		zap.String("bucket", reqInfo.BucketName), zap.Stringer("container_id", bktInfo.CID))
 
 	if p.ObjectLockEnabled {
 		sp := &layer.PutSettingsParams{
@@ -744,8 +744,6 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-
-	h.log.Info("bucket is created", zap.Stringer("container_id", bktInfo.CID))
 
 	api.WriteSuccessResponseHeadersOnly(w)
 }
--- a/api/handler/multipart_upload.go
+++ b/api/handler/multipart_upload.go
@@ -348,13 +348,6 @@ func (h *handler) UploadPartCopy(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	h.log.Debug("copy details",
-		zap.String("reqId", reqInfo.RequestID),
-		zap.String("bucket", reqInfo.BucketName),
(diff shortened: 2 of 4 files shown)


KirillovDenis:
It seems we can also remove 
https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/layer/multipart_upload.go#L546-L550
for the same reason

KirillovDenis:
And if we start drop redundant logging then let's drop the following:
* [this](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/handler/multipart_upload.go#L351-L356) because of [that](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/layer/multipart_upload.go#L233-L237)
* [this](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/handler/put.go#L733) because of [that](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/handler/put.go#L748) (or we can merge them into one log with `info` level right after bucket was created) 
* [this](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/handler/tagging.go#L55), [this](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/handler/tagging.go#L109) and [this](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/handler/tagging.go#L145)  because of [that](https://github.com/nspcc-dev/neofs-s3-gw/blob/2b905283f65a330714e38955b699a1d2abeca6bb/api/layer/tagging.go#L182)

api/handler/put.go
```diff
@@ -745,7 +742,7 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	h.log.Info("bucket is created", zap.Stringer("container_id", bktInfo.CID))
+	h.log.Info("bucket is created", zap.String("reqId", reqInfo.RequestID), zap.Stringer("container_id", bktInfo.CID))
```
KirillovDenis, line 745: Let's add log bucket name also. And place this between line 731 and 733. It allows to see created container even if we couldn't enable versioning.

alexvanin (approved):

api/handler/put.go
KirillovDenis: > And place this between line 731 and 733. It allows us to see created container even if we couldn't enable versioning.

What about this ?

ir0nbee: Somehow  missed it. Not the logging call is in the place.

KirillovDenis (approved):

PR #766, opened 2023-01-18
[#766] Require only one healthy server

KirillovDenis:
Signed-off-by: Denis Kirillov <d.kirillov@yadro.com>


roman-khimov (changes requested):
Same story, nspcc-dev/neofs-http-gw#244, nspcc-dev/neo-go#2878.

realloc:
Same story, @roman-khimov, you reference wrong model irrelevant here.

PR #768, opened 2023-02-09
acl: Make some code refactoring

cthulhu-rider:
Also fix `AccessControlPolicy` XML.

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -69,13 +69,15 @@ const (
 	awsPermRead        amazonS3Permission = "READ"
 )
 
-// GranteeType is aws grantee permission type constants.
-type GranteeType string
+// enum of Amazon S3 ACL permission grantees.
+type granteeType string
 
+// various grantee types defined in
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_Grantee.html
 const (
-	acpCanonicalUser         GranteeType = "CanonicalUser"
-	acpAmazonCustomerByEmail GranteeType = "AmazonCustomerByEmail"
-	acpGroup                 GranteeType = "Group"
+	granteeCanonicalUser         granteeType = "CanonicalUser"
+	granteeAmazonCustomerByEmail granteeType = "AmazonCustomerByEmail"
+	granteeGroup                 granteeType = "Group"
 )
 
 type bucketPolicy struct {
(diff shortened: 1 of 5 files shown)


api/handler/acl.go
```diff
@@ -56,13 +56,17 @@ const (
 	s3GetObjectVersion           = "s3:GetObjectVersion"
 )
 
-// AWSACL is aws permission constants.
-type AWSACL string
```
roman-khimov, line 60: Doesn't seem to be very useful.

api/handler/acl.go
```diff
@@ -56,13 +56,17 @@ const (
...
+type amazonS3Permission string
 
+// various access permissions defined in
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_Grant.html.
+//
+// See also https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#permissions.
 const (
-	aclFullControl AWSACL = "FULL_CONTROL"
-	aclWrite       AWSACL = "WRITE"
-	aclRead        AWSACL = "READ"
```
roman-khimov, line 65: Especially these renames.

api/handler/acl.go
```diff
@@ -69,13 +69,15 @@ const (
 	awsPermRead        amazonS3Permission = "READ"
 )
 
-// GranteeType is aws grantee permission type constants.
-type GranteeType string
```
roman-khimov, line 69: Can also be omitted, s3-gw is not something that is supposed to be reused package-by-package.

api/handler/acl.go
cthulhu-rider: I can't understand what `acl` stands for - Amazon S3 or NeoFS.

api/handler/acl.go
cthulhu-rider: It's not AWS ACL, it's permissions. Also it is unused outside the package, and has couple of unexported constants. I wanna realize all these things from the definition.

api/handler/acl.go
cthulhu-rider: Then there is no sense to export these things.

api/handler/acl.go
cthulhu-rider: Btw all values are defined in https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3#pkg-constants. I don't know why we define them locally.

KirillovDenis (approved):

api/handler/acl.go
roman-khimov: I guess the main problem is that this package has _a lot_ more inside.

roman-khimov (approved):
In general, please avoid renamings unless they're strictly required. Still, this improves things, so can be accepted.

PR #803, opened 2023-07-19
Rename internal errors to s3errors

### LINKED ISSUE #786: Rename api/errors to api/s3errors
Using standard package name is never a good idea and it's all about S3 errors, thus `s3errors`.

smallhive:
close #786

### DIFF
--- a/api/s3errors/errors.go
+++ b/api/s3errors/errors.go
@@ -1,4 +1,4 @@
-package errors
+package s3errors
 
 import (
 	"fmt"
@@ -270,7 +270,7 @@ const (
 
 	ErrPostPolicyConditionInvalidFormat
 
-	//CORS configuration errors.
+	// CORS configuration errors.
 	ErrCORSUnsupportedMethod
 	ErrCORSWildcardExposeHeaders
 )
--- a/syncTree.sh
+++ b/syncTree.sh
@@ -6,7 +6,7 @@ REVISION="feaa9eace7098c343598bf08fb50746a1e8d2deb"
 
 echo "tree service revision ${REVISION}"
 
-FILES=$(curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | sed -n "s,.*\"/nspcc-dev/neofs-node/blob/${REVISION}/pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p")
+FILES=$(curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | jq | sed -n "s,.*\"pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p")
 
 for file in $FILES; do
   if [[ $file == *"neofs"* ]]; then
--- a/api/auth/center.go
+++ b/api/auth/center.go
@@ -18,7 +18,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	v4 "github.com/nspcc-dev/neofs-s3-gw/api/auth/signer/v4"
 	"github.com/nspcc-dev/neofs-s3-gw/api/cache"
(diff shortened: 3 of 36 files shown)


syncTree.sh
```diff
@@ -6,9 +6,10 @@ REVISION="feaa9eace7098c343598bf08fb50746a1e8d2deb"
 
 echo "tree service revision ${REVISION}"
 
-FILES=$(curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | sed -n "s,.*\"/nspcc-dev/neofs-node/blob/${REVISION}/pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p")
+FILES=$(curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | jq | sed -n "s,.*\"pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p")
 
 for file in $FILES; do
+  echo ${file}
```
roman-khimov, line 12: What problem does it solve?

syncTree.sh
```diff
@@ -6,9 +6,10 @@ REVISION="feaa9eace7098c343598bf08fb50746a1e8d2deb"
 
 echo "tree service revision ${REVISION}"
 
-FILES=$(curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | sed -n "s,.*\"/nspcc-dev/neofs-node/blob/${REVISION}/pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p")
+FILES=$(curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | jq | sed -n "s,.*\"pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p")
```
roman-khimov, line 9: What's the effect?

api/s3errors/errors.go
```diff
@@ -270,7 +270,7 @@ const (
 
 	ErrPostPolicyConditionInvalidFormat
 
-	//CORS configuration errors.
+	// CORS configuration errors.
```
roman-khimov, line 273: Not related, btw.

syncTree.sh
smallhive: Without these changes syncTree doesn't work - it doesn't find any files in _REVISION_.
I tend to think, GH changed the response format a little bit:
* shorter files pass, removed `/nspcc-dev/neofs-node/blob/${REVISION}` from file path
* and JSON is not formatted - just one string.

Unfortunately, I don't know how it was before, but without `jq` sed finds only one file. With formatted JSON - all files. How I understand it checks input by each row

syncTree.sh
smallhive: None, debug info, removed

api/s3errors/errors.go
smallhive: gofmt

syncTree.sh
smallhive:
```shell
$ REVISION="feaa9eace7098c343598bf08fb50746a1e8d2deb"

Without jq
$ curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | sed -n "s,.*\"pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p"
types_neofs.pb.go

With jq
$ curl -s https://github.com/nspcc-dev/neofs-node/tree/${REVISION}/pkg/services/tree | jq | sed -n "s,.*\"pkg/services/tree/\(.*\.pb\.go\)\".*,\1,p"
service.pb.go
service_grpc.pb.go
service_neofs.pb.go
types.pb.go
types_neofs.pb.go

api/s3errors/errors.go
roman-khimov: I know, but it's a whitespace change -> separate commit.

roman-khimov (approved):

PR #807, opened 2023-08-03
Upgrade to sdk rc10

### LINKED ISSUE #806: Upgrade to SDK RC10
Upgrade dependency once it's out.

smallhive:
closes #806


cmd/s3-gw/app.go
```diff
@@ -184,36 +187,40 @@ func (a *App) initAPI(ctx context.Context) {
...
-		a.log.Fatal("failed to create resolver", zap.Error(err))
+		a.log.Fatal("rpcclient", zap.Error(err))
 	}
-}
 
-func (a *App) getResolverConfig() ([]string, *resolver.Config) {
-	resolveCfg := &resolver.Config{
-		NeoFS:      neofs.NewResolverNeoFS(a.pool),
-		RPCAddress: a.cfg.GetString(cfgRPCEndpoint),
+	nnsHash, err := sidechain.SystemContractHash(cl, sidechain.NNSContract)
```
smallhive, line 201: Do we have a similar functionality inside contracts? I mean resolving system contract hashes

cmd/s3-gw/app_metrics.go
```diff
@@ -195,25 +195,26 @@ func (m *poolMetricsCollector) register() {
...
 	m.requestDuration.Reset()
 
-	for _, node := range stat.Nodes() {
+	for _, node := range st.Nodes() {
 		m.overallNodeErrors.WithLabelValues(node.Address()).Set(float64(node.OverallErrors()))
 		m.overallNodeRequests.WithLabelValues(node.Address()).Set(float64(node.Requests()))
 
-		m.currentErrors.WithLabelValues(node.Address()).Set(float64(node.CurrentErrors()))
+		// ???????
+		// m.currentErrors.WithLabelValues(node.Address()).Set(float64(node.CurrentErrors()))
```
smallhive, line 210: New SDK stat doesn't have _CurrentErrors_. It is a part of the pool internal logic for each client heal checks. Should we:
1. ignore _currentErrors_
2. fill it with `OverallErrors`
3. Somehow pass these values to statistics?

cmd/s3-gw/app.go
roman-khimov: It's a part of a regular RPC client, so this shouldn't be in contracts.

api/handler/handlers_test.go
```diff
@@ -85,7 +85,7 @@ func prepareHandlerContext(t *testing.T) *handlerContext {
 
 	layerCfg := &layer.Config{
 		Caches:      layer.DefaultCachesConfigs(zap.NewExample()),
-		AnonKey:     layer.AnonymousKey{Key: key},
+		GateKey:     key,
```
roman-khimov, line 87: It's supposed to be a feature, see #271. Although I don't quite understand it.

api/layer/layer.go
```diff
@@ -307,26 +308,24 @@ func TimeNow(ctx context.Context) time.Time {
...
-	if err := user.IDFromKey(&ownerID, n.gateKey.PublicKey().Bytes()); err != nil {
-		panic(fmt.Errorf("id from key: %w", err))
-	}
+	ownerID.SetScriptHash(n.gateKey.PublicKey().GetScriptHash())
 
 	return ownerID
 }
 
 func (n *layer) prepareAuthParameters(ctx context.Context, prm *PrmAuth, bktOwner user.ID) {
+	prm.PrivateKey = &n.gateKey.PrivateKey
```
roman-khimov: Why are you moving this? Previous logic was different.

cmd/s3-gw/app_metrics.go
roman-khimov: `current` as in "for the current connection"? Monitoring sw can track differences in `overall` counters, but it can't different how many reconnects have happened in-between of those. I'm not sure it has any value though, so I'd probably just ignore it.

api/handler/multipart_upload.go
```diff
@@ -443,6 +443,12 @@ func (h *handler) CompleteMultipartUploadHandler(w http.ResponseWriter, r *http.
 			h.logAndSendError(w, "could not translate acl of completed multipart upload to ast", reqInfo, err, additional...)
 			return
 		}
+
+		if sessionTokenSetEACL, err = getSessionTokenSetEACL(r.Context()); err != nil {
```
roman-khimov, line 447: Is there any corresponding bug?

cmd/s3-gw/app.go
smallhive: Ok, I don't mind having such a function

cmd/s3-gw/app_metrics.go
smallhive: I think `current` is for the exact node. Ok, I will remove

internal/neogo/sidechain/sidechain.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+)
+
+const (
+	NNSContract = int32(1)
+)
+
+// SystemContractHash allows to get system contract hash.
+func SystemContractHash(cl *rpcclient.Client, id int32) (util.Uint160, error) {
```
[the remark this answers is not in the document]
smallhive: Ok, removed the package and moved it next to the initialization

api/resolver/resolver.go
```diff
@@ -4,195 +4,49 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
+	"math/big"
 
-	"github.com/nspcc-dev/neofs-sdk-go/container"
+	"github.com/nspcc-dev/neofs-contract/nns"
```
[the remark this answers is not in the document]
smallhive: :heavy_check_mark:  https://github.com/nspcc-dev/neofs-contract/pull/347

api/handler/handlers_test.go
smallhive: On current `master` it doesn't work:
```
$ aws s3api create-bucket --bucket $BUCKET --endpoint http://localhost:9080 --no-sign-request

An error occurred (InternalError) when calling the CreateBucket operation (reached max retries: 0): We encountered an internal error, please try again.
```
In the gate logs:
```
2023-08-04T07:42:20.324+0400    error   handler/util.go:29      call method     {"status": 500, "request_id": "dc3da4ba-b5b0-42aa-800f-2dbb71b80ddc", "method": "CreateBucket", "bucket": "heh3", "object": "", "description": "couldn't get bearer token signature key", "error": "couldn't get box data from context"}
```

In any case, we have to sign requests. Many client calls requires `Signer` as an explicit parameter. I tend to thing doesn't mater what key

api/layer/layer.go
smallhive: previous logic uses or bearer token or private key. The bearer token had more weight. Right now this `prm.PrivateKey` is used to sign client requests as explicit parameter

api/handler/multipart_upload.go
smallhive: Yes, we have session token, and pass session token, but this token is always empty. 
It leads to an infinity wait for the `setEacl` result. But it fails, because the gate isn't a container owner, thus we have a session, but it's empty

api/handler/handlers_test.go
roman-khimov: It can be used for reads, I guess.

smallhive:
Added anonymous key using for the `containerList` operation

smallhive:
Update to the latest rc-10 SDK and neofs-contract versions

api/resolver/resolver.go
```diff
@@ -4,195 +4,47 @@ import (
...
-type BucketResolver struct {
-	mu        sync.RWMutex
-	resolvers []*Resolver
-}
+var (
+	// ErrNotFound shows the container not found by name.
+	ErrNotFound = errors.New("not found")
+)
 
+// Resolver allows to resolve container id by it's name.
```
roman-khimov: its

api/layer/container.go
```diff
@@ -84,6 +84,11 @@ func (n *layer) containerList(ctx context.Context) ([]*data.BucketInfo, error) {
 		res []cid.ID
 		rid = api.GetRequestID(ctx)
 	)
+
+	if api.IsAnonymousRequest(ctx) {
```
roman-khimov, line 88: Why only this request? I think we can have the same in other ones like object get.

cmd/s3-gw/app.go
```diff
@@ -189,31 +197,45 @@ func (a *App) initMetrics() {
 	a.metrics = newAppMetrics(a.log, gateMetricsProvider, a.cfg.GetBool(cfgPrometheusEnabled))
 }
 
-func (a *App) initResolver() {
-	var err error
-	a.bucketResolver, err = resolver.NewBucketResolver(a.getResolverConfig())
+func (a *App) initResolver(ctx context.Context) {
```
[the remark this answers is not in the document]
smallhive: I left func `initResolver` here, but reworked the common resolvers constructor. Depending on `endpoint` it will return either NoOpResolver or NNSResolver

api/resolver/resolver.go
smallhive: fixed

api/layer/container.go
smallhive: Moved this check to the old place, inside `Owner` func. After that it will work like previous implementation

cmd/s3-gw/app.go
```diff
@@ -483,10 +460,6 @@ func (a *App) configReload() {
 		return
 	}
 
-	if err := a.bucketResolver.UpdateResolvers(a.getResolverConfig()); err != nil {
```
roman-khimov, line 486: Same thing as with HTTP, we still need a way to change config in runtime.

api/layer/anonymous.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+// Signer returns signer for the corresponding key.
+func (a Anonymous) Signer() user.Signer {
+	return a.signer
+}
+
+// Owner returns [user.ID] for the corresponding key.
+func (a Anonymous) Owner() user.ID {
```
roman-khimov, line 30: Do we need it if it's just another way to say `UserID()`? I think `Anonymous` can be replaced with `user.ID` completely.

api/layer/container.go
```diff
@@ -84,6 +84,7 @@ func (n *layer) containerList(ctx context.Context) ([]*data.BucketInfo, error) {
 		res []cid.ID
 		rid = api.GetRequestID(ctx)
 	)
+
```
roman-khimov, line 87: Oops

api/layer/container.go
roman-khimov: But what about keys? I suppose we should use a different key when we're doing anonymous requests. Maybe `layer`  should return `user.Signer` now to provide both at the same time (gw normally and anon if no auth)?

internal/neofs/neofs.go
```diff
@@ -255,18 +250,14 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
...
 
 	if prm.BearerToken != nil {
-		prmPut.UseBearer(*prm.BearerToken)
-	} else if prm.PrivateKey != nil {
-		prmPut.UseSigner(neofsecdsa.SignerRFC6979(*prm.PrivateKey))
+		opts.SetBearerToken(*prm.BearerToken)
 	}
 
-	idObj, err := x.pool.PutObject(ctx, prmPut)
+	idObj, err := slicer.Put(ctx, x.pool, *obj, x.gateSigner, prm.Payload, opts)
```
roman-khimov, line 260: Shouldn't be a slicer until #808.

api/layer/container.go
smallhive: I revised the code one more time.
We have the next component chain (gate component names):
* layer -> neofs -> pool

`layer` - handles API requests, calls `neofs` component to prepare data and then it call `pool` to execute request in neofs node.

In fact, the layer even doesn't need to know something about a key or signer. We will drop the key from it.
The _neofs_ component stores the same key (pool also stores it) and it is a good place to make magic with anonymous keys. We just need to pass isAnonymous flag to _neofs_ component and anonymous signer will be used

smallhive: Made changes in separate commits. At first look, I replaced all places when we cat override the signer. In other places, random keys are already in use by SDK client.

If these changes ok, even after few iterations, I will squash them to proper commit, if required

api/resolver/resolver.go
```diff
@@ -4,195 +4,66 @@ import (
...
-	resolve func(context.Context, string) (cid.ID, error)
+// NewNNSResolver is a constructor for the NNSResolver.
+func NewNNSResolver(reader *rpcNNS.ContractReader) *NNSResolver {
+	return &NNSResolver{reader: reader}
 }
 
-func (r *Resolver) SetResolveFunc(fn func(context.Context, string) (cid.ID, error)) {
-	r.resolve = fn
-}
+// Resolve looks up the container id by it's name via NNS contract.
```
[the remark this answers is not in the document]
smallhive: updated

cmd/s3-gw/app.go
smallhive: returned

api/layer/anonymous.go
smallhive: Replaced with user.ID

api/layer/container.go
smallhive: Removed

internal/neofs/neofs.go
```diff
@@ -212,6 +203,11 @@ func (x *NeoFS) DeleteContainer(ctx context.Context, id cid.ID, token *session.C
 
 // CreateObject implements neofs.NeoFS interface method.
 func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oid.ID, error) {
+	ni, err := x.pool.NetworkInfo(ctx, client.PrmNetworkInfo{})
```
roman-khimov, line 217: This is rather expensive to do per-object.

internal/neofs/neofs.go
```diff
@@ -278,6 +270,32 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
 		return oid.ID{}, fmt.Errorf("save object via connection pool: %w", err)
 	}
 
+	bChunk := make([]byte, ni.MaxObjectSize())
+	var n int
+	var idObj oid.ID
+
+	for {
```
roman-khimov, line 288: The same question about `io.Copy` is relevant here as well.

api/layer/layer.go
```diff
@@ -39,15 +39,13 @@ type (
...
 	layer struct {
-		neoFS       NeoFS
-		log         *zap.Logger
-		anonKey     AnonymousKey
-		resolver    BucketResolver
+		neoFS NeoFS
+		log   *zap.Logger
+		// used in case of user wants to do something like anonymous.
+		// Typical using is a flag --no-sign-request in aws-cli.
+		anonymous   user.ID
```
roman-khimov, line 47: How this...

cmd/s3-gw/app.go
```diff
@@ -86,17 +88,29 @@ type (
...
+	conns, key, poolStat := getPool(ctx, log.logger, v)
+
+	signer := user.NewAutoIDSignerRFC6979(key.PrivateKey)
+
+	// authmate doesn't require anonKey for work, but let's create random one.
+	anonKey, err := keys.NewPrivateKey()
+	if err != nil {
+		log.logger.Fatal("newApp: couldn't generate random key", zap.Error(err))
+	}
+	anonSigner := user.NewAutoIDSignerRFC6979(anonKey.PrivateKey)
```
roman-khimov, line 100: ...is related to this key? I thought they're supposed to be the same.

roman-khimov: This would make things more predictable, I think. Consider some logs, for example, you can have one key logged in one place and another in another, then try to make some sense of it.

cmd/s3-gw/app.go
```diff
@@ -189,31 +189,17 @@ func (a *App) initMetrics() {
...
-	}
+func (a *App) initResolver(ctx context.Context) {
+	endpoint := a.cfg.GetString(cfgRPCEndpoint)
 
-	order := a.cfg.GetStringSlice(cfgResolveOrder)
-	if resolveCfg.RPCAddress == "" {
-		order = remove(order, resolver.NNSResolver)
-		a.log.Warn(fmt.Sprintf("resolver '%s' won't be used since '%s' isn't provided", resolver.NNSResolver, cfgRPCEndpoint))
-	}
+	a.log.Info("rcp endpoint", zap.String("address", endpoint))
```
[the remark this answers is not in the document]
smallhive: fixed

internal/neofs/neofs.go
smallhive: Pass in constructor

internal/neofs/neofs.go
smallhive: Updated to the same solution as in HTTP gate

cmd/s3-gw/app.go
smallhive: Updated

roman-khimov (approved):

PR #809, opened 2023-08-15
Workflow to run S3 tests. Fixes #782

### LINKED ISSUE #782: GH workflow for S3 subset of neofs-testcases
While #780 is for S3-specific compatibility suite we also have a set of S3 tests in neofs-testcases which means that this subset must be run as well for this repo. Adjust tests to have this group if needed.

MaxGelbakhiani:
Merge after
https://github.com/nspcc-dev/s3-tests/pull/34
https://github.com/nspcc-dev/s3-tests/pull/33

- Workflow runs 250 sanity tests on PR and full test suite on any other event. 
- Sanity run takes less than 30 minutes. 
- For now, for Sanity I picked a set of tests, that are always green during my test runs
- Full test suite runs around 5 hours.

After merging we'll need to add action secrets:
  secrets.TEST_RESULTS_WALLET
  secrets.TEST_RESULTS_PASSWORD
And action variables:
  vars.TEST_RESULTS_NEOFS_NETWORK_DOMAIN
  vars.TEST_RESULTS_CID
  vars.TEST_RESULTS_HTTP_GATE
  
 SSL is going to be fixed on later stages

### DIFF
--- a/.github/test-env
+++ b/.github/test-env
@@ -1,0 +1,71 @@
+# Basenet settings
+LOCAL_DOMAIN=neofs.devenv
+IPV4_PREFIX=192.168.130
+CA_CERTS_TRUSTED_STORE=./vendor/certs
+
+# Bastion image
+BASTION_VERSION=10
+BASTION_IMAGE=debian
+
+# NeoGo privnet
+#CHAIN_PATH="/path/to/devenv.dump.gz"
+CHAIN_URL="https://github.com/nspcc-dev/neofs-contract/releases/download/v0.17.0/devenv_mainchain.gz"
+
+#NEOGO
+NEOGO_VERSION=0.101.1
+NEOGO_IMAGE=nspccdev/neo-go
+NEO_GO_PLATFORM=linux-amd64
+NEO_GO_URL=https://github.com/nspcc-dev/neo-go/releases/download/v${NEOGO_VERSION}/neo-go-${NEO_GO_PLATFORM}
+
+# NeoFS InnerRing nodes
+IR_VERSION=0.37.0
+IR_IMAGE=nspccdev/neofs-ir
+
+# NeoFS Storage nodes
+NODE_VERSION=0.37.0
+NODE_IMAGE=nspccdev/neofs-storage
+NODE_PLATFORM=amd64
+
+# NATS Server
+NATS_VERSION=2.7.2
+NATS_IMAGE=nats
+
+# HTTP Gate
+HTTP_GW_VERSION=0.27.0
+HTTP_GW_IMAGE=nspccdev/neofs-http-gw
+
+# REST Gate
+REST_GW_VERSION=0.5.0
+REST_GW_IMAGE=nspccdev/neofs-rest-gw
+
+# S3 Gate
+S3_GW_VERSION=_TAG_
+S3_GW_IMAGE=nspccdev/neofs-s3-gw
+S3_GW_PLATFORM=linux-amd64
+S3_GW_HOST=s3.neofs.devenv
+S3_GW_PORT=8080
+S3_GW_TLS=True
+
+# Coredns
+COREDNS_VERSION=v016
+COREDNS_IMAGE=nspccdev/coredns
+
+# NeoFS LOCODE database
+LOCODE_DB_URL=https://github.com/nspcc-dev/neofs-locode-db/releases/download/v0.3.0/locode_db.gz
+
+# NeoFS CLI binary
+NEOFS_CLI_URL=https://github.com/nspcc-dev/neofs-node/releases/download/v${NODE_VERSION}/neofs-cli-${NODE_PLATFORM}.tar.gz
+
+# Compiled NeoFS Smart Contracts
+NEOFS_CONTRACTS_VERSION=v0.17.0
+NEOFS_CONTRACTS_URL=https://github.com/nspcc-dev/neofs-contract/releases/download/${NEOFS_CONTRACTS_VERSION}/neofs-contract-${NEOFS_CONTRACTS_VERSION}.tar.gz
+
+# NeoFS adm binary
+NEOFS_ADM_URL=https://github.com/nspcc-dev/neofs-node/releases/download/v${NODE_VERSION}/neofs-adm-${NODE_PLATFORM}.tar.gz
+
+# Control service addresses used for healthchecks
+NEOFS_IR_CONTROL_GRPC_ENDPOINT=127.0.0.1:16512
+NEOFS_STORAGE_CONTROL_GRPC_ENDPOINT_1=s01.${LOCAL_DOMAIN}:8081
+NEOFS_STORAGE_CONTROL_GRPC_ENDPOINT_2=s02.${LOCAL_DOMAIN}:8081
+NEOFS_STORAGE_CONTROL_GRPC_ENDPOINT_3=s03.${LOCAL_DOMAIN}:8081
+NEOFS_STORAGE_CONTROL_GRPC_ENDPOINT_4=s04.${LOCAL_DOMAIN}:8081
--- a/.github/workflows/s3-tests.yml
+++ b/.github/workflows/s3-tests.yml
@@ -273,6 +273,30 @@ jobs:
           echo "$TEST_RESULTS_WALLET" | base64 -d > wallet.json
         working-directory: neofs-testcases
 
+      - name: Define expiration
+        if: always()
+        env:
+          TEST_RESULTS_NEOFS_NETWORK_DOMAIN: ${{ vars.TEST_RESULTS_NEOFS_NETWORK_DOMAIN }}
+          MASTER_EXPIRATION_PERIOD: ${{ vars.MASTER_EXPIRATION_PERIOD }}
+          PR_EXPIRATION_PERIOD: ${{ vars.PR_EXPIRATION_PERIOD }}
+          MANUAL_RUN_EXPIRATION_PERIOD: ${{ vars.MANUAL_RUN_EXPIRATION_PERIOD }}
+          OTHER_EXPIRATION_PERIOD: ${{ vars.OTHER_EXPIRATION_PERIOD }}
+        run: |
+          CURRENT_EPOCH=$(neofs-cli netmap epoch --rpc-endpoint st1.$TEST_RESULTS_NEOFS_NETWORK_DOMAIN:8080)
+          if [[ "${{ github.event_name }}" == "push" ]]; then
+            EXP_EPOCH=$((MASTER_EXPIRATION_PERIOD + CURRENT_EPOCH))
+          elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            EXP_EPOCH=$((PR_EXPIRATION_PERIOD + CURRENT_EPOCH))
+          elif [[ "${{ github.event_name }}" == "release" ]]; then
+            EXP_EPOCH=0 # For test reports from releases - no expiration period
+          elif [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            EXP_EPOCH=$((MANUAL_RUN_EXPIRATION_PERIOD + CURRENT_EPOCH))
+          else
+            EXP_EPOCH=$((OTHER_EXPIRATION_PERIOD + CURRENT_EPOCH))
+          fi
+          echo "EXP_EPOCH=$EXP_EPOCH" >> $GITHUB_ENV
+        working-directory: neofs-testcases
+
       - name: Put allure report to NeoFS
         id: put_report
         if: always() && steps.prepare_test_env.outcome == 'success'
@@ -282,7 +306,13 @@ jobs:
           TEST_RESULTS_CID: ${{ vars.TEST_RESULTS_CID }}
         run: |
           sudo chmod -R a+rw ${GITHUB_WORKSPACE}/allure-report
-          source ${GITHUB_WORKSPACE}/s3-tests/virtualenv/bin/activate && python3.8 ./tools/src/process-allure-reports.py --neofs_domain $TEST_RESULTS_NEOFS_NETWORK_DOMAIN --run_id $RUN_ID --cid $TEST_RESULTS_CID --allure_report ${GITHUB_WORKSPACE}/allure-report --wallet wallet.json
+          source ${GITHUB_WORKSPACE}/s3-tests/virtualenv/bin/activate
+          python ./tools/src/process-allure-reports.py \
+            --expire-at $EXP_EPOCH \
+            --neofs_domain $TEST_RESULTS_NEOFS_NETWORK_DOMAIN \
+            --run_id $RUN_ID --cid $TEST_RESULTS_CID \
+            --allure_report ${GITHUB_WORKSPACE}/allure-report \
+            --wallet wallet.json
         working-directory: neofs-testcases
 
       - name: Post the link to the report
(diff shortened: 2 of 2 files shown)


.github/test-env
```diff
@@ -0,0 +1,63 @@
...
+# REST Gate
+REST_GW_VERSION=0.5.0
+REST_GW_IMAGE=nspccdev/neofs-rest-gw
+
+# S3 Gate
+S3_GW_VERSION=_TAG_
+S3_GW_IMAGE=nspccdev/neofs-s3-gw
+S3_GW_PLATFORM=linux-amd64
+S3_GW_HOST=s3.neofs.devenv
+# S3_GW_TLS=False
```
vvarg229, line 46: Maybe delete the comment?

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+    types: [opened, synchronize]
+    paths-ignore:
+      - '**/*.md'
+  release:
+    types:
+      - published
+  workflow_dispatch:
+    inputs:
+      neofs-s3-gw_ref:
+        description: 'neofs-s3-gw ref. Default ref - latest master. Examples: v0.27.0, 8fdcc6d7e798e6511be8806b81894622e72d7fdc, branch_name'
```
vvarg229, line 20: This will not work, only the master will always run.
Yes, the neofs-node project has this bug too.

.github/sanity-tests
```diff
@@ -0,0 +1,250 @@
+s3tests_boto3/functional/test_s3.py::test_100_continue
```
vvarg229, line 1: Maybe it would be better to mark them in the test code with the @pytest.mark.sanity decorator ?

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - name: Set RUN_ID
+        env:
+          TIMESTAMP: ${{ steps.date.outputs.timestamp }}
+        run: echo "RUN_ID=${{ github.run_number }}-$TIMESTAMP" >> $GITHUB_ENV
+
+      - name: Checkout neofs-s3-gw repository
```
vvarg229, line 42:
```
      - name: Checkout neofs-node repository
        uses: actions/checkout@v3
        with:
          path: neofs-s3-gw
```

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+      - name: Get TAG for docker images
+        run: |
+          echo "CURRENT_TAG=$( make version | sed 's/^v//' )" >> $GITHUB_ENV
+        working-directory: neofs-s3-gw
+
+################################################################
+
+      - name: Build neofs-s3-gw docker image
+        run: |
+          cp .github/test-env .env #looks ugly, but I want to try
```
vvarg229, line 109: Does make require an .env file?

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+
+      - name: Prepare Dev-Env to run tests
+        id: prepare_test_env
+        timeout-minutes: 30
+        run: |
+          make prepare-test-env
+          echo "$(pwd)/vendor" >> $GITHUB_PATH
+        working-directory: neofs-dev-env
+
+      - name: Set zero fee
```
vvarg229, line 153: Please add a comment as to why this

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+        run: |
+          make update.container_fee val=0 && make update.container_alias_fee val=0
+        working-directory: neofs-dev-env
+
+      - name: Copy neofs-s3-gw binaries to vendor directory
```
vvarg229, line 158: Shouldn't this be done before `make prepare-test-env` ?

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+        working-directory: neofs-s3-gw
+
+
+#################################################################
+
+      - name: Add NeoFS S3-gw TAGs to s3-gw-test env config
+        run: |
+          sed -i -e 's/S3_GW_VERSION=.*$/S3_GW_VERSION=${{ env.CURRENT_TAG }}/' .github/test-env
+
+          echo "####### DEBUGGING INFO: ##############"
```
vvarg229, line 125: Maybe it's a good idea to remove debugging

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+        working-directory: neofs-s3-dev-env
+
+      - name: Prepare test S3-gw test env
+        run: |
+          make prepare.s3-gw
+        working-directory: neofs-s3-dev-env
+
+      - name: Copy s3tests config to s3-tests directory
+        run: |
+          cp services/s3-gw/s3tests.conf ${GITHUB_WORKSPACE}/s3-tests/${S3_TESTS_CONFIG}
```
vvarg229, line 229: But there's no such file:
image

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+        working-directory: neofs-s3-dev-env
+
+      - name: Prepare s3tests config
+        run: |
+          sed -i -e "s/host =.*$/host = ${S3_GW_IP}/" ${S3_TESTS_CONFIG}
```
vvarg229, line 234: Dangerous regular expression - maybe there's more than one string that matches the `host =.*` pattern

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+        working-directory: s3-tests
+
+      - name: Test config
+        run: |
+          cat ${S3_TESTS_CONFIG}
+        working-directory: s3-tests
+
+      - name: Setup test env
+        run: |
+          python3.8 -m venv virtualenv
```
vvarg229, line 247: Here you add the ability to run on python 3.6 too: https://github.com/nspcc-dev/s3-tests/pull/32
Why is python 3.8 the only one here?

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,318 @@
...
+          ./bootstrap
+        working-directory: s3-tests
+
+      - name: Run Sanity tests for pull requests
+        timeout-minutes: 60
+        if: github.event_name == 'pull_request'
+        run: |
+          source virtualenv/bin/activate
+          S3TEST_CONF=${S3_TESTS_CONFIG} pytest --alluredir=${GITHUB_WORKSPACE}/allure-results -v -s $(tr '\n' ' ' <${GITHUB_WORKSPACE}/neofs-s3-gw/.github/sanity-tests)
```
vvarg229, line 256: Maybe it would be better to mark them in the test code with the @pytest.mark.sanity decorator ?

vvarg229 (changes requested):

.github/sanity-tests
roman-khimov: Yeah, that'd be much easier to maintain.

.github/workflows/s3-tests.yml
roman-khimov: @vvarg229, why `neofs-node` here? We should be using the latest stable in these tests.

vvarg229: > @vvarg229, why `neofs-node` here? We should be using the latest stable in these tests.

Yeah, sure, sorry, that's a typo

.github/workflows/s3-tests.yml
vvarg229: I created a bug:
https://github.com/nspcc-dev/neofs-node/issues/2509

vvarg229 (changes requested):

.github/test-env
MaxGelbakhiani: Fixed

.github/workflows/s3-tests.yml
MaxGelbakhiani: Fixed

.github/workflows/s3-tests.yml
MaxGelbakhiani: services/s3-gw/s3tests.conf is a result file of running 

`make prepare.s3-gw `
for s3-dev-env on the previous step

image

.github/workflows/s3-tests.yml
MaxGelbakhiani: By default, sed uses only single-line matching. To enable multiline matching you need to enable it with 'N' command, like this:
`sed 'N; s/.* MULTILINE TEXT.*//' test.file`

So, greedy regex is pretty safe here. 


But, ok. I switched to a non-greedy '.*?' pattern

.github/workflows/s3-tests.yml
MaxGelbakhiani: That's outdated. I switched to 3.8

.github/workflows/s3-tests.yml
MaxGelbakhiani: As far as I see it doesn't, as I build s3-gw container from the latest code and not using the release one from dev-env.

.github/workflows/s3-tests.yml
MaxGelbakhiani: You're right, it doesn't

.github/workflows/s3-tests.yml
MaxGelbakhiani: That's a part of s3-dev-env workflow

.github/workflows/s3-tests.yml
MaxGelbakhiani: Don't get your point, but I use latest checkout for s3-gw:

`      - name: Checkout neofs-s3-gw repository
        uses: actions/checkout@v3
        with:
          repository: nspcc-dev/neofs-s3-gw
          ref: ${{ github.sha }}
          path: neofs-s3-gw`

.github/sanity-tests
MaxGelbakhiani: Fixed. Removed this test list and marked tests with @attr('sanity') in a separate commit

.github/workflows/s3-tests.yml
MaxGelbakhiani: Fixed. marked tests with @attr('sanity')

MaxGelbakhiani: https://github.com/nspcc-dev/s3-tests/pull/36

.github/sanity-tests
MaxGelbakhiani: https://github.com/nspcc-dev/s3-tests/pull/36

.github/workflows/s3-tests.yml
vvarg229: You may not use those lines in the code:
> repository: nspcc-dev/neofs-s3-gw
>           ref: ${{ github.sha }}

vvarg229 (approved):

.github/workflows/s3-tests.yml
vvarg229: It is best to replace lines like this with:
`sed -i -E "s/host =.*?$/host = ${S3_GW_IP}/" ${S3_TESTS_CONFIG}`
with the following:
`sed -i -E "s/host =__HOSTS_TO_CHANGE__/host = ${S3_GW_IP}/" ${S3_TESTS_CONFIG}`
Of course, this will require changes to the config file too.
That way you can be sure that some other line that someone else put in the config file won't change.
It will also make it easier for other users to interact with the config file.

.github/workflows/s3-tests.yml
vvarg229: > You're right, it doesn't

If not needed, then why `cp .github/test-env .env` ?

vvarg229 (changes requested):
It might be a good idea to fix regular expressions

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,314 @@
...
+
+      - name: Download latest stable neofs-cli for uploading reports to NeoFS
+        uses: dsaltares/fetch-gh-release-asset@1.1.1
+        with:
+          repo: 'nspcc-dev/neofs-node'
+          version: 'tags/v0.37.0'
+          file: 'neofs-cli-amd64'
+          target: 'neofs-node-stable/neofs-cli'
+
+      - name: Checkout neofs-testcases repository
```
roman-khimov, line 78: Why do you need it for s3-tests?

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,314 @@
...
+          echo "$(pwd)" >> $GITHUB_PATH
+        working-directory: neofs-node-stable
+
+      - name: Create wallet
+        if: always()
+        env:
+          TEST_RESULTS_WALLET: ${{ secrets.TEST_RESULTS_WALLET }}
+        run: |
+          echo "$TEST_RESULTS_WALLET" | base64 -d > wallet.json
+        working-directory: neofs-testcases
```
roman-khimov, line 287: Why this dir?

.github/workflows/s3-tests.yml
```diff
@@ -0,0 +1,314 @@
...
+        id: put_report
+        if: always() && steps.prepare_test_env.outcome == 'success'
+        env:
+          TEST_RESULTS_PASSWORD: ${{ secrets.TEST_RESULTS_PASSWORD }}
+          TEST_RESULTS_NEOFS_NETWORK_DOMAIN: ${{ vars.TEST_RESULTS_NEOFS_NETWORK_DOMAIN }}
+          TEST_RESULTS_CID: ${{ vars.TEST_RESULTS_CID }}
+        run: |
+          sudo chmod -R a+rw ${GITHUB_WORKSPACE}/allure-report
+          source ${GITHUB_WORKSPACE}/s3-tests/virtualenv/bin/activate && python3.8 ./tools/src/process-allure-reports.py --neofs_domain $TEST_RESULTS_NEOFS_NETWORK_DOMAIN --run_id $RUN_ID --cid $TEST_RESULTS_CID --allure_report ${GITHUB_WORKSPACE}/allure-report --wallet wallet.json
+        working-directory: neofs-testcases
```
roman-khimov, line 299: And here as well.

.github/workflows/s3-tests.yml
MaxGelbakhiani: Now, I removed it

.github/workflows/s3-tests.yml
MaxGelbakhiani: Go rid of this sed substitution with suitable employment of .env file for s3-dev-env.

.github/workflows/s3-tests.yml
MaxGelbakhiani: Removed it

.github/workflows/s3-tests.yml
MaxGelbakhiani: neofs-cli is used for uploading reports to NeoFS

.github/workflows/s3-tests.yml
MaxGelbakhiani: At the moment ./tools/src/process-allure-reports.py is a part of neofs-testcases. That's the reason why I have to checkout neofs-testcases and use this script to upload test-result to NeoFS

.github/workflows/s3-tests.yml
MaxGelbakhiani: The same as above

vvarg229 (dismissed):

roman-khimov:
I've just noticed that we're completely missing expirations here, they should be added

roman-khimov (approved):

PR #815, opened 2023-08-25
Fix image build

### LINKED ISSUE #814: make image is broken
```
Step 7/13 : RUN make
 ---> Running in 7b7559bcde4f
tree service revision feaa9eace7098c343598bf08fb50746a1e8d2deb
./syncTree.sh: line 9: jq: command not found
```
Caused by cad239ee25523aef17e8577a9687c6dc36f34b49.

smallhive:
closes #814 

Changing to _Alpine_ version is required to install `jq` package. But this alpine image doesn't contain some important packages, thus installed them too.

Using jq also required because it formats JSON from Github request and we _grep_ results from it

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20 as builder
+FROM golang:1.20-alpine as builder
 
 ARG BUILD=now
 ARG REPO=github.com/nspcc-dev/neofs-s3-gw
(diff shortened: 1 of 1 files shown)


roman-khimov (approved):

roman-khimov:
Why a `merge` commit here?

smallhive:
> Why a `merge` commit here?

Good question, fixed

PR #825, opened 2023-09-14
Don't add eACL rules to System group

### LINKED ISSUE #817: Do not add eACL rules related to SYSTEM group
* recently https://github.com/nspcc-dev/neofs-sdk-go/releases/tag/v1.0.0-rc.11 was done
* in particular, https://github.com/nspcc-dev/neofs-sdk-go/pull/515 change was introduced.

S3 gateway adds SYSTEM-targeted records to the eACLs:
https://github.com/nspcc-dev/neofs-s3-gw/blob/d6fba1d2e9b31ae29e9e4ff6375b78669418731f/api/handler/acl.go#L157-L165

now it became useless, so this behavior should be fixed/reconsidered

smallhive:
closes #817

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -154,16 +154,6 @@ type ServiceRecord struct {
 	GroupRecordsLength int
 }
 
-func (s ServiceRecord) ToEACLRecord() *eacl.Record {
-	serviceRecord := eacl.NewRecord()
-	serviceRecord.SetAction(eacl.ActionAllow)
-	serviceRecord.SetOperation(eacl.OperationGet)
-	serviceRecord.AddFilter(eacl.HeaderFromService, eacl.MatchUnknown, serviceRecordResourceKey, s.Resource)
-	serviceRecord.AddFilter(eacl.HeaderFromService, eacl.MatchUnknown, serviceRecordGroupLengthKey, strconv.Itoa(s.GroupRecordsLength))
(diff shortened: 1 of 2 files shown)


api/handler/acl.go
```diff
@@ -154,16 +154,6 @@ type ServiceRecord struct {
 	GroupRecordsLength int
 }
 
-func (s ServiceRecord) ToEACLRecord() *eacl.Record {
```
roman-khimov, line 157: Do you need `ServiceRecord` type after it?

smallhive: It is still in use [here](https://github.com/nspcc-dev/neofs-s3-gw/blob/bb77e306f43683d3049d7fc8f319d1e9b7249f0a/api/handler/acl.go#L868). On the one hand we should remove all system role, on the other hand - what if some eACL table contains it? Will this code broke something if we remove this usage?

smallhive: Removing this code doesn't break the tests, but I'm not sure, what do you think. @roman-khimov

roman-khimov: Hmm. S3 is not in production, so we're not likely to get issues missing this code (there no S3-based EACLs that have it). At the same time it's still a part of the protocol and we can keep it (just in case) until it goes out.

roman-khimov (approved):

PR #831, opened 2023-09-25
Add slicer as an option

### LINKED ISSUE #808: Add an option to slice objects in the gateway itself
Traditionally we use NeoFS API as is and object slicing is performed on the SN end (with appropriate session tokens created and used). We can try slicing in the gateway itself now and see if it works any better or solves any problems. It should be an experimental option for now.

smallhive:
close #808

### DIFF
--- a/cmd/s3-authmate/main.go
+++ b/cmd/s3-authmate/main.go
@@ -726,7 +726,11 @@ func createNeoFS(ctx context.Context, log *zap.Logger, cfg PoolConfig, anonSigne
 		return nil, fmt.Errorf("networkInfo: %w", err)
 	}
 
-	neoFS := neofs.NewNeoFS(p, signer, anonSigner, int64(ni.MaxObjectSize()))
+	neofsCfg := neofs.Config{
+		MaxObjectSize: int64(ni.MaxObjectSize()),
+	}
+
+	neoFS := neofs.NewNeoFS(p, signer, anonSigner, neofsCfg)
 
 	return neofs.NewAuthmateNeoFS(neoFS), nil
 }
--- a/cmd/s3-authmate/main.go
+++ b/cmd/s3-authmate/main.go
@@ -79,6 +79,7 @@ var (
 	containerPolicies        string
 	awcCliCredFile           string
 	timeoutFlag              time.Duration
+	slicerEnabledFlag        bool
 
 	// pool timeouts flag.
 	poolDialTimeoutFlag        time.Duration
@@ -293,6 +294,11 @@ It will be ceil rounded to the nearest amount of epoch.`,
 				Destination: &poolStreamTimeoutFlag,
 				Value:       poolStreamTimeout,
 			},
+			&cli.BoolFlag{
+				Name:        "internal-slicer",
+				Usage:       "Enable slicer for object uploading",
+				Destination: &slicerEnabledFlag,
+			},
 		},
 		Action: func(c *cli.Context) error {
 			ctx, log := prepare()
@@ -322,7 +328,7 @@ It will be ceil rounded to the nearest amount of epoch.`,
 			}
 			anonSigner := user.NewAutoIDSignerRFC6979(anonKey.PrivateKey)
 
-			neoFS, err := createNeoFS(ctx, log, poolCfg, anonSigner)
+			neoFS, err := createNeoFS(ctx, log, poolCfg, anonSigner, slicerEnabledFlag)
 			if err != nil {
 				return cli.Exit(fmt.Sprintf("failed to create NeoFS component: %s", err), 2)
 			}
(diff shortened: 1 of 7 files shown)


cmd/s3-authmate/main.go
```diff
@@ -293,6 +294,11 @@ It will be ceil rounded to the nearest amount of epoch.`,
 				Destination: &poolStreamTimeoutFlag,
 				Value:       poolStreamTimeout,
 			},
+			&cli.BoolFlag{
+				Name:        "with-slicer",
```
roman-khimov, line 298: `internal-slicer`?

config/config.env
```diff
@@ -119,6 +119,12 @@ S3_GW_CORS_DEFAULT_MAX_AGE=600
...
 
+# Timeout between retrieving actual epoch from NeoFS. Actual only if slicer.enabled = true.
+S3_GW_NEOFS_EPOCH_UPDATE_INTERVAL=30s
+
 # List of allowed AccessKeyID prefixes
 # If not set, S3 GW will accept all AccessKeyIDs
 S3_GW_ALLOWED_ACCESS_KEY_ID_PREFIXES=Ck9BHsgKcnwfCTUSFm6pxhoNS4cBqgN2NQ8zVgPjqZDX 3stjWenX15YwYzczMr88gy3CQr4NYFBQ8P7keGzH5QFn
+
+# Allows to use slicer for Object uploading.
+S3_GW_SLICER_ENABLED=false
```
roman-khimov, line 130: Now newline? Can't be!

cmd/s3-gw/app_settings.go
```diff
@@ -121,11 +122,17 @@ const ( // Settings.
...
+	cfgEpochUpdateInterval = "neofs.epoch_update_interval"
+
 	// List of allowed AccessKeyID prefixes.
 	cfgAllowedAccessKeyIDPrefixes = "allowed_access_key_id_prefixes"
 
 	// envPrefix is an environment variables prefix used for configuration.
 	envPrefix = "S3_GW"
+
+	// Shows if slicer is enabled. If enabled slicer will be used for object put.
+	cfgSlicerEnabled = "slicer.enabled"
```
roman-khimov, line 135: `internal_slicer`? I'm not sure what would be the best name for it.

config/config.env
```diff
@@ -119,6 +119,12 @@ S3_GW_CORS_DEFAULT_MAX_AGE=600
 # If not set, default value 0 will be used -- it means that object will be processed according to the container's placement policy
 S3_GW_NEOFS_SET_COPIES_NUMBER=0
 
+# Timeout between retrieving actual epoch from NeoFS. Actual only if slicer.enabled = true.
+S3_GW_NEOFS_EPOCH_UPDATE_INTERVAL=30s
```
roman-khimov, line 123: Seems to be a bit aggressive.

cmd/s3-gw/app.go
```diff
@@ -107,10 +107,25 @@ func newApp(ctx context.Context, log *Logger, v *viper.Viper) *App {
...
 
-	neoFS := neofs.NewNeoFS(conns, signer, anonSigner, neofsCfg)
+	// If slicer is disabled, we should use "static" getter, which doesn't make periodic requests to the NeoFS.
+	var epochGetter neofs.EpochGetter = ni
+
+	if neofsCfg.IsSlicerEnabled {
+		epochUpdateInterval := v.GetDuration(cfgEpochUpdateInterval)
+
+		if epochUpdateInterval == 0 {
+			epochUpdateInterval = defaultEpochUpdateInterval
```
roman-khimov, line 122: Likely we can calculate a half of blocks per epoch times time per block for the default, all of this data is available from the network. And it'd be the best one.

internal/neofs/epoch.go
```diff
@@ -0,0 +1,71 @@
...
+// CurrentEpoch implements [EpochGetter].
+func (g *PeriodicGetter) CurrentEpoch() uint64 {
+	return g.epoch.Load()
+}
+
+func (g *PeriodicGetter) update(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
roman-khimov, line 59: You don't (and in fact can't right now) stop the timer in this case.

cmd/s3-gw/app.go
smallhive: Replaced

config/config.env
smallhive: Replaced with 2m. The value I got after epoch_duration and ms_per_block math.

config/config.env
smallhive: Updated

cmd/s3-gw/app_settings.go
smallhive: I'm not sure, that I like _internal_ in this case

internal/neofs/epoch.go
smallhive: Updated

cmd/s3-gw/app_settings.go
roman-khimov: Internal for the S3 gateway. Otherwise it happens in the NeoFS node, thus can be considered external to the S3 gateway. Maybe there are better options.

roman-khimov (approved):

PR #835, opened 2023-10-04
Preallocate buffers to object uploads

### LINKED ISSUE #833: Optimize allocations
I'm always frustrated when I'm looking at gw memory profiles, it makes too many allocations for the job it does.

Allocate less, reuse.

smallhive:
closes #833


internal/neofs/neofs.go
```diff
@@ -46,16 +47,23 @@ type NeoFS struct {
 	anonSigner  user.Signer
 	cfg         Config
 	epochGetter EpochGetter
+	buffers     chan []byte
 }
 
 // NewNeoFS creates new NeoFS using provided pool.Pool.
 func NewNeoFS(p *pool.Pool, signer user.Signer, anonSigner user.Signer, cfg Config, epochGetter EpochGetter) *NeoFS {
+	buffers := make(chan []byte, cfg.UploadChunkPoolLength)
```
roman-khimov, line 55: https://pkg.go.dev/sync#Pool ?

smallhive: [This option](https://github.com/nspcc-dev/neofs-s3-gw/pull/835/commits/ce7d2d84be14c655882371851085b27e03447117) better then original by memory, but lose by CPU.
I have to adopt it, in the code, it looks better, but the numbers are not so impressive.

The first variant with the channels has one possible bottleneck - reading from the channel. Concurrence is limited by channel capacity. We may try to allocate a new slice if the channel is locked. Unfortunately, it is a new allocation...

The second option is trying to experiment with channel sizes with different bench loads.

internal/neofs/neofs.go
```diff
@@ -46,16 +48,29 @@ type NeoFS struct {
...
 func NewNeoFS(p *pool.Pool, signer user.Signer, anonSigner user.Signer, cfg Config, epochGetter EpochGetter) *NeoFS {
+	buffers := sync.Pool{}
+	buffers.New = func() any {
+		b := make([]byte, cfg.MaxObjectSize)
+		return &b
+	}
+
+	for i := 0; i < cfg.UploadChunkPoolLength; i++ {
+		b := make([]byte, cfg.MaxObjectSize)
+		buffers.Put(&b)
```
roman-khimov, line 64: Not really needed (along with `UploadChunkPoolLength`), it'll be allocated and released dynamically.

internal/neofs/neofs.go
```diff
@@ -316,8 +331,14 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
 		return oid.ID{}, fmt.Errorf("save object via connection pool: %w", err)
 	}
 
-	chunk := make([]byte, x.cfg.MaxObjectSize)
-	_, err = io.CopyBuffer(writer, prm.Payload, chunk)
+	data := x.buffers.Get()
+	chunk, _ := data.(*[]byte)
```
roman-khimov, line 335: Omit `, _`?

internal/neofs/neofs.go
```diff
@@ -316,8 +331,14 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
...
+
+	defer func() {
```
roman-khimov, line 337: Can be a bit more efficient if you're to do it after `CopyBuffer`, you don't need to wait for `Close()` to return the buffer to the pool.

roman-khimov (commented):
Can slicer reuse buffers as well?

Pool-based code is simpler in general.

internal/neofs/neofs.go
smallhive: After some benchmarks, `sync.Pool` won the `channel` version. Updated

smallhive:
> Can slicer reuse buffers as well?
> 
> Pool-based code is simpler in general.

Supported

internal/neofs/neofs.go
```diff
@@ -316,8 +325,12 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
 		return oid.ID{}, fmt.Errorf("save object via connection pool: %w", err)
 	}
 
-	chunk := make([]byte, x.cfg.MaxObjectSize)
-	_, err = io.CopyBuffer(writer, prm.Payload, chunk)
+	data := x.buffers.Get()
+	chunk := data.(*[]byte)
```
roman-khimov, line 329: Likely some Go 1.2X will bring generic-based `Pool` to us.

roman-khimov (approved):

internal/neofs/neofs.go
smallhive: It would be great, but 1.21 still have `any` version

PR #837, opened 2023-10-06
.github: update s3 tests run step

evgeniiz321:
In order to support https://github.com/nspcc-dev/s3-tests/pull/43. There is just one set of tests, it runs for 8min30sec, there is no point of having two different sets of tests right now.

### DIFF
--- a/.github/workflows/s3-tests.yml
+++ b/.github/workflows/s3-tests.yml
@@ -235,19 +235,10 @@ jobs:
           ./bootstrap
         working-directory: s3-tests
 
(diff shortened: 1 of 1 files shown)


vvarg229:
image
How is it that the tests were cut by 1 hour?

roman-khimov (approved):

evgeniiz321:
Failing tests were the reason of 1 hour run. Without them - it is much faster.

vvarg229 (approved):

PR #839, opened 2023-10-09
Sdk optimizations

smallhive:


internal/neofs/neofs.go
```diff
@@ -293,6 +293,10 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
 		opts.SetCopiesNumber(prm.CopiesNumber)
 		opts.SetCurrentNeoFSEpoch(x.epochGetter.CurrentEpoch())
 
+		data := x.buffers.Get()
+		chunk := data.(*[]byte)
+		opts.SetPayloadBuffer(*chunk)
```
roman-khimov, line 298: You can move it below the bearer token handling to minimize buffer use window (homomorphic hash can take some time to be calculated).

smallhive: `CalculateHomomorphicChecksum` is just not a very good function name. [This is a set bool flag](https://github.com/nspcc-dev/neofs-sdk-go/blob/0a4e7df349e847bfb0f054513a67d89fceda1563/object/slicer/options.go#L37).

roman-khimov (approved):

PR #848, opened 2023-10-12
PutObject metadata with empty string

### LINKED ISSUE #846: PutObject: handle metadata with an empty string as a value more gracefully
If we put an object with metadata value equal to empty string "", we got 500 from the s3 gate:

```
2023-10-12T00:44:17.218Z	error	handler/util.go:29	call method	{"status": 500, "request_id": "3bb78287-09cd-4b0d-901b-fe341066115c", "method": "PutObject", "bucket": "yournamehere-xxaxwdqfj5jh3mk8-1", "object": "foo", "description": "could not upload object", "error": "save object via connection pool: init writing on API client: client failure: status: code = 1024 message = empty attribute value", "body close errors": []}
```

Corresponding code: 
```
metadata_dict = {"meta1": ""}
client.put_object(Bucket=bucket_name, Key="foo", Body="bar", Metadata=metadata_dict)
```

Expected behavior is to create an object and then return this empty string "" upon get requests inside Metadata.

smallhive:
Closes #846

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This document outlines major changes between releases.
 
 ## [Unreleased]
 
+### Added
+- Passing an empty meta parameter value raises "Your metadata headers are not supported." error.
+
 ## [0.29.0] - 2023-09-28
 
 ### Added
--- a/api/handler/util.go
+++ b/api/handler/util.go
@@ -39,6 +39,10 @@ func transformToS3Error(err error) error {
 		return s3errors.GetAPIError(s3errors.ErrAccessDenied)
 	}
 
+	if errorsStd.Is(err, layer.ErrMetaEmptyParameterValue) {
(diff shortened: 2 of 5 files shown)


roman-khimov (commented):
This must be documented.

smallhive:
I've added comments to the code and the README file

README.md
```diff
@@ -54,6 +54,11 @@ $ S3_GW_PEERS_0_ADDRESS=grpcs://192.168.130.72:8080 \
   neofs-s3-gw
 ```
 
+## Restrictions
```
roman-khimov, line 57: We better have it in `docs/aws_s3_compat.md`, another section on metadata (like we have for ACLs) would be appropriate here.

smallhive: Make sense, missed this doc location. Updated

roman-khimov (approved):

PR #871, opened 2023-10-20
Extend docs about Authenticated Users group

### LINKED ISSUE #858: "authenticated-read" canned ACL is not supported
Due to https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html (Canned ACL section)

> authenticated-read - Owner gets FULL_CONTROL. The AuthenticatedUsers group gets READ access.

Currently we get for bucket acls - 

```
{'Grantee': 
{'DisplayName': '', 'EmailAddress': '', 'ID': '', 'Type': 'Group', 'URI': 'http://acs.amazonaws.com/groups/global/AllUsers'}, 'Permission': 'FULL_CONTROL'}, 
{'Grantee': {'DisplayName': '', 'EmailAddress': '', 'ID': '031a6c6fbbdf02ca351745fa86b9ba5a9452d785ac4f7fc2b7548ca2a46c4fcf4a', 'Type': 'CanonicalUser', 'URI': ''}, 'Permission': 'FULL_CONTROL'}
```

URI is not correct (it is http://acs.amazonaws.com/groups/global/AllUsers, should be http://acs.amazonaws.com/groups/global/AuthenticatedUsers) and Permission should be READ instead of FULL_CONTROL

Same applied for object ACLs as well.

smallhive:
We explicitly mention GW doesn't support this group. Because in NeoFS only two groups: owner and others. Right now there is no possible way to determine who is who.

Closes #858.

### DIFF
--- a/docs/aws_s3_compat.md
+++ b/docs/aws_s3_compat.md
@@ -51,7 +51,8 @@ Principal must be `"AWS": "*"` (to refer all users) or `"CanonicalUser": "0313b1
 }
 ```
 * AWS conditions and wildcard are not supported in [resources](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html)
-* Only `CanonicalUser` (with hex encoded public key) and `All Users Group` are supported in [ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html)
+* Only `CanonicalUser` (with hex encoded public key) and `All Users Group` are supported in [ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html).
+`Authenticated Users group` is not supported. It is a part of `All Users Group` and can't be separated from it.
 
 |    | Method       | Comments        |
 |----|--------------|-----------------|


docs/aws_s3_compat.md
```diff
@@ -51,7 +51,8 @@ Principal must be `"AWS": "*"` (to refer all users) or `"CanonicalUser": "0313b1
 }
 ```
 * AWS conditions and wildcard are not supported in [resources](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html)
-* Only `CanonicalUser` (with hex encoded public key) and `All Users Group` are supported in [ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html)
+* Only `CanonicalUser` (with hex encoded public key) and `All Users Group` are supported in [ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html).
+`Authenticated Users group` is not supported. It is a part of `All Users Group` and can't be divided from it.
```
roman-khimov, line 55: s/divided/separated/

smallhive: Replaced

roman-khimov (approved):

PR #893, opened 2023-11-02
Check x-amz-expires header on maximum border

### LINKED ISSUE #853: X-Amz-Expires upper limit should be limited to 7 days
According to https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html

If you create a presigned URL with the Amazon S3 console, the expiration time can be set between 1 minute and 12 hours. If you use the AWS CLI or AWS SDKs, the expiration time can be set as high as 7 days.

Currently there is no checks for upper limit and we can specify as high as we want.

smallhive:
Fixing authmate wasn't enough.

Closes #853.

### DIFF
--- a/api/auth/center.go
+++ b/api/auth/center.go
@@ -21,6 +21,7 @@ import (
 	"github.com/nspcc-dev/neofs-s3-gw/api/s3errors"
 	"github.com/nspcc-dev/neofs-s3-gw/creds/accessbox"
 	"github.com/nspcc-dev/neofs-s3-gw/creds/tokens"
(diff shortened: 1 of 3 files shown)


internal/models/const.go
```diff
@@ -0,0 +1,11 @@
+package models
```
roman-khimov, line 1: `limits`?

smallhive: Don't mind, renamed

roman-khimov (approved):

PR #924, opened 2024-02-08
Return error if no tags in bucket

### LINKED ISSUE #898: GetBucketTagging: gw should return 404 when there are no tags
test_set_bucket_tagging

Execute GetBucketTagging on a freshly created bucket without no tags - got 200:
```
2023-11-08T02:09:52.314Z	info	api/router.go:165	call method	{"status": 200, "host": "s3.neofs.devenv:8080", "request_id": "011b1bf8-3c07-4d15-bb70-e704d18f45c1", "method": "CreateBucket", "bucket": "yournamehere-gtgr2e70qm9a58wz-1", "object": "", "description": "OK"}
2023-11-08T02:09:52.343Z	info	api/router.go:165	call method	{"status": 200, "host": "s3.neofs.devenv:8080", "request_id": "ac630ca4-b054-4983-90a1-a0c04cffa2c3", "method": "GetBucketTagging", "bucket": "yournamehere-gtgr2e70qm9a58wz-1", "object": "", "description": "OK"}
```
It is expected to get 404 NoSuchTagSet

smallhive:
Closes #898.

### DIFF
--- a/api/layer/tagging.go
+++ b/api/layer/tagging.go
@@ -116,8 +116,16 @@ func (n *layer) GetBucketTagging(ctx context.Context, bktInfo *data.BucketInfo) 
 	}
 
 	tags, err := n.treeService.GetBucketTagging(ctx, bktInfo)
-	if err != nil && !errorsStd.Is(err, ErrNodeNotFound) {
+	if err != nil {
+		if errorsStd.Is(err, ErrNodeNotFound) {
+			return nil, s3errors.GetAPIError(s3errors.ErrBucketTaggingNotFound)
+		}
+
 		return nil, err
+	}
+
+	if len(tags) == 0 {
+		return nil, s3errors.GetAPIError(s3errors.ErrBucketTaggingNotFound)
 	}
 
 	n.cache.PutTagging(owner, bucketTaggingCacheKey(bktInfo.CID), tags)


api/layer/tagging.go
```diff
@@ -116,10 +116,18 @@ func (n *layer) GetBucketTagging(ctx context.Context, bktInfo *data.BucketInfo)
...
-	if err != nil && !errorsStd.Is(err, ErrNodeNotFound) {
+	if err != nil {
+		if errorsStd.Is(err, ErrNodeNotFound) {
+			return nil, s3errors.GetAPIError(s3errors.ErrBucketTaggingNotFound)
+		}
+
 		return nil, err
 	}
 
+	if len(tags) == 0 {
```
roman-khimov, line 127: I think you can move it before the `err` check and then save on `ErrNodeNotFound` check (`tags` will be `nil` with 0 length anyway).

smallhive: The idea doesn't save the empty/nil value to the cache. If we have `ErrNodeNotFound`, then `tags` is nil/empty, we don't need to try to save it to the cache. But simultaneously we should respond to the client with a correct error.

In this case, it is better to return the previous implementation and append-only one condition after error checks
```go
if len(tags) == 0 {
	return nil, s3errors.GetAPIError(s3errors.ErrBucketTaggingNotFound)
}
```

roman-khimov (approved):

PR #933, opened 2024-02-28
Update to go 1.20 min

smallhive:


roman-khimov (commented):
`internal/neofs/tree.go` can be simplified in `strings.Trim*` functions, although `CutPrefix` is not exactly perfect here, it's mostly that some checks are not needed.

`api/layer/encryption/encryption.go` can be improved to use `bytes.Clone` instead of open-coded copying.

.github/workflows/tests.yml
```diff
@@ -18,10 +18,11 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           cache: true
           go-version: '1.22'
```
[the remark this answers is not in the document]
smallhive: Fixed

roman-khimov:
https://github.com/nspcc-dev/neofs-s3-gw/pull/933#pullrequestreview-1906053457

smallhive:
`encryption.go` updated. `CutPrefix` adds the second returning parameter, it is mostly unconvinient

api/layer/encryption/encryption.go
```diff
@@ -71,7 +71,7 @@ func NewParams(key []byte) (*Params, error) {
 	}
 	var p Params
 	p.customerKey = make([]byte, aes256KeySize)
```
roman-khimov, line 73: So why do we need this then?

smallhive: Obviously not required, updated

roman-khimov (approved):

PR #940, opened 2024-03-26
Do not use `github.com/nspcc-dev/neofs-api-go/v2 module` directly

cthulhu-rider:
it's going to be deprecated. Any app should use https://github.com/nspcc-dev/neofs-sdk-go only

* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/570

PR #939, opened 2024-03-26
CHANGELOG: release 0.30.0

smallhive:

PR #941, opened 2024-03-26
Makefile: Fix `sync-tree`

cthulhu-rider:
i tried to fix for #940, and saw a request for a fix within #939 late. We can handle this before or just cherry-pick there, up to you @roman-khimov @smallhive 

in the future, i'd try to switch to using some stable GitHub API ([trees](https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28) or [contents](https://docs.github.com/en/rest/repos/contents))

### DIFF
--- a/syncTree.sh
+++ b/syncTree.sh
@@ -6,7 +6,7 @@ REVISION="feaa9eace7098c343598bf08fb50746a1e8d2deb"
 
 echo "tree service revision ${REVISION}"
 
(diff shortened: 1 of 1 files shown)


roman-khimov (approved):
GH has nice RESTful APIs, but they all require access tokens (although maybe GITHUB_TOKEN or whatever it's called can work this way). I'd rather use git submodules here, frankly, but all of these constructions are obviously temporary, so if this fixes the build for now -- let's do it.

PR #945, opened 2024-04-02
.github: update system tests flow for new dev env

evgeniiz321:


.github/workflows/system-tests.yml
```diff
@@ -96,80 +96,71 @@ jobs:
...
+          sudo chmod a+x neofs-adm
+          sudo chmod a+x neofs-ir
+          sudo chmod a+x neofs-lens
+          sudo chmod a+x neofs-node
+          sudo chmod a+x neofs-rest-gw
+          sudo chmod a+x neo-go
+          sudo chmod a+x neofs-s3-authmate
+          sudo chmod a+x neofs-s3-gw
+        working-directory: neofs-testcases
+
```
smallhive, line 157: Maybe, it makes sense to `chmod` binaries before copying? I mean afters rows 137, 142 make `chmod -r a+x ./bin`
and then copy updated binaries?
I don't insist, just an another option

smallhive (dismissed):

.github/workflows/system-tests.yml
```diff
@@ -46,17 +46,17 @@ jobs:
...
           path: neofs-testcases
 
-      - name: Checkout neofs-s3-gw repository
+      - name: Checkout neofs-node repository
         uses: actions/checkout@v4
         with:
-          path: neofs-s3-gw
+          repository: nspcc-dev/neofs-node
+          ref: master
+          path: neofs-node
```
roman-khimov, line 54: Wow. Why `master` node? It isn't supposed to be this way, we were using _stable_ node, but `master` gateway here and that's the combination we want. We don't want tests to fail because of broken node, we want them to fail because of broken gateway.

.github/workflows/system-tests.yml
```diff
@@ -96,80 +96,71 @@ jobs:
...
-          sed -i -e 's/S3_GW_VERSION=_TAG_/S3_GW_VERSION=${{ env.CURRENT_TAG }}/' .github/testcases-env
-        working-directory: neofs-s3-gw
-
-      - name: Copy testcases-env file to .env for neofs-dev-env
-        run: |
-          cp .github/testcases-env ${GITHUB_WORKSPACE}/neofs-dev-env/.env
-        working-directory: neofs-s3-gw
-
-################################################################
+      - name: Download latest stable neofs-rest-gw
```
roman-khimov, line 110: Do we need to download REST at all for these tests?

.github/workflows/system-tests.yml
```diff
@@ -96,80 +96,71 @@ jobs:
...
-        run: |
-          make get
-          sudo ./bin/update_hosts.sh
-          sudo chmod a+w vendor/hosts
-        working-directory: neofs-dev-env
+      - name: Download latest stable neo-go
+        uses: dsaltares/fetch-gh-release-asset@1.1.1
+        with:
+          repo: 'nspcc-dev/neo-go'
+          version: 'tags/v0.104.0'
```
roman-khimov, line 122: Outdated.

.github/workflows/system-tests.yml
evgeniiz321: Tests can't start on the current latest 'stable' node (v0.40.1). They work fine on the current master. Can we make a temporary release maybe? or I can hardcode a hash of current latest commit.

roman-khimov: OK, then let's temporarily use a stable commit.

.github/workflows/system-tests.yml
evgeniiz321: The problem is that binaries come from different places, so it is better to `chmod` them in one place, just to be sure nothing is left.

.github/workflows/system-tests.yml
evgeniiz321: Removed

.github/workflows/system-tests.yml
evgeniiz321: Done

.github/workflows/system-tests.yml
evgeniiz321: Updated to the latest release version

roman-khimov (approved):

PR #950, opened 2024-05-12
.github: update s3tests workflow to use new dev env

evgeniiz321:


.github/workflows/s3-tests.yml
```diff
@@ -84,90 +106,50 @@ jobs:
       - run: go version
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v4
```
smallhive, line 109: Is v4 more suitable?

evgeniiz321: new is always better :) it was a copypaste from somewhere

.github/workflows/s3-tests.yml
```diff
@@ -42,38 +42,60 @@ jobs:
...
+          repo: 'nspcc-dev/neofs-node'
+          version: 'tags/v0.41.1'
+          file: 'neofs-node-linux-amd64'
+          target: 's3-tests/neofs-node'
+
+      - name: Download latest stable neo-go
+        uses: dsaltares/fetch-gh-release-asset@1.1.1
+        with:
+          repo: 'nspcc-dev/neo-go'
+          version: 'tags/v0.104.0'
```
roman-khimov, line 96: v0.105.1 at least.

evgeniiz321: Updated.

roman-khimov (approved):

PR #957, opened 2024-05-28
Switch to v2 object split scheme

smallhive:


internal/neofs/tree.go
```diff
@@ -1320,7 +1334,6 @@ func metaFromMultipart(info *data.MultipartInfo, fileName string) map[string]str
 	info.Meta[uploadIDKV] = info.UploadID
```
roman-khimov, line 1334: Can the first object ID be used for UploadID?

api/layer/tree_service.go
```diff
@@ -74,11 +74,10 @@ type TreeService interface {
 	// If object id to remove is not found returns ErrNoNodeToRemove error.
 	AddPart(ctx context.Context, bktInfo *data.BucketInfo, multipartNodeID uint64, info *data.PartInfo) (oldObjIDToDelete oid.ID, err error)
 	GetParts(ctx context.Context, bktInfo *data.BucketInfo, multipartNodeID uint64) ([]*data.PartInfo, error)
-	// GetLastPart returns the latest uploaded part.
-	//
+	// GetPartByNumber returns the part by number. If part was uploaded few times the newest one will be returned.
 	// Return errors:
 	//   - [ErrPartListIsEmpty] if there is no parts in the upload id.
-	GetLastPart(ctx context.Context, bktInfo *data.BucketInfo, multipartNodeID uint64) (*data.PartInfo, error)
+	GetPartByNumber(ctx context.Context, bktInfo *data.BucketInfo, multipartNodeID uint64, number int) (*data.PartInfo, error)
```
roman-khimov, line 80: Why this change? I don't see you requesting anything but the last part.

api/layer/neofs_mock.go
```diff
@@ -270,19 +270,13 @@ func (t *TestNeoFS) CreateObject(_ context.Context, prm PrmObjectCreate) (oid.ID
...
 	t.currentEpoch++
 
-	if prm.Multipart != nil && prm.Multipart.SplitID != "" {
-		var split object.SplitID
-		if err := split.Parse(prm.Multipart.SplitID); err != nil {
-			return oid.ID{}, fmt.Errorf("split parse: %w", err)
-		}
-		obj.SetSplitID(&split)
-
+	if prm.Multipart != nil {
```
carpawell, line 273: so the whole `if` works right only because we pray it is called right? i mean a link object should never have a previous one sert, a regular object's payload should never be rewritten, etc

api/layer/tree_mock.go
```diff
@@ -390,7 +390,18 @@ func (t *TreeServiceMock) GetLastPart(ctx context.Context, bktInfo *data.BucketI
 		return 1
 	})
 
-	return parts[len(parts)-1], nil
+	var pi *data.PartInfo
+	for _, part := range parts {
```
carpawell, line 394: aren't parts sorted?

api/layer/tree_service.go
carpawell: but how is it related to V1 -> V2 move?

api/layer/neofs_mock.go
```diff
@@ -293,6 +293,39 @@ func (t *TestNeoFS) CreateObject(_ context.Context, prm PrmObjectCreate) (oid.ID
...
+					payload = append(payload, partialObject.Payload()...)
+				}
+			}
+
+			pid, isSet := prm.Multipart.HeaderObject.ID()
+			if !isSet {
+				return oid.ID{}, errors.New("HeaderObject id is not set")
+			}
+
+			realHeaderObj := object.New()
```
carpawell, line 314: can you, please, explain why we do think in tests in `CreateObject`?

api/layer/tree_service.go
carpawell: it was expected to simplify the process. can you, please, describe a little why it is required to _now_ and was not needed _then_?

api/layer/tree_mock.go
carpawell: why not try to get part directly via slice index? is there any cases when N-th part is not N-th in slice?

cmd/s3-gw/app_settings.go
```diff
@@ -413,6 +413,8 @@ func newLogger(v *viper.Viper) *Logger {
 		panic(fmt.Sprintf("build zap logger instance: %v", err))
 	}
 
+	zap.ReplaceGlobals(l)
```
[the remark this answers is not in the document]
carpawell: i would say globals are not good in general, also, i think we do not use it in any of our projects

api/layer/tree_mock.go
smallhive: There may be more than one part with the same index. They are sorted by `ServerCreated` field. We should find the latest/actual one.

roman-khimov: But why? Even when you reupload, you invalidate the old one, but you still have only one proper current one.

api/handler/multipart_upload.go
```diff
@@ -648,12 +649,12 @@ func encodeListMultipartUploadsToResponse(info *layer.ListMultipartUploadsInfo,
 		m := MultipartUpload{
 			Initiated: u.Created.UTC().Format(time.RFC3339),
 			Initiator: Initiator{
-				ID:          u.Owner.String(),
+				ID:          hex.EncodeToString(u.OwnerPubKey.Bytes()),
```
roman-khimov, line 652: StringCompressed()

api/handler/multipart_upload.go
```diff
@@ -648,12 +649,12 @@ func encodeListMultipartUploadsToResponse(info *layer.ListMultipartUploadsInfo,
...
 				DisplayName: u.Owner.String(),
 			},
 			Key: u.Key,
 			Owner: Owner{
-				ID:          u.Owner.String(),
+				ID:          hex.EncodeToString(u.OwnerPubKey.Bytes()),
```
roman-khimov, line 657: And everywhere below

internal/neofs/tree.go
```diff
@@ -226,6 +228,15 @@ func newMultipartInfo(node NodeResponse) (*data.MultipartInfo, error) {
 			}
 		case ownerKV:
 			_ = multipartInfo.Owner.DecodeString(string(kv.GetValue()))
+		case ownerPubKeyKV:
+			payload, err := hex.DecodeString(string(kv.GetValue()))
+			if err != nil {
+				return nil, fmt.Errorf("get pub key: %w", err)
+			}
+
+			if err = multipartInfo.OwnerPubKey.DecodeBytes(payload); err != nil {
```
roman-khimov, line 237: https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/crypto/keys#NewPublicKeyFromString

internal/neofs/tree.go
roman-khimov: So you can upload the first object (even with 0 payload!) and use its hash, right?

api/handler/object_list.go
```diff
@@ -199,6 +200,15 @@ func fillContents(src []*data.ObjectInfo, encode string, fetchOwner bool) []Obje
 			ETag:         obj.HashSum,
 		}
 
+		if size, ok := obj.Headers[layer.AttributeDecryptedSize]; ok {
+			sz, err := strconv.ParseInt(size, 10, 64)
+			if err != nil {
+				zap.L().Warn("Failed to parse size from header", zap.String("header_value", size), zap.Error(err))
```
roman-khimov, line 206: It's an error. It should be returned in some way.

internal/neofs/tree.go
```diff
@@ -1022,7 +1025,18 @@ func (c *TreeClient) GetLastPart(ctx context.Context, bktInfo *data.BucketInfo,
 		return 1
 	})
 
-	return parts[len(parts)-1], nil
+	var pi *data.PartInfo
+	for _, part := range parts {
```
roman-khimov, line 1040: Can be a lot of parts.

api/layer/tree_mock.go
smallhive: What do you mean invalidate? There is no such functionality. Only [add and get](https://github.com/nspcc-dev/neofs-s3-gw/blob/06acb581f7b68f0b7412e0dd340fb25f75d7aef1/api/layer/tree_service.go#L82). Any reuploaded parts presented in the list.

internal/neofs/tree.go
smallhive: Yes, it can. Right now there is no mechanism to reduce their amount. We may have separate issue for this

internal/neofs/tree.go
smallhive: No, any part must have at least 5Mb

internal/neofs/tree.go
smallhive: Updated

api/handler/object_list.go
smallhive: Returned

cmd/s3-gw/app_settings.go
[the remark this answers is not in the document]
smallhive: Well, It is not clear for me why you don't like global loggers, but 2vs1 - removed

api/layer/tree_mock.go
smallhive: We may try to filter out duplicates on the level down and try to have there is only valid part list

roman-khimov: OK, seems to be unavoidable now.

internal/neofs/tree.go
roman-khimov: I mean NeoFS side. The first object can be created during upload init.

smallhive: Well, in this case In theory - yes. However, UploadID it just a `uuid` string. It is simple to generate.
Trying the approach with first object id as an UploadID requires many code changes, extended control for this _zero_ object during uploading and completing.
I see many problems and no profit for changing it right now.

roman-khimov: 1. Per S3 spec UploadID is just a string. It can be anything, including OID.
2. Client will use this UploadID for _every_ subsequent request which means you'll get this OID directly, you do not need to map UUID to first OID.
3. The first object would then be _always_ available, even if you're reuploading the first chunk.
4. There is nothing to control there, just create it for `CreateMultipartUpload`. Nothing you need to do with it afterwards (except for regular usage as first object OID).

smallhive: Do we do this inside this PR or in a separate issue?

roman-khimov: It's a part of v2 switch to me. It was specifically designed to allow for it.

api/layer/tree_service.go
carpawell: > V1 has only previous link

yes, but if you change the first one, you have to change `previous` for every next file, cause every next part should be recalculated... but it wasn't like that before?

carpawell: also, i saw that the last part is always zero-length (in NeoFS terms), can we put the first one with an empty payload too just to fix the first child ID? it will not be reuploaded then ever. @roman-khimov

roman-khimov: https://github.com/nspcc-dev/neofs-s3-gw/pull/957#discussion_r1634639794

carpawell: whoops, missed that, i am totally "for" uploading the first object as an initial multipart uploading ID and fixing this important split part

internal/neofs/tree.go
smallhive: Updated

roman-khimov (commented):
I've expected more red lines, but probably we're still too dependent on the tree service.

api/layer/multipart_upload.go
```diff
@@ -417,6 +420,114 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
...
+	encInfo := FormEncryptionInfo(multipartInfo.Meta)
+	if err := p.Encryption.MatchObjectEncryption(encInfo); err != nil {
+		n.log.Warn("mismatched obj encryptionInfo", zap.Error(err))
+		return nil, s3errors.GetAPIError(s3errors.ErrInvalidEncryptionParameters)
+	}
+
+	var (
+		bktInfo         = p.Bkt
+		attributes      [][2]string
+		multipartHash   = sha256.New()
```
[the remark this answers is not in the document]
smallhive: It is correct. Unfortunately, I failed to avoid it. Even empty data must be hashed and marshaled to be available for the next parts. This required calculating a hash for the whole object and the next part uses the "hash state" from the previous one

api/handler/object_list.go
```diff
@@ -277,7 +278,7 @@ func encodeListObjectVersionsToResponse(info *layer.ListObjectVersionsInfo, buck
 			Key:          ver.ObjectInfo.Name,
 			LastModified: ver.ObjectInfo.Created.UTC().Format(time.RFC3339),
 			Owner: Owner{
-				ID:          ver.ObjectInfo.Owner.String(),
+				ID:          hex.EncodeToString(ver.ObjectInfo.OwnerPublicKey.Bytes()),
```
roman-khimov, line 313: StringCompressed?

api/layer/tree_service.go
roman-khimov: So, do we still need this `number` parameter?

smallhive: Sure, we still have to take the previous part to take hashes from them. Just take the latest part isn't helpful for us

roman-khimov: But `GetLastPart` did exactly that without `number`, no?

smallhive: In common case of course, in reuploading case no
For instance: we have parts [1,2,3]. We are reuploading part 2. `GetLastPart` will return 3, which is not valid. GetPartByNumber will return 1 (because we take currentPartID - 1) - as the last stable point

roman-khimov (approved):

carpawell (approved):
This is (has been for 2 weeks) my first experience with our S3 review. Overall, looks working to me. However, I still do not understand why some changes are done _now_ and in the _replace v1 to v2_ commit. We have not changed the split noticeably, it is just now more convenient and more informative for SNs, but it is still an object chain, and still, the whole chain is backward-dependent, so it is strange.

Also, reuploading parts logic blows my mind, but it is not this PR's scope.

PR #958, opened 2024-06-04
Update ACL structures to have flexible types

smallhive:
The PR is not solving test issues completely. It gives the ability to use wildcards in `Principal` and `Action` fields

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
 	"crypto/elliptic"
@@ -56,6 +57,8 @@ var (
 	errInvalidPublicKey = errors.New("invalid public key")
 )
 
+var rawWildcardJSON = []byte(`"*"`)
+
 const (
 	arnAwsPrefix     = "arn:aws:s3:::"
 	allUsersWildcard = "*"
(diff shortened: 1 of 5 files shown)


api/handler/acl.go
```diff
@@ -56,6 +57,8 @@ var (
 	errInvalidPublicKey = errors.New("invalid public key")
 )
 
+var rawWildcardJSON = []byte{'"', '*', '"'}
```
roman-khimov, line 60: []byte(`"*"`) is likely more readable.

api/handler/json_types.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type (
+	stringOrSlice struct {
+		values []string
+	}
+)
+
+func (s *stringOrSlice) UnmarshalJSON(bytes []byte) error {
+	if bytes[0] == '[' {
```
roman-khimov, line 14: `  [....` can be a valid JSON too.

api/handler/acl.go
smallhive: Updated

api/handler/json_types.go
smallhive: Didn't catch your idea. In the case of `[` we have an array. If this array is not a strings array, it is not our problem and there will be an unmarshal error

smallhive (commented):
Extended docs

api/handler/json_types.go
roman-khimov: `[...` vs `     [...`, both are valid JSON.

smallhive: Added test with spaces. In case we have the next JSON with spaces:
```json
{    "action"     :      [     "s3:PutObject"     ]     }
```
`bytes` will contain `[     "s3:PutObject"     ]`.
All spaces will be skipped, https://go.dev/src/encoding/json/decode.go:178

smallhive: Updated

roman-khimov (approved):

PR #970, opened 2024-06-26
Allow to configure multiple rpc endpoints

smallhive:


api/resolver/init.go
```diff
@@ -17,7 +17,12 @@ import (
 // The Container should be used like regular [NNSResolver].
 type Container struct {
 	mu       sync.RWMutex
-	resolver *NNSResolver
+	resolver NNSResolver
+}
+
+// NNSResolver allows to resolve container id by its name.
+type NNSResolver interface {
```
roman-khimov, line 24: Interface with the only implementation in the same package can't exist.

config/config.yaml
```diff
@@ -49,7 +49,9 @@ tree:
   service: node1.neofs:8080
 
 # RPC endpoint and order of resolving of bucket names
-rpc_endpoint: http://morph-chain.neofs.devenv:30333
+morph:
```
roman-khimov, line 52: Let's call this `fschain` immediately. We will eventually remove `morph` from configs like this.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ This document outlines major changes between releases.
 
 ## [Unreleased]
 
+### Changed
+- Single `rpc_endpoint` config option replaced with multiple endpoints `morph.endpoints` option (#968)
+
```
roman-khimov, line 9: Migration guide is needed.

smallhive: Added

api/resolver/init.go
smallhive: Sure, it was too much

config/config.yaml
smallhive: Renamed

roman-khimov (approved):
I'm not excited about error handling here (if case some node is down), but it's already something.

PR #971, opened 2024-06-28
Remove actual object version

### LINKED ISSUE #955: After a parallel deletion of versioned objects, list objects returns info about them
s3tests_boto3.functional.test_s3#test_multipart_upload_overwrite_existing_object

Create versioned objects
Issue parallel deletion commands to the same objects
Issue list objects
Expect to see an empty list, but got an info about deleted objects

https://rest.fs.neo.org/7ZVSQHtioVRiDT2btN9xEmQhczXV9vryCEAHUawfD8Sc/239-1715825418/index.html#suites/6d7c9ec7c4f3e4460ae14fd266a64e04/c644040f4856b216/

smallhive:
Closes #955.

### DIFF
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -569,7 +569,7 @@ func (n *layer) deleteObject(ctx context.Context, bkt *data.BucketInfo, settings
 		}
 
 		if obj.DeleteMarkVersion, obj.Error = n.removeOldVersion(ctx, bkt, nodeVersion, obj); obj.Error != nil {
-			return obj
+			n.log.Info("remove old version", zap.Error(obj.Error))
 		}
 
 		obj.Error = n.treeService.RemoveVersion(ctx, bkt, nodeVersion.ID)


api/layer/layer.go
```diff
@@ -569,7 +569,7 @@ func (n *layer) deleteObject(ctx context.Context, bkt *data.BucketInfo, settings
 		}
 
 		if obj.DeleteMarkVersion, obj.Error = n.removeOldVersion(ctx, bkt, nodeVersion, obj); obj.Error != nil {
-			return obj
+			n.log.Error("remove old version", zap.Error(obj.Error))
```
roman-khimov, line 572: Why does it happen? I'm not sure it's "Error" level, maybe it's perfectly fine in some cases.

smallhive: I didn't find the exact reason for the problem, but it is directly connected to the multithreading in the test. Without threads, the code works perfectly. I consider that in some cases not all objects are removed because some error handling returns as here. Well, the `error` maybe too much, replaces with `Info`

cthulhu-rider (approved):

api/layer/layer.go
cthulhu-rider: @smallhive do we know exact `obj.Error` which caused test failures? May be we can exclude only "normal" errors here

smallhive: I tried this approach. The error in this place is Object not found and sometimes Object already removed or something like that.
In this line we are trying to remove some _old_ version. Returning from here allows for the  _actual_ version be not removed, and this fix removes all versions.
Anyway, this line fixes the tests and don't mark any other tests with red

PR #974, opened 2024-08-01
Update multipart

smallhive:


internal/neofs/neofs.go
```diff
@@ -284,13 +284,11 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
...
 		if prm.Multipart.HeaderObject != nil {
-			id, isSet := prm.Multipart.HeaderObject.ID()
-			if !isSet {
-				return oid.ID{}, errors.New("HeaderObject id is not set")
-			}
-
-			obj.SetParentID(id)
 			obj.SetParent(prm.Multipart.HeaderObject)
+
+			if id, isSet := prm.Multipart.HeaderObject.ID(); isSet {
```
roman-khimov, line 289: So what if `!isSet`?

roman-khimov (approved):
Seems to be legit.

carpawell (approved):

internal/neofs/neofs.go
smallhive: Nothing, it is ok situation, according to the split v2

PR #976, opened 2024-08-09
Accounts in eacl

smallhive:
Refs https://github.com/nspcc-dev/neofs-api/issues/278

The changes require test changes as well.

ACL owner will be changed from
```json
{
    "Owner": {
        "DisplayName": "Nj9FF9jYTsyX2XniTw5Kqy3Zf91nbhdYiM",
        "ID": "033845e8ebc78251029261cbc0e9bae104d2b7cc9fee5c79720a30332d5b3d1aa0"
    },
    "Grants": [
        {
            "Grantee": {
                "DisplayName": "Nj9FF9jYTsyX2XniTw5Kqy3Zf91nbhdYiM",
                "ID": "033845e8ebc78251029261cbc0e9bae104d2b7cc9fee5c79720a30332d5b3d1aa0",
                "Type": "CanonicalUser"
            },
            "Permission": "FULL_CONTROL"
        }
    ]
}
```
to
```json
{
    "Owner": {
        "DisplayName": "Nj9FF9jYTsyX2XniTw5Kqy3Zf91nbhdYiM",
        "ID": "Nj9FF9jYTsyX2XniTw5Kqy3Zf91nbhdYiM"
    },
    "Grants": [
        {
            "Grantee": {
                "DisplayName": "Nj9FF9jYTsyX2XniTw5Kqy3Zf91nbhdYiM",
                "ID": "Nj9FF9jYTsyX2XniTw5Kqy3Zf91nbhdYiM",
                "Type": "CanonicalUser"
            },
            "Permission": "FULL_CONTROL"
        }
    ]
}
```

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -1568,21 +1568,6 @@ func isValidGrant(grant *Grant) bool {
 		(grant.Grantee.Type == granteeCanonicalUser || (grant.Grantee.Type == granteeGroup && grant.Grantee.URI == allUsersGroup))
 }
 
-func getAllowRecord(op eacl.Operation, pk *keys.PublicKey) *eacl.Record {
-	record := eacl.NewRecord()
-	record.SetOperation(op)
-	record.SetAction(eacl.ActionAllow)
-
-	t := eacl.NewTarget()
-	// Unknown role is used, because it is ignored when keys are set
-	t.SetRole(eacl.RoleUnknown)
-	t.SetAccounts([]user.ID{user.NewFromScriptHash(pk.GetScriptHash())})
-
-	record.SetTargets(*t)
-
-	return record
-}
-
 func getAllowRecordWithUser(op eacl.Operation, acc user.ID) *eacl.Record {
 	record := eacl.NewRecord()
 	record.SetOperation(op)
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -3,9 +3,7 @@ package handler
 import (
 	"bytes"
 	"context"
-	"crypto/ecdsa"
 	"crypto/elliptic"
-	"encoding/hex"
 	"encoding/json"
 	"encoding/xml"
 	"errors"
@@ -24,6 +22,7 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/session"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"go.uber.org/zap"
 )
 
@@ -175,7 +174,7 @@ func (r *resourceInfo) IsBucket() bool {
 }
 
 type astOperation struct {
-	Users  []string
+	Users  []user.ID
 	Op     eacl.Operation
 	Action eacl.Action
 }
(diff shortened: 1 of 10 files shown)


api/handler/acl.go
```diff
@@ -493,14 +493,18 @@ func (h *handler) PutBucketPolicyHandler(w http.ResponseWriter, r *http.Request)
 }
 
 func parseACLHeaders(header http.Header, key *keys.PublicKey) (*AccessControlPolicy, error) {
```
roman-khimov, line 495: Why keeping the key here?

api/handler/acl.go
```diff
@@ -946,16 +954,12 @@ func formRecords(resource *astResource) ([]*eacl.Record, error) {
...
-			targetKeys := make([]ecdsa.PublicKey, 0, len(astOp.Users))
-			for _, user := range astOp.Users {
-				pk, err := keys.NewPublicKeyFromString(user)
-				if err != nil {
-					return nil, errInvalidPublicKey
-				}
-				targetKeys = append(targetKeys, (ecdsa.PublicKey)(*pk))
-			}
+			t := eacl.NewTarget()
 			// Unknown role is used, because it is ignored when keys are set
```
roman-khimov, line 958: keys?

api/handler/acl.go
```diff
@@ -989,18 +993,38 @@ func addToList(operations []*astOperation, rec eacl.Record, target eacl.Target)
 	if found != nil {
 		if !groupTarget {
 			for _, key := range target.BinaryKeys() {
```
roman-khimov, line 995: Can be removed, I think. Only S3 gateways sets them and if it only sets accounts then we may not care about keys.

api/handler/acl.go
```diff
@@ -1011,8 +1035,15 @@ func addToList(operations []*astOperation, rec eacl.Record, target eacl.Target)
 		}
 		if !groupTarget {
 			for _, key := range target.BinaryKeys() {
```
roman-khimov, line 1037: Same here.

api/handler/acl.go
```diff
@@ -1358,13 +1424,15 @@ func permissionToOperations(permission amazonS3Permission) []eacl.Operation {
 }
 
 func encodeObjectACL(log *zap.Logger, bucketACL *layer.BucketACL, bucketName, objectVersion string) *AccessControlPolicy {
+	account := user.NewFromScriptHash(bucketACL.Info.OwnerPublicKey.GetScriptHash()).String()
```
roman-khimov, line 1427: Why not using the owner account? You no longer need the key.

api/handler/acl.go
smallhive: Ok, removed

api/handler/acl.go
smallhive: Removed

api/handler/acl.go
smallhive: I didn't want to make strong steps with key usage. Replaced with `Owner`

api/handler/acl.go
roman-khimov: But I think the only thing you do now with this key is getting an account. In which case the key is irrelevant and as S3 never sets any keys into ACLs, why should it care about them during parsing? I do want to make this easier to use in the end, we only have keys in NeoFS itself for compatibility and that can be reconsidered after some transition period.

smallhive: I agree with this point of view. Replaced with `user.ID`

roman-khimov (approved):

PR #1000, opened 2024-09-05
workflow: Update component versions

smallhive:


roman-khimov:
Ouch. I thought we're up to date.

smallhive:
It is just a theory because locally all tests are green. I run them many times in different variations

roman-khimov:
We need to update anyway.

roman-khimov (approved):

PR #1001, opened 2024-09-10
Sign multipart objects in the gate

### LINKED ISSUE #975: Sign multiparts and do not use sessions
S3 gateway slices big objects if it is required (multipart API is used). It does almost all the work neofs-node usually does, so no code simplification from opening a session is achieved, so why not just finish the job that is 90% done?  

Sign multiparts before sending to NeoFS, do not open/store/cache a session, it will not be required.

Do nothing, it already works (but looks strange when you tell about it works to _any_ neofs dev).

## Additional context
Can be a solution for https://github.com/nspcc-dev/neofs-s3-gw/issues/969 (however, neofs-node is also guilty here).

smallhive:
Closes #975.

### DIFF
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -488,6 +488,20 @@ func (n *layer) objectPutAndHash(ctx context.Context, prm PrmObjectCreate, bktIn
 	hash := sha256.New()
 	prm.Payload = wrapReader(prm.Payload, 64*1024, func(buf []byte) {
 		hash.Write(buf)
+	})
+	id, err := n.neoFS.CreateObject(ctx, prm)
+	if err != nil {
+		return oid.ID{}, nil, err
+	}
+	return id, hash.Sum(nil), nil
+}
+
+// multipartObjectPut writes only Multipart hash collection.
+// It doesn't calculate hash from payload, it already calculated in prm.Multipart.PayloadHash, if required.
+func (n *layer) multipartObjectPut(ctx context.Context, prm PrmObjectCreate, bktInfo *data.BucketInfo) (oid.ID, error) {
+	n.prepareAuthParameters(ctx, &prm.PrmAuth, bktInfo.Owner)
+
+	prm.Payload = wrapReader(prm.Payload, 64*1024, func(buf []byte) {
 		if prm.Multipart != nil {
 			for _, h := range prm.Multipart.MultipartHashes {
 				h.Write(buf)
@@ -496,9 +510,9 @@ func (n *layer) objectPutAndHash(ctx context.Context, prm PrmObjectCreate, bktIn
 	})
 	id, err := n.neoFS.CreateObject(ctx, prm)
 	if err != nil {
-		return oid.ID{}, nil, err
+		return oid.ID{}, err
 	}
-	return id, hash.Sum(nil), nil
+	return id, nil
 }
 
 // ListObjectsV1 returns objects in a bucket for requests of Version 1.
--- a/internal/neofs/neofs.go
+++ b/internal/neofs/neofs.go
@@ -26,6 +26,7 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/container"
 	"github.com/nspcc-dev/neofs-sdk-go/container/acl"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	"github.com/nspcc-dev/neofs-sdk-go/eacl"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
@@ -246,6 +247,19 @@ func (x *NeoFS) DeleteContainer(ctx context.Context, id cid.ID, token *session.C
 	return nil
 }
 
+func (x *NeoFS) signMultipartObject(obj *object.Object, signer neofscrypto.Signer, payloadHash, homoHash hash.Hash) error {
+	obj.SetPayloadChecksum(checksum.NewFromHash(checksum.SHA256, payloadHash))
+	if homoHash != nil {
+		obj.SetPayloadHomomorphicHash(checksum.NewFromHash(checksum.TillichZemor, homoHash))
+	}
+
+	if err := obj.SetIDWithSignature(signer); err != nil {
+		return fmt.Errorf("set id with signature: %w", err)
+	}
+
+	return nil
+}
+
 // CreateObject implements neofs.NeoFS interface method.
 func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oid.ID, error) {
 	attrNum := len(prm.Attributes) + 1 // + creation time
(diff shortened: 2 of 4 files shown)


internal/neofs/neofs.go
```diff
@@ -310,6 +342,10 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
 			prm.Payload = bytes.NewReader(obj.Payload())
 			obj.SetPayloadSize(uint64(len(obj.Payload())))
 
+			if err := x.signMultipartObject(&obj, x.signer(ctx)); err != nil {
+				return oid.ID{}, errors.New("object sign failed")
```
roman-khimov, line 346: failed to sign object

internal/neofs/neofs.go
```diff
@@ -246,6 +247,37 @@ func (x *NeoFS) DeleteContainer(ctx context.Context, id cid.ID, token *session.C
...
 }
 
+func (x *NeoFS) signMultipartObject(header *object.Object, signer neofscrypto.Signer) error {
+	var (
+		payloadChecksum = sha256.New()
+		sig             neofscrypto.Signature
+	)
+
+	payloadChecksum.Write(header.Payload())
+	header.SetPayloadChecksum(checksum.NewFromHash(checksum.SHA256, payloadChecksum))
```
roman-khimov, line 257: I thought you have them somewhere already for multiparts. Or you can calculate them on the go, just like regular ones. The rationale is that you'd have warmer caches then.

internal/neofs/neofs.go

internal/neofs/neofs.go
smallhive: It exists, but it calculates on the fly during payload writing to the network. Updated to recalculated hash

roman-khimov: But tz is still the same.

smallhive: Yes, we calculate `tz` for the whole Big object. Separately for each part, we didn't calculate it before. The node did it on my behalf. Now we calculate it in place

api/layer/object.go
```diff
@@ -488,17 +488,32 @@ func (n *layer) objectPutAndHash(ctx context.Context, prm PrmObjectCreate, bktIn
...
+	n.prepareAuthParameters(ctx, &prm.PrmAuth, bktInfo.Owner)
+
+	prm.Payload = wrapReader(prm.Payload, 64*1024, func(buf []byte) {
 		if prm.Multipart != nil {
 			for _, h := range prm.Multipart.MultipartHashes {
 				h.Write(buf)
 			}
 		}
 	})
+
```
carpawell, line 511: space change

internal/neofs/neofs.go
```diff
@@ -246,6 +247,35 @@ func (x *NeoFS) DeleteContainer(ctx context.Context, id cid.ID, token *session.C
...
+	var (
+		sig neofscrypto.Signature
+	)
+
+	header.SetPayloadChecksum(checksum.NewFromHash(checksum.SHA256, payloadHash))
+
+	if x.IsHomomorphicHashingEnabled() {
+		var homoChecksum = tz.New()
+
+		homoChecksum.Write(header.Payload())
```
carpawell, line 260: `header` but `Payload()`?

internal/neofs/neofs.go
```diff
@@ -246,6 +247,35 @@ func (x *NeoFS) DeleteContainer(ctx context.Context, id cid.ID, token *session.C
...
+	if err != nil {
+		return fmt.Errorf("calculate ID: %w", err)
+	}
+	header.SetID(id)
+
+	if err = sig.Calculate(signer, id.Marshal()); err != nil {
+		return fmt.Errorf("sign object ID: %w", err)
+	}
+
+	header.SetSignature(&sig)
```
carpawell, line 274: just `CalculateAndSetID` and not 4 method calls above?

carpawell (commented):
Dont we need some tests for it? Test code can be unified with SDK/node. Otherwise, it is impossible to keep everything in sync to me. Multiple codes do the same thing.

api/layer/object.go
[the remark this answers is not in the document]
smallhive: It is a completely new function! But git shows it strangely. BTW removed an empty row

internal/neofs/neofs.go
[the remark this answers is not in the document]
smallhive: Correct, is it wrong? Each part should have a tz checksum for it. For instance, the slicer works in the same way

smallhive:
> Dont we need some tests for it? Test code can be unified with SDK/node. Otherwise, it is impossible to keep everything in sync to me. Multiple codes do the same thing.

I checked the code with s3 gate tests locally and GH pipeline did it as well

api/layer/object.go
smallhive: Removed

internal/neofs/neofs.go
smallhive: But variable name, replaced with `obj`

internal/neofs/neofs.go
smallhive: It seems shorter, replaced

api/layer/object.go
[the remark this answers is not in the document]
roman-khimov: diff only cares about lines, so if there is any similar code already the result can be surprising.

internal/neofs/neofs.go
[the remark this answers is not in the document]
roman-khimov: Sure. But you've just moved SHA256 calculation out of this function. Why not doing this for tz?

internal/neofs/neofs.go
carpawell: @smallhive, my bad, i meant `SetIDWithSignature` but copied a different method name. it allows even fewer code lines for the same functionality

carpawell:
> s3 gate tests locally and GH pipeline

@smallhive, i guess without https://github.com/nspcc-dev/neofs-s3-gw/pull/974 it also worked well but contained errors, right? I do not mean this must be done in this PR (at least something should appear in SDK) but I think we can discuss it at least. @roman-khimov

internal/neofs/neofs.go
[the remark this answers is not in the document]
smallhive: It was a double-content calculation in the first version. I separated it into another function to remove it. This version contains less code. Replaced

internal/neofs/neofs.go
smallhive: This is even better, updated. We should make such changes in the Slicer as well. I inspired it's code

smallhive:
> @smallhive, i guess without #974 it also worked well but contained errors, right? I do not mean this must be done in this PR (at least something should appear in SDK) but I think we can discuss it at least. @roman-khimov

It's make sence

internal/neofs/neofs.go
```diff
@@ -306,13 +320,31 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
 		}
 
 		if prm.Multipart.Link != nil {
+			// pl := obj.Payload()
```
carpawell, line 323: not needed?

smallhive: indeed

roman-khimov (approved):

carpawell (approved):

PR #1002, opened 2024-09-10
Update default timeout

### LINKED ISSUE #973: default timeouts are extremely aggressive
1. connection timeouts less than 10s are extremely aggressive and leads to connection breaks
2. connection timeouts should be the same across all services
3. i'd recommend to set all network timeouts to 1 minute

smallhive:
Closes #973.

Node timeout left as is, 10s. According to the [REST gate](https://github.com/nspcc-dev/neofs-rest-gw/pull/244/files#diff-38df760413887d0cc4ee5707919e6390456a863d326f6080644cc3d9c72e4bbaR22)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This document outlines major changes between releases.
 
 ### Changed
 - Go 1.22+ is required to build now (#927)
+- stream_timeout default timeout from 10s to 60s (#973)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This document outlines major changes between releases.
 
 ### Changed
 - Go 1.22+ is required to build now (#927)
+- stream_timeout default timeout (#973)
```
roman-khimov, line 11: It's be nice to mention new timeout value.

smallhive: It makes sense, updated

roman-khimov (approved):

PR #1004, opened 2024-09-13
Org-wide linter fixes

roman-khimov:


roman-khimov:
Argh, `make sync-tree`. This complicates things considerably.

roman-khimov:
Suggest merging the fixes (they're good anyway), but not switching the the linter setup for now.

PR #1018, opened 2024-10-30
Disable ACL by default

### LINKED ISSUE #1016: Object PUTs into private buckets lead to EACL bloat
## Current Behavior
PutObject with full owner control ACL leading to EACL bloat and eventual 500 with `container setEacl: status: code = 1024 message = Invalid params (-32602) - byte-slice is too big (67987)` error.

## Expected Behavior
Buckets are private by default. Puts with full object control for objects do nothing EACL-wise as explained in https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3/s3manager#UploadInput and its references. Other ACLs are rejected unless ACLs are enabled (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html also).

## Possible Solution
Fix the damn things.

## Steps to Reproduce
`aws s3api put-object --bucket amzn-s3-demo-bucket --key key-name --body path-to-file --acl bucket-owner-full-control`

## Context
Real software loves specifying ACLs for whatever reason.

## Your Environment
* Version of the product used: 0.32.0
* Server setup and configuration files used:
* Operating System and version (`uname -a`): Debian stable

smallhive:
Closes #1016.


api/cache/eacls.go
```diff
@@ -0,0 +1,71 @@
...
+	// BucketACLStateCache contains cache with bucket ACL state.
+	BucketACLStateCache struct {
+		cache  gcache.Cache
+		logger *zap.Logger
+	}
+)
+
+const (
+	// DefaultEACLCacheSize is a default maximum number of entries in cache.
+	DefaultEACLCacheSize = 1e6
```
roman-khimov, line 23: Too much.

api/handler/copy.go
```diff
@@ -91,6 +91,17 @@ func (h *handler) CopyObjectHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if containsACL {
+		state, err := h.cachedACLGetter.GetState(r.Context(), dstBktInfo.CID)
+		if err != nil {
+			h.logAndSendError(w, "could not get bucket acl state", reqInfo, err)
+			return
+		}
+
+		if state == data.BucketACLBucketOwnerEnforced {
```
roman-khimov, line 100: Looks like you're not handling "PUT requests that specify bucket owner full control ACLs, such as the bucket-owner-full-control canned ACL" correctly.

api/handler/multipart_upload.go
```diff
@@ -113,6 +113,17 @@ func (h *handler) CreateMultipartUploadHandler(w http.ResponseWriter, r *http.Re
 	}
 
 	if containsACLHeaders(r) {
+		state, err := h.cachedACLGetter.GetState(r.Context(), bktInfo.CID)
+		if err != nil {
+			h.logAndSendError(w, "could not get bucket acl state", reqInfo, err)
+			return
+		}
+
+		if state == data.BucketACLBucketOwnerEnforced {
```
roman-khimov, line 122: And here.

api/handler/put.go
```diff
@@ -203,6 +206,19 @@ func (h *handler) PutObjectHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	if containsACL {
+		state, err := h.cachedACLGetter.GetState(r.Context(), bktInfo.CID)
+		if err != nil {
+			h.logAndSendError(w, "could not get bucket acl state", reqInfo, err)
+			return
+		}
+
+		if state == data.BucketACLBucketOwnerEnforced {
```
roman-khimov, line 219: And here.

api/handler/put.go
```diff
@@ -168,6 +168,9 @@ const (
 	basicACLReadOnly  = "public-read"
 	basicACLPublic    = "public-read-write"
 	cannedACLAuthRead = "authenticated-read"
+
+	amzBucketOwnerField    = "x-amz-object-ownership"
```
roman-khimov, line 173: Can't this collide with some real header? I'd expect a bit more magic value here. Although if the rule is always the last one (after default policies) it can't change much.

api/handler/put.go
```diff
@@ -436,6 +452,19 @@ func (h *handler) PostObject(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	if containsACL {
+		state, err := h.cachedACLGetter.GetState(r.Context(), bktInfo.CID)
+		if err != nil {
+			h.logAndSendError(w, "could not get bucket acl state", reqInfo, err)
+			return
+		}
+
+		if state == data.BucketACLBucketOwnerEnforced {
```
roman-khimov, line 474: And here.

api/handler/copy.go
```diff
@@ -101,6 +101,15 @@ func (h *handler) CopyObjectHandler(w http.ResponseWriter, r *http.Request) {
 			h.logAndSendError(w, "access control list not supported", reqInfo, s3errors.GetAPIError(s3errors.ErrAccessControlListNotSupported))
 			return
 		}
+		if state == data.BucketACLBucketOwnerPreferred {
```
roman-khimov, line 104: I expect ACLs to work properly in this case.

api/handler/copy.go
smallhive: Updated

api/handler/put.go
smallhive: It would and there is no guarantee the rule is always be the last, renamed

api/handler/copy.go
smallhive: Yeah, it seems it should work as was before. Removed

api/handler/json_types.go
```diff
@@ -37,3 +38,14 @@ func (s stringOrSlice) MarshalJSON() ([]byte, error) {
 
 	return json.Marshal(values)
 }
+
+func (s stringOrSlice) Equal(v stringOrSlice) bool {
```
roman-khimov, line 42: Looks like this belongs to a different commit.

roman-khimov (approved):

roman-khimov (approved):

PR #1026, opened 2024-11-20
Use PutBucketOwnershipControls instead of PutBucketPolicy

### LINKED ISSUE #1023: Don't rely on Sid for any policies
## Current Behavior
#1018 has introduced some logic dependent on `BucketEnableACL` Sid, but it's absolutely incorrect. 

## Expected Behavior
Sid is informational and can be named in any way, what matters is Action/Resource/Effect.
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_sid.html

## Possible Solution
Do not check for Sid, check contents only.

## Context
#1016, https://github.com/nspcc-dev/neofs-testcases/pull/894

smallhive:
Closes #1023.


api/handler/acl.go
```diff
@@ -1708,11 +1760,87 @@ func isBucketOwnerForced(table *eacl.Table) bool {
...
+func isBucketOwnerPreferred(table *eacl.Table) bool {
+	if table == nil {
+		return false
+	}
+
+	for _, r := range table.Records() {
+		if r.Action() == eacl.ActionDeny && r.Operation() == eacl.OperationPut {
+			for _, f := range r.Filters() {
+				if f.Key() == amzBucketOwnerField &&
+					f.Value() == amzBucketOwnerPreferred &&
```
roman-khimov, line 1766: Why these filters still if you're using target accounts as well? I thought a single special entity (be it a filter or user) is enough.

api/handler/acl.go
```diff
@@ -393,6 +420,14 @@ func (h *handler) PutObjectACLHandler(w http.ResponseWriter, r *http.Request) {
 		r.Header.Set(api.AmzACL, "")
 	}
 
+	if isBucketOwnerPreferred(eacl.EACL) {
+		if !isValidOwnerPreferred(r) {
+			h.logAndSendError(w, "header x-amz-acl:bucket-owner-full-control must be set", reqInfo, s3errors.GetAPIError(s3errors.ErrAccessDenied))
```
roman-khimov, line 425: Not sure it's a correct behavior for "preferred" case.

api/handler/copy.go
```diff
@@ -111,6 +111,14 @@ func (h *handler) CopyObjectHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if isBucketOwnerPreferred(eacl.EACL) {
```
roman-khimov, line 114: Why doing it for `!containsACL`?

docs/aws_s3_compat.md
```diff
@@ -290,41 +309,6 @@ See also `GetObject` and other method parameters.
 | 🟡 | PutBucketPolicy         | See ACL limitations         |
 | 🔵 | PutBucketReplication    |                             |
 
-By default bucket ACLs is disabled. See details [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html).
```
roman-khimov, line 293: This should be noted in the ACL section with reference to "ownership controls".

roman-khimov (commented):
Documentation is not updated for get/delete.

evgeniiz321:
@roman-khimov I can verify that PutBucketOwnershipControls is the only _right_ way to disable/enable ACLs. 
All kind of operations with policies are only intended to either make people to disable ACLs (via PutBucketOwnershipControls) for new buckets or somehow emulate the disabled ACL behavior - e.g. make people to use only specific type of ACL and return errors otherwise. The question is do we want to support such policies? If we do - I would suggest to add a support of these type of conditions:

```
            "Condition": {
                "StringNotEquals": {
                    "s3:x-amz-object-ownership": "BucketOwnerEnforced"
                }
            }
 ``` 
 ```
          "Condition": {
            "StringEquals": {
               "s3:x-amz-acl": "bucket-owner-full-control"
            }
         }
  ```

api/handler/acl.go
smallhive: I had the same thought about it, but internally the gate makes some conversation between EACL <-> AST <-> POLICY formats. Without all these checks it is impossible to save all required context when some ACL is applied to an object/bucket

api/handler/copy.go
smallhive: `containsACL` means the user has passed some ACL headers. This check is not related to passed parameters. This check is about the actual ACL and ownership mode

docs/aws_s3_compat.md
smallhive: Moved

smallhive:
> Documentation is not updated for get/delete.

Updated

api/handler/acl.go
smallhive: Do you mean `PutObjectACLHandler` or such behavior in general?

roman-khimov:
"Special" ACLs can be moved to a separate issue (and documented for now), majority of users are expected to be OK with disabled ones (where this setting doesn't matter) and if they're enabled then likely people want something more interesting than these ones.

api/handler/acl.go
roman-khimov: Requiring one specific ACL.

api/handler/copy.go
roman-khimov: But if there are no ACLs passed we can just do whatever is appropriate in the default case, no?

smallhive: I think in the `BucketOwnerPreferred` mode, the header `bucket-owner-full-control` must be set

roman-khimov: I hope @evgeniiz321 will figure it out.

roman-khimov (approved):
Let's merge and see what tests will show. The main part of the problem is solved, details can be fine-tuned separately if needed.

PR #1034, opened 2024-11-29
Parallel uploads for multipart

### LINKED ISSUE #1020: Parallel uploads for multipart
Multipart uploads don't work in general case

## Current Behavior
AWS SDK uploads _parts_ for _multipart_ in 5 parallel threads. The gate expects parts subsequently one by one

## Expected Behavior
All parts should be uploaded in any order

## Possible Solution
Collect the final object hash in a different way

## Steps to Reproduce
- Create multipart upload with any available tool
- Upload parts in not subsequent order and get `OperationAborted: 409` error

## Context
Related to #1016

## Your Environment

* Version of the product used: [59db6be2c65b89aaa94047d98cb1fb38bc4393c3](https://github.com/nspcc-dev/neofs-s3-gw/commit/968ea3a608db8d7763c03c1dc9f5af7f9c2d87b1)

smallhive:
Closes #1020

### DIFF
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -42,6 +42,10 @@ const (
 	UploadMaxPartNumber = 10000
 	uploadMinSize       = 5 * 1048576    // 5MB
 	uploadMaxSize       = 5 * 1073741824 // 5GB
+
+	headerS3MultipartUpload  = "S3MultipartUpload"
+	headerS3MultipartNumber  = "S3MultipartNumber"
+	headerS3MultipartCreated = "S3MultipartCreated"
 )
 
 type (
@@ -140,6 +144,24 @@ type (
 		OwnerPubKey keys.PublicKey
 		Created     time.Time
 	}
+
+	slotAttributes struct {
+		PartNumber int64
+		// in nanoseconds
+		CreatedAt int64
+		FilePath  string
+	}
+
+	uploadPartAsSlotParams struct {
+		bktInfo          *data.BucketInfo
+		multipartInfo    *data.MultipartInfo
+		tzHash           hash.Hash
+		attributes       [][2]string
+		uploadPartParams *UploadPartParams
+		creationTime     time.Time
+		payloadReader    io.Reader
+		decSize          int64
+	}
 )
 
 func (n *layer) CreateMultipartUpload(ctx context.Context, p *CreateMultipartParams) (string, error) {
@@ -220,10 +242,6 @@ func (n *layer) UploadPart(ctx context.Context, p *UploadPartParams) (string, er
 		return "", err
 	}
 
-	if err = n.reUploadFollowingParts(ctx, *p, p.PartNumber, p.Info.Bkt, multipartInfo); err != nil {
-		return "", fmt.Errorf("reuploading parts: %w", err)
-	}
-
 	return objInfo.HashSum, nil
 }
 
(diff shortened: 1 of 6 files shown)


api/layer/multipart_upload.go
```diff
@@ -1184,3 +1250,213 @@ func (n *layer) manualSlice(ctx context.Context, bktInfo *data.BucketInfo, prm P
...
+		return nil, fmt.Errorf("get part by number: %w", err)
+	}
+
+	var (
+		id              oid.ID
+		chunk           *[]byte
+		elements        []data.LinkObjectPayload
+		isReturnToPool  bool
+		splitFirstID    = zeroPart.OID
+		splitPreviousID = zeroPart.OID
```
roman-khimov, line 1268: Wow. So it's like a tree starting from the first OID? The idea was to make them completely separate. But if this works maybe it's a better scheme in the end (somewhat easier to clean up).

api/layer/multipart_upload.go
```diff
@@ -1184,3 +1250,213 @@ func (n *layer) manualSlice(ctx context.Context, bktInfo *data.BucketInfo, prm P
...
+		UploadID: params.uploadPartParams.Info.UploadID,
+		Number:   params.uploadPartParams.PartNumber,
+		OID:      id,
+		Size:     params.decSize,
+		ETag:     hex.EncodeToString(currentPartHash.Sum(nil)),
+		Created:  prm.CreationTime,
+		Elements: elements,
+	}
+
+	// encoding hash.Hash state to save it in tree service.
```
roman-khimov, line 1323: I think it's not necessary for these parts. These hashes can't be reused.

api/layer/multipart_upload.go
```diff
@@ -1184,3 +1250,213 @@ func (n *layer) manualSlice(ctx context.Context, bktInfo *data.BucketInfo, prm P
...
+	filters.AddFilter(headerS3MultipartUpload, uploadID, object.MatchStringEqual)
+
+	var prmSearch = PrmObjectSearch{
+		Container: bucketInfo.CID,
+		Filters:   filters,
+	}
+
+	n.prepareAuthParameters(ctx, &prmSearch.PrmAuth, bucketInfo.Owner)
+
+	oids, err := n.neoFS.SearchObjects(ctx, prmSearch)
```
roman-khimov, line 1407: Will be pain performance-wise, but the only alternative now is the tree.

roman-khimov (commented):
Nice overall.

api/layer/multipart_upload.go
smallhive: It is working for the `Slot` part and helps to avoid double reupload

api/layer/multipart_upload.go
smallhive: this method works for all types of parts, for complete reupload too. I prefer to have the single method without extra `if`

api/layer/multipart_upload.go
```diff
@@ -1080,12 +1080,14 @@ func (n *layer) getUploadParts(ctx context.Context, p *UploadInfoParams) (*data.
 			return -1
 		}
 
-		if a.ServerCreated.Before(b.ServerCreated) {
-			return -1
-		}
+		if a.Number == b.Number {
+			if a.ServerCreated.Before(b.ServerCreated) {
```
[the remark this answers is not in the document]
smallhive: Separated

roman-khimov (approved):

PR #1039, opened 2024-12-04
Change default chunk payload string length

### LINKED ISSUE #1032: Panic under high load
This problem is reproduced in 100% of cases, with 50 PUT threads per node (12core, 24GB RAM).

smallhive:
Closes #1032.

We should try it out


api/auth/signer/v4/chunked_reader.go
```diff
@@ -37,7 +40,7 @@ var (
 // The chunkedReader returns io.EOF when the final 0-length chunk is read.
 func NewChunkedReader(r io.ReadCloser, streamSigner *ChunkSigner) io.ReadCloser {
 	return &chunkedReader{
-		r: bufio.NewReader(r),
+		r: bufio.NewReaderSize(r, MaxChunkedPayloadLineLength),
```
roman-khimov, line 43: I don't think it's a fix. I can always create something that is larger than this limit and make the gateway panic which is unacceptable.

api/auth/signer/v4/chunked_reader.go
```diff
@@ -199,7 +202,7 @@ func readChunkLine(b *bufio.Reader) ([]byte, []byte, error) {
 		}
 		return nil, nil, err
 	}
-	if len(p) >= maxLineLength {
+	if len(p) >= MaxChunkedPayloadLineLength {
```
roman-khimov, line 205: And here it's too late, panic happens earlier than that. The real question is why `ReadSlice` doesn't return `ErrBufferFull` above.

smallhive: Moving these rows upper is not a solution either, because panic inside `bufio.ReadSlice` and we just don't have the opportunity to check the length

api/auth/signer/v4/chunked_reader.go
smallhive: I'm not sure too, but we would try

api/auth/signer/v4/chunked_reader.go
roman-khimov: I know, but we also need to know _why_ it panics. It shouldn't.

roman-khimov:
Can be related to multithreading in fact. Why does it happen in a separate routine? You can use https://pkg.go.dev/io#TeeReader instead of `wrapReader`.

smallhive:
> wrapReader

I don't think this fix will change the situation, but it looks better.

Maybe this panic is connected to https://github.com/golang/go/issues/22330

roman-khimov (approved):

PR #1041, opened 2024-12-06
Fix multipart

smallhive:
Ref https://github.com/nspcc-dev/neofs-testcases/pull/906

### DIFF
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -285,6 +285,7 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 	if err != nil {
 		return nil, fmt.Errorf("getLastPart: %w", err)
 	}
+	reqInfo := api.GetReqInfo(ctx)
 
 	// The previous part is not uploaded yet.
 	if lastPart == nil {
@@ -303,6 +304,12 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 		if err != nil {
 			return nil, err
 		}
+
+		n.log.Debug("upload part as slot",
+			zap.String("reqId", reqInfo.RequestID),
+			zap.String("bucket", bktInfo.Name), zap.Stringer("cid", bktInfo.CID),
+			zap.String("multipart upload", p.Info.UploadID),
+			zap.Int("part number", p.PartNumber), zap.String("object", p.Info.Key), zap.Stringer("oid", objInfo.ID), zap.String("ETag", objInfo.HashSum), zap.Int64("decSize", decSize))
 
 		return objInfo, nil
 	}
@@ -372,13 +379,6 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 		n.buffers.Put(chunk)
 	}
 
-	reqInfo := api.GetReqInfo(ctx)
-	n.log.Debug("upload part",
-		zap.String("reqId", reqInfo.RequestID),
-		zap.String("bucket", bktInfo.Name), zap.Stringer("cid", bktInfo.CID),
-		zap.String("multipart upload", p.Info.UploadID),
-		zap.Int("part number", p.PartNumber), zap.String("object", p.Info.Key), zap.Stringer("oid", id))
-
 	partInfo := &data.PartInfo{
 		Key:      p.Info.Key,
 		UploadID: p.Info.UploadID,
@@ -389,6 +389,12 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 		Created:  prm.CreationTime,
 		Elements: elements,
 	}
+
+	n.log.Debug("upload part",
+		zap.String("reqId", reqInfo.RequestID),
+		zap.String("bucket", bktInfo.Name), zap.Stringer("cid", bktInfo.CID),
+		zap.String("multipart upload", p.Info.UploadID),
+		zap.Int("part number", p.PartNumber), zap.String("object", p.Info.Key), zap.Stringer("oid", id), zap.String("ETag", partInfo.ETag), zap.Int64("decSize", decSize))
 
 	// encoding hash.Hash state to save it in tree service.
 	// the required interface is guaranteed according to the docs, so just cast without checks.
@@ -606,6 +612,7 @@ func (n *layer) reUploadPart(ctx context.Context, uploadParams UploadPartParams,
 	uploadParams.Size = int64(obj.PayloadSize())
 	uploadParams.Reader = bytes.NewReader(obj.Payload())
 
+	n.log.Debug("reUploadPart", zap.String("oid", id.String()), zap.Uint64("payload size", obj.PayloadSize()))
 	if _, err = n.uploadPart(ctx, multipartInfo, &uploadParams); err != nil {
 		return fmt.Errorf("upload id=%s: %w", id.String(), err)
 	}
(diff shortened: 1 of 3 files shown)


api/layer/multipart_upload.go
```diff
@@ -606,6 +612,7 @@ func (n *layer) reUploadPart(ctx context.Context, uploadParams UploadPartParams,
 	uploadParams.Size = int64(obj.PayloadSize())
 	uploadParams.Reader = bytes.NewReader(obj.Payload())
 
+	n.log.Error("reUploadPart", zap.String("oid", id.String()), zap.Uint64("payload size", obj.PayloadSize()))
```
roman-khimov, line 668: Debug.

internal/neofs/tree.go
```diff
@@ -284,15 +284,17 @@ func newPartInfo(node NodeResponse) (*data.PartInfo, error) {
 		case homoHashKV:
 			partInfo.HomoHash = []byte(value)
 		case elementsKV:
-			elements := strings.Split(value, ",")
-			partInfo.Elements = make([]data.LinkObjectPayload, len(elements))
-			for i, e := range elements {
-				var element data.LinkObjectPayload
-				if err = element.Unmarshal(e); err != nil {
-					return nil, fmt.Errorf("invalid element: %w", err)
+			if value != "" {
```
roman-khimov, line 287: Seems unrelated.

api/layer/multipart_upload.go
```diff
@@ -1287,26 +1325,13 @@ func (n *layer) uploadPartAsSlot(ctx context.Context, params uploadPartAsSlotPar
...
-	}
-
-	id, elements, err = n.manualSlice(ctx, params.bktInfo, prm, splitFirstID, splitPreviousID, *chunk, params.payloadReader)
-	if isReturnToPool {
-		n.buffers.Put(chunk)
+		Payload:      params.payloadReader,
+		PayloadSize:  uint64(params.decSize),
 	}
 
+	id, objHashBts, err := n.objectPutAndHash(ctx, prm, params.bktInfo)
```
roman-khimov, line 1332: So, no links to first ID?

api/layer/multipart_upload.go
```diff
@@ -595,14 +595,67 @@ func (n *layer) reUploadFollowingParts(ctx context.Context, uploadParams UploadP
...
+func (n *layer) reUploadSegmentedPart(ctx context.Context, uploadParams UploadPartParams, part *data.PartInfo, bktInfo *data.BucketInfo, multipartInfo *data.MultipartInfo) error {
+	completePartPayload := make([]byte, 0, n.neoFS.MaxObjectSize())
+
+	for _, element := range part.Elements {
+		elementObj, err := n.objectGet(ctx, bktInfo, element.OID)
+		if err != nil {
+			return fmt.Errorf("get part oid=%s, emelent oid=%s: %w", part.OID.String(), element.OID.String(), err)
+		}
+
+		completePartPayload = append(completePartPayload, elementObj.Payload()...)
```
roman-khimov, line 621: 10G object?

smallhive: This is a huge problem because we have to know the whole part size before uploading. I am thinking about trying to add some extra fields in the meta, to get this info it

api/layer/multipart_upload.go
smallhive: Updated

api/layer/multipart_upload.go
smallhive: No, we upload Slot just as a regular object

internal/neofs/tree.go
smallhive: Panics otherwise, in case of the slot the part doesn't contain elements

api/layer/multipart_upload.go
smallhive: UPD: Replaced with io.Pipe. It should be less memory aggressive

api/layer/multipart_upload.go
roman-khimov: Expected to a part of the first commit.

api/layer/multipart_upload.go
```diff
@@ -595,9 +596,83 @@ func (n *layer) reUploadFollowingParts(ctx context.Context, uploadParams UploadP
...
+
+func (n *layer) reUploadSegmentedPart(ctx context.Context, uploadParams UploadPartParams, part *data.PartInfo, bktInfo *data.BucketInfo, multipartInfo *data.MultipartInfo) error {
+	var (
+		eg                     errgroup.Group
+		pipeReader, pipeWriter = io.Pipe()
+	)
+
+	eg.Go(func() error {
+		for _, element := range part.Elements {
+			elementObj, err := n.objectGet(ctx, bktInfo, element.OID)
```
roman-khimov, line 621: I think it's overcomplicated. If you have a proper object, you may not care about its individual elements any more, you can just issue a GET for it, let node assemble individual parts, get an io.Reader and do io.Copy from it into larger object io.Writer.

smallhive: It is a key. If the part has elements, we slice it manually. This means we don't have a proper object and have to do all these actions

api/layer/multipart_upload.go
smallhive: Fixed

api/layer/multipart_upload.go
```diff
@@ -589,14 +596,100 @@ func (n *layer) reUploadFollowingParts(ctx context.Context, uploadParams UploadP
...
+	eg.Go(func() error {
+		var (
+			err        error
+			elementObj *object.Object
+		)
+
+		for _, element := range part.Elements {
+			elementObj, err = n.objectGet(ctx, bktInfo, element.OID)
+			if err != nil {
+				err = fmt.Errorf("get part oid=%s, emelent oid=%s: %w", part.OID.String(), element.OID.String(), err)
```
roman-khimov, line 628: emelent?

api/layer/multipart_upload.go
```diff
@@ -589,14 +596,100 @@ func (n *layer) reUploadFollowingParts(ctx context.Context, uploadParams UploadP
...
+				// no return intentionally.
+			}
+		}
+
+		if err != nil {
+			pipeCloseErr := pipeWriter.Close()
+			return fmt.Errorf("pipe: %w. PipeClose: %w", err, pipeCloseErr)
+		}
+
+		if err = pipeWriter.Close(); err != nil {
```
roman-khimov, line 662: You can always just `pipeCloseErr := pipeWriter.Close()` and then deal with errors.

api/layer/multipart_upload.go
smallhive: fixed

api/layer/multipart_upload.go
smallhive: I hope I understood in the right way

api/layer/multipart_upload.go
```diff
@@ -589,9 +596,90 @@ func (n *layer) reUploadFollowingParts(ctx context.Context, uploadParams UploadP
...
+					zap.String("uploadID", multipartInfo.UploadID),
+					zap.Int("partNumber", part.Number),
+					zap.String("part.OID", part.OID.String()),
+					zap.String("part element OID", element.OID.String()),
+				)
+				// no return intentionally.
+			}
+		}
+
+		if pipeCloseErr := pipeWriter.Close(); pipeCloseErr != nil {
```
roman-khimov, line 657: So if `pipeCloseErr == nil` and `err != nil` we're fine? I want to deduplicate `Close()`, but errors should be passed properly still.

smallhive: Updated

roman-khimov (approved):

Issue #902, opened 2023-11-09
GetObject: can't get object under not owner user even when correct ACL were set


evgeniiz321:
test_object_copy_not_owned_object_bucket

Create bucket
Put object
Put object acl to allow access from a different user
Put bucket acl to allow access from a different user
Try to get object - access denied

```
2023-11-09T04:10:17.937Z	info	api/router.go:165	call method	{"status": 200, "host": "s3.neofs.devenv:8080", "request_id": "43e678ea-a540-4d91-9fc8-fb0db759ac1b", "method": "GetObjectACL", "bucket": "yournamehere-j11jteqzoakk8cxr-1", "object": "foo123bar", "description": "OK"}
2023-11-09T04:10:17.945Z	debug	layer/layer.go:480	get object	{"reqId": "87ec59d6-253d-444f-a9e6-d3a159bb17a4", "bucket": "yournamehere-j11jteqzoakk8cxr-1", "cid": "B4P5iBZ981MvCc3vjuyyiVXZM8HAg4wLzq2MCpPiuTwC", "object": "foo123bar", "oid": "6MPjzkbWEM8SJbK4uU7wz5kQRymMUtTa2mGs3Erf1YPq"}
2023-11-09T04:10:18.964Z	info	api/router.go:165	call method	{"status": 200, "host": "s3.neofs.devenv:8080", "request_id": "87ec59d6-253d-444f-a9e6-d3a159bb17a4", "method": "PutObjectACL", "bucket": "yournamehere-j11jteqzoakk8cxr-1", "object": "foo123bar", "description": "OK"}
2023-11-09T04:10:18.982Z	info	api/router.go:165	call method	{"status": 200, "host": "s3.neofs.devenv:8080", "request_id": "64f2cb68-5aa5-425a-8e85-98234a54599a", "method": "GetBucketACL", "bucket": "yournamehere-j11jteqzoakk8cxr-1", "object": "", "description": "OK"}
2023-11-09T04:10:20.005Z	info	api/router.go:165	call method	{"status": 200, "host": "s3.neofs.devenv:8080", "request_id": "e789e26a-9aa7-4eff-a429-e5b327871712", "method": "PutBucketACL", "bucket": "yournamehere-j11jteqzoakk8cxr-1", "object": "", "description": "OK"}
[...]

PR #1050, opened 2024-12-27
*: Remove tree service for object operations

smallhive:
Refs #902

### DIFF
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -688,34 +688,6 @@ func isErrObjectAlreadyRemoved(err error) bool {
 	}
 }
 
-func dismissNotFoundError(obj *VersionedObject) *VersionedObject {
-	if s3errors.IsS3Error(obj.Error, s3errors.ErrNoSuchKey) ||
-		s3errors.IsS3Error(obj.Error, s3errors.ErrNoSuchVersion) {
-		obj.Error = nil
-	}
-
-	return obj
-}
-
-func (n *layer) getNodeVersionToDelete(ctx context.Context, bkt *data.BucketInfo, obj *VersionedObject) (*data.NodeVersion, error) {
-	objVersion := &ObjectVersion{
-		BktInfo:               bkt,
-		ObjectName:            obj.Name,
-		VersionID:             obj.VersionID,
-		NoErrorOnDeleteMarker: true,
-	}
-
-	return n.getNodeVersion(ctx, objVersion)
-}
-
-func (n *layer) removeOldVersion(ctx context.Context, bkt *data.BucketInfo, nodeVersion *data.NodeVersion, obj *VersionedObject) (string, error) {
-	if nodeVersion.IsDeleteMarker() {
-		return obj.VersionID, nil
-	}
-
-	return "", n.objectDelete(ctx, bkt, nodeVersion.OID)
-}
-
 // DeleteObjects from the storage.
 func (n *layer) DeleteObjects(ctx context.Context, p *DeleteObjectParams) []*VersionedObject {
 	for i, obj := range p.Objects {
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -1,8 +1,8 @@
 package layer
 
 import (
+	"bytes"
 	"context"
-	"crypto/rand"
 	"errors"
 	"fmt"
 	"io"
@@ -19,9 +19,11 @@ import (
 	"github.com/nspcc-dev/neofs-s3-gw/api/layer/encryption"
 	"github.com/nspcc-dev/neofs-s3-gw/api/s3errors"
 	"github.com/nspcc-dev/neofs-s3-gw/creds/accessbox"
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/eacl"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/session"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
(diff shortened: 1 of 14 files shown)


api/layer/layer.go
```diff
@@ -556,81 +566,128 @@ func (n *layer) CopyObject(ctx context.Context, p *CopyObjectParams) (*data.Exte
...
+func isErrObjectAlreadyRemoved(err error) bool {
+	var (
+		ol  apistatus.ObjectAlreadyRemoved
+		olp *apistatus.ObjectAlreadyRemoved
+	)
+	switch {
+	case errors.As(err, &ol), errors.As(err, &olp):
+		return true
+	default:
+		return strings.Contains(err.Error(), "object already removed")
```
roman-khimov, line 687: Why `errors.As` isn't enough?

api/layer/object.go
```diff
@@ -243,7 +249,13 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
 		CopiesNumber: p.CopiesNumber,
 	}
 
-	prm.Attributes = make([][2]string, 0, len(p.Header))
+	prm.Attributes = make([][2]string, 0, len(p.Header)+1)
+
+	if bktSettings.Unversioned() || bktSettings.VersioningSuspended() {
+		prm.Attributes = append(prm.Attributes, [2]string{attrS3VersioningState, data.VersioningUnversioned})
+	} else if bktSettings.VersioningEnabled() {
+		prm.Attributes = append(prm.Attributes, [2]string{attrS3VersioningState, data.VersioningEnabled})
```
roman-khimov, line 255: Why for each object?

smallhive: I need to understand whether an object's version is versioned. I could try to include this information only in versioned objects and skip this attribute for unversioned objects (I mean if bucket settings set to versioned/unversioned state). This may require some changes in logic and filters, but I consider it possible.

Do you have another suggestion how to store this info?

api/layer/object.go
```diff
@@ -262,7 +272,7 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
 	n.log.Debug("put object",
 		zap.String("reqId", reqInfo.RequestID),
 		zap.String("bucket", p.BktInfo.Name), zap.Stringer("cid", p.BktInfo.CID),
-		zap.String("object", p.Object), zap.Stringer("oid", id))
+		zap.String("object", p.Object), zap.Stringer("oid", id), zap.Int64("size", p.Size))
 
 	newVersion.OID = id
 	newVersion.ETag = hex.EncodeToString(hash)
```
roman-khimov, line 278: Ah, looks like you still have `treeService.AddVersion` for the object below. This explains epoch-bound degradation. Can we drop it?

api/layer/object.go
roman-khimov: 1. We need compatibility with existing data, it doesn't have any attributes like this.
2. If we can avoid additional data in header we better avoid it.

smallhive: Updated, I left the attribute only for the versioned state

api/layer/layer.go
```diff
@@ -556,81 +566,128 @@ func (n *layer) CopyObject(ctx context.Context, p *CopyObjectParams) (*data.Exte
...
+						return obj
+					}
+				}
+			} else {
+				if err := deleteOID.DecodeString(obj.VersionID); err != nil {
+					obj.Error = fmt.Errorf("decode version: %w", err)
+					return obj
+				}
+
+				if obj.Error = n.objectDelete(ctx, bkt, deleteOID); obj.Error != nil {
```
roman-khimov, line 601: The answer seems to be here. And tombstones are to be deleted after expiration anyway. Maybe not a problem.

api/layer/layer.go
```diff
@@ -556,81 +566,128 @@ func (n *layer) CopyObject(ctx context.Context, p *CopyObjectParams) (*data.Exte
...
-		}
-
-		if obj.DeleteMarkVersion, obj.Error = n.removeOldVersion(ctx, bkt, nodeVersion, obj); obj.Error != nil {
-			if strings.Contains(obj.Error.Error(), "2050 message = object is locked") {
-				return obj
+	if settings.VersioningEnabled() {
+		if len(obj.VersionID) > 0 {
+			var deleteOID oid.ID
+
+			if obj.VersionID == data.UnversionedObjectVersionID {
```
roman-khimov, line 574: In which cases is this true?

api/layer/versioning.go
```diff
@@ -31,8 +35,18 @@ func (n *layer) ListObjectVersions(ctx context.Context, p *ListObjectVersionsPar
...
+		var isNullVersionCounted bool
+
 		for i, version := range sortedVersions {
 			version.IsLatest = i == 0
+			if version.NodeVersion.IsUnversioned && isNullVersionCounted {
+				continue
+			}
+
+			if version.NodeVersion.IsUnversioned {
+				isNullVersionCounted = true
```
roman-khimov, line 48: Break here?

api/layer/versioning.go
```diff
@@ -31,8 +35,18 @@ func (n *layer) ListObjectVersions(ctx context.Context, p *ListObjectVersionsPar
 			return cmp.Compare(b.NodeVersion.Timestamp, a.NodeVersion.Timestamp) // sort in reverse order
 		})
 
+		// The object with "null" version should be only one. We get only last (actual) one.
+		var isNullVersionCounted bool
+
 		for i, version := range sortedVersions {
 			version.IsLatest = i == 0
+			if version.NodeVersion.IsUnversioned && isNullVersionCounted {
```
roman-khimov, line 43: Can you have both versioned and unversioned?

api/layer/object.go
```diff
@@ -373,68 +386,249 @@ func (n *layer) headLastVersionIfNotDeleted(ctx context.Context, bkt *data.Bucke
...
+		if err != nil {
+			n.log.Warn("couldn't head object",
+				zap.Stringer("oid", &ids[i]),
+				zap.Stringer("cid", bkt.CID),
+				zap.Error(err))
+
+			return nil, fmt.Errorf("couldn't head object: %w", err)
+		}
+
+		// if head.Type() == object.TypeTombstone || head.Type() == object.TypeLink || head.Type() == object.TypeLock {
```
roman-khimov, line 484: Why this commented out code?

api/layer/versioning.go
smallhive: Yes

api/layer/versioning.go
smallhive: No, we should squash null versions in one, but separate versions should be still here

api/layer/object.go
smallhive: I use object-type filter instead, commend is removed

api/layer/layer.go
smallhive: When client sends version "null"

roman-khimov (approved):

PR #1058, opened 2025-01-22
Deduplicate object attributes

### LINKED ISSUE #1054: CopyObject failure
## Current Behavior
```
2025-01-14 09:30:25	error	handler/util.go:34	call method	{"status": 500, "request_id": "3758078d-f6f4-4490-b8d7-4b7695acef40", "method": "CopyObject", "bucket": "gitlab-uploads", "object": "@hashed/4e/07/4e07408562bedb8b60ce05c1decfe3ad16b72230967de01f640b7e4729b49fce/81eb2ddd3dcab6dfe8f2a542b09d71f0/Screenshot.png", "description": "couldn't copy object", "error": "copy payload with buffer: status: code = 1024 message = duplication of attributes detected", "src_bucket_name": "gitlab-uploads", "src_object_name": "tmp/uploads/1736847024-2141369-0011-5046-14d5234f1c86c947ddde71cf610ef682"}
```

## Expected Behavior
CopyObject doing its job.

## Possible Solution
Fix it?

## Steps to Reproduce
Currently it's reproduced in Gitlab on upload and I'm not sure it's 100% reproducible. It uploads to a temporary object and then does `CopyObject` that fails.

## Regression
Not likely.

## Your Environment
* Version of the product used: 0.34.0
* Operating System and version (`uname -a`): Debian bookworm

smallhive:
Closes #1054.

### DIFF
--- a/api/layer/object.go
+++ b/api/layer/object.go
@@ -233,6 +233,12 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
 		}
 	}
 
+	for _, v := range p.Header {
+		if v == "" {
+			return nil, ErrMetaEmptyParameterValue
+		}
+	}
+
 	prm := PrmObjectCreate{
 		Container:    p.BktInfo.CID,
 		Creator:      owner,
@@ -241,16 +247,7 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
 		Payload:      r,
 		CreationTime: TimeNow(ctx),
 		CopiesNumber: p.CopiesNumber,
-	}
-
-	prm.Attributes = make([][2]string, 0, len(p.Header))
-
-	for k, v := range p.Header {
-		if v == "" {
-			return nil, ErrMetaEmptyParameterValue
-		}
-
-		prm.Attributes = append(prm.Attributes, [2]string{k, v})
+		Attributes:   p.Header,
 	}
 
 	id, hash, err := n.objectPutAndHash(ctx, prm, p.BktInfo)
--- a/internal/neofs/neofs.go
+++ b/internal/neofs/neofs.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
+	"maps"
 	"math"
 	"strconv"
 	"sync"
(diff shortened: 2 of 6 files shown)


internal/neofs/neofs.go
```diff
@@ -748,3 +750,21 @@ func (x *NeoFS) SearchObjects(ctx context.Context, prm layer.PrmObjectSearch) ([
 
 	return oids, nil
 }
+
+func deduplicateAttributes(attrs []object.Attribute) []object.Attribute {
+	var (
+		uniq   = make(map[string]string, len(attrs))
+		result = make([]object.Attribute, 0, len(attrs))
```
roman-khimov, line 757: You can do this in-place without additional memory overhead, just sort and drop consecutive identical keys.

internal/neofs/neofs.go
```diff
@@ -296,10 +296,12 @@ func (x *NeoFS) CreateObject(ctx context.Context, prm layer.PrmObjectCreate) (oi
 	objectNonceAttr := object.NewAttribute(objectNonceAttribute, base64.StdEncoding.EncodeToString(nonce))
 	attrs = append(attrs, *objectNonceAttr)
 
+	deduplicated := deduplicateAttributes(attrs)
```
roman-khimov, line 299: Likely you're just better not creating duplicates in the first place. We're inheriting all of the attributes of the original object, likely the original `prm.Attributes` set is clean, we just need to deal with overrides here (either old or new one wins in case it's present already) according to the S3 conventions.

roman-khimov (commented):
And maybe `CreateObject()` is better to accept a map (it's always there a layer above).

internal/neofs/neofs.go
smallhive: It is not correct. At this point, we have only user attributes and don't have attributes from CreateObject. At a minimum, I failed to reproduce the duplication error with a copy. I assume there is a user attributes duplication, but from the code it is not clear

smallhive:
> And maybe `CreateObject()` is better to accept a map (it's always there a layer above).

This option I like better, reworked

internal/neofs/neofs.go
smallhive: Replaced with map

api/layer/object.go
```diff
@@ -243,14 +243,14 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
...
-	prm.Attributes = make([][2]string, 0, len(p.Header))
+	prm.Attributes = make(map[string]string, len(p.Header))
 
 	for k, v := range p.Header {
 		if v == "" {
 			return nil, ErrMetaEmptyParameterValue
 		}
 
-		prm.Attributes = append(prm.Attributes, [2]string{k, v})
+		prm.Attributes[k] = v
```
roman-khimov, line 253: Copying from map to map? Do we need this copy?

internal/neofs/neofs.go
```diff
@@ -262,39 +263,29 @@ func (x *NeoFS) signMultipartObject(obj *object.Object, signer neofscrypto.Signe
...
 
-	for i := range prm.Attributes {
-		a = object.NewAttribute(prm.Attributes[i][0], prm.Attributes[i][1])
-		attrs = append(attrs, *a)
+	nonce := make([]byte, objectNonceSize)
+	if _, err := rand.Read(nonce); err != nil {
+		return oid.ID{}, fmt.Errorf("object nonce: %w", err)
 	}
 
+	uniqAttributes := maps.Clone(prm.Attributes)
```
roman-khimov, line 276: Do you need to copy it?

api/layer/object.go
smallhive: Well, I think not. We can only checkthe  original map for empty values and pass p.Header to params

internal/neofs/neofs.go
smallhive: No, but the idea was to save original parameters unchanged, to avoid any side effects

roman-khimov (approved):

PR #1061, opened 2025-01-24
Fix eacl issue in copyobject

### LINKED ISSUE #1055: eACL issue in CopyObject
## Current Behavior
```
error	handler/util.go:34	call method	{"status": 500, "request_id": "a91733e0-ca93-411e-a0d0-cd0693f9769b", "method": "CopyObject", "bucket": "gitlab-uploads", "object": "@hashed/4e/07/4e07408562bedb8b60ce05c1decfe3ad16b72230967de01f640b7e4729b49fce/7eb8fb8657a8254241a1c6559a2b2ac2/dejavu.jpeg", "description": "could not put bucket acl", "error": "save eACL via connection pool: container setEacl: status: code = 1024 message = Invalid params (-32602) - byte-slice is too big (68047)"}
```

## Expected Behavior
I thought we're done with ACLs after #1016 and associated fixes.

## Possible Solution
Fix the damn thing. No EACLs should be touched at all.

## Steps to Reproduce
Upload a picture via GL.

## Regression
No.

## Your Environment
* Version of the product used: 0.34.0
* Operating System and version (`uname -a`): Debian bookworm

smallhive:
Closes #1055

### DIFF
--- a/api/handler/copy.go
+++ b/api/handler/copy.go
@@ -97,7 +97,7 @@ func (h *handler) CopyObjectHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if containsACL {
-		if isBucketOwnerForced(eacl.EACL) {
+		if IsBucketOwnerForced(eacl.EACL) {
 			if !isValidOwnerEnforced(r) {
 				h.logAndSendError(w, "access control list not supported", reqInfo, s3errors.GetAPIError(s3errors.ErrAccessControlListNotSupported))
 				return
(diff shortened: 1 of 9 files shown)


api/handler/copy.go
```diff
@@ -225,7 +225,8 @@ func (h *handler) CopyObjectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if containsACL {
+	// check all headers one more time.
+	if containsACLHeaders(r) {
```
roman-khimov, line 229: But why?

smallhive: Upper on the code, we clean headers in some situations. In this case, var `containsACL` may contain invalid value

roman-khimov: This should be written somewhere in the comment.

smallhive: Updated comments

roman-khimov (approved):

PR #1066, opened 2025-01-29
Dont use tree for tags

smallhive:


api/handler/put.go
```diff
@@ -318,10 +318,16 @@ func (h *handler) PutObjectHandler(w http.ResponseWriter, r *http.Request) {
...
+			NodeVersion:  extendedObjInfo.NodeVersion,
+			CopiesNumber: h.cfg.CopiesNumber,
 		}
-		if _, err = h.obj.PutObjectTagging(r.Context(), tagPrm); err != nil {
+
+		if !settings.VersioningEnabled() {
+			tagPrm.ObjectVersion.VersionID = ""
+		}
+
+		if err = h.obj.PutObjectTagging(r.Context(), tagPrm); err != nil {
```
roman-khimov, line 330: So, these are tags included into the object put request? Do we need a separate object in this case?

api/layer/tagging.go
```diff
@@ -38,74 +83,100 @@ func (n *layer) GetObjectTagging(ctx context.Context, p *GetObjectTaggingParams)
...
 
-	n.cache.PutTagging(n.Owner(ctx), objectTaggingCacheKey(p.ObjectVersion), p.TagSet)
+	slices.SortFunc(objects, sortObjectsFunc)
+
+	var (
+		lastObj = objects[0]
+		tags    = make(map[string]string)
+	)
+
+	if err = json.Unmarshal(lastObj.Payload(), &tags); err != nil {
```
roman-khimov, line 134: What if I add two tags in the first request and then add three more in the second? I would expect to receive five tags then.

api/layer/neofs_mock.go
```diff
@@ -491,40 +491,42 @@ func (t *TestNeoFS) SearchObjects(_ context.Context, prm PrmObjectSearch) ([]oid
...
+func checkFilters(obj *object.Object, filters object.SearchFilters) bool {
+	var isOk = true
+
+	attrs := make(map[string]string, len(obj.Attributes()))
+	for _, attr := range obj.Attributes() {
+		attrs[attr.Key()] = attr.Value()
+	}
+
+	for _, f := range filters {
+		val, preset := attrs[f.Header()]
```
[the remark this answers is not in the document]
smallhive: Correct

api/handler/copy.go
```diff
@@ -78,6 +78,27 @@ func (h *handler) CopyObjectHandler(w http.ResponseWriter, r *http.Request) {
...
+			Object:  reqInfo.ObjectName,
+		}
+
+		ei, err := h.obj.GetExtendedObjectInfo(r.Context(), headObjectPrm)
+		if err != nil {
+			h.logAndSendError(w, "could not find object", reqInfo, err)
+			return
+		}
+
+		srcObjPrm.VersionID = ei.ObjectInfo.VersionID()
```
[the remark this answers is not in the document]
smallhive: Finding correct version of source object and it's tags

api/handler/put.go
smallhive: An object is unchangeable, tags are changeable at any time. Storing tags inside an object doesn't give any advantages, but adds more routine to check actual tags state

api/layer/neofs_mock.go
```diff
@@ -491,40 +491,42 @@ func (t *TestNeoFS) SearchObjects(_ context.Context, prm PrmObjectSearch) ([]oid
...
 		}
 	}
 
 	return oids, nil
 }
+
+func checkFilters(obj *object.Object, filters object.SearchFilters) bool {
+	var isOk = true
+
+	attrs := make(map[string]string, len(obj.Attributes()))
```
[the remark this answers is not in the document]
smallhive: This change allows us to check field presence correctly. Comment is updated

api/layer/tagging.go
smallhive: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html#tagging-apis

[PUT Object tagging](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUTtagging.html) – Replaces tags on an object ...

The freshest object contains actual set of tags

roman-khimov (approved):

PR #1078, opened 2025-02-11
New search

cthulhu-rider:


cthulhu-rider:
@smallhive pls pay attentions to this [check](https://github.com/nspcc-dev/neofs-s3-gw/blob/8f1c0fb730bb664e5c2de40c3ce0a08b0dbcce78/api/layer/object.go#L491): u'll need `ROOT` filter for this, but it's not supported yet https://github.com/nspcc-dev/neofs-node/issues/3116. Im also surprised that chunky objects can get into this place, because the request is pretty specific in attributes, i dont expect "small" objects to have them

roman-khimov:
We need to ensure multipart/chunks have no attributes that can match for real objects and then this filter could be deleted completely.

PR #1082, opened 2025-02-19
Integrate search v2

### LINKED ISSUE #1081: Integrate search v2
The node has a new great feature - the search v2. We should integrate it into the gate as fast as we can

smallhive:
Closes #1081

### DIFF
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -65,6 +65,17 @@ type (
 		Headers        map[string]string
 	}
 
+	// ObjectListResponseContent holds response data for object listing.
+	ObjectListResponseContent struct {
+		ID      oid.ID
+		IsDir   bool
+		Size    int64
+		Owner   user.ID
+		HashSum string
+		Created time.Time
+		Name    string
+	}
+
 	// NotificationInfo store info to send s3 notification.
 	NotificationInfo struct {
 		Name    string
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -150,7 +150,6 @@ type (
 		PartNumber int64
 		// in nanoseconds
 		CreatedAt int64
-		FilePath  string
 	}
 
 	uploadPartAsSlotParams struct {
@@ -1433,8 +1432,6 @@ func (n *layer) getSlotAttributes(obj object.Object) (*slotAttributes, error) {
 			attributes.PartNumber, err = strconv.ParseInt(attr.Value(), 10, 64)
 		case headerS3MultipartCreated:
 			attributes.CreatedAt, err = strconv.ParseInt(attr.Value(), 10, 64)
-		case object.AttributeFilePath:
-			attributes.FilePath = attr.Value()
 		default:
 			continue
 		}
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -960,7 +960,6 @@ func (n *layer) CompleteMultipartUpload(ctx context.Context, p *CompleteMultipar
 	prm := PrmObjectCreate{
 		Container:    p.Info.Bkt.CID,
 		Creator:      p.Info.Bkt.Owner,
-		Filepath:     p.Info.Key,
 		CreationTime: TimeNow(ctx),
 		CopiesNumber: multipartInfo.CopiesNumber,
 		Multipart: &Multipart{
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -22,6 +22,7 @@ import (
 	"github.com/nspcc-dev/neofs-s3-gw/api/data"
 	"github.com/nspcc-dev/neofs-s3-gw/api/layer/encryption"
 	"github.com/nspcc-dev/neofs-s3-gw/api/s3errors"
+	"github.com/nspcc-dev/neofs-sdk-go/client"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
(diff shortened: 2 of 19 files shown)


roman-khimov:
Replaces #1078?

api/layer/multipart_upload.go
```diff
@@ -1416,52 +1417,33 @@ func (n *layer) uploadPartAsSlot(ctx context.Context, params uploadPartAsSlotPar
 
 func (n *layer) getFirstArbitraryPart(ctx context.Context, uploadID string, bucketInfo *data.BucketInfo) (int64, error) {
 	var filters object.SearchFilters
+	filters.AddFilter(headerS3MultipartNumber, "0", object.MatchNumGE) // only primary attribute is sorted
```
roman-khimov, line 1420: For search to be efficient you want the first filter to be the most limiting. There can be _a lot_ of multiparts in the process of uploading and this will make node loop through all parts of all of them to find something that matches. While multipart ID itself can have a lot of parts, I'd rather bet on it as having less items to walk through. And you can do any kind of secondary sorting after receiving the result.

roman-khimov: > a lot of parts

BTW, limited to 10000. Practically way less than that usually.

internal/neofs/neofs.go
```diff
@@ -752,3 +752,28 @@ func (x *NeoFS) SearchObjects(ctx context.Context, prm layer.PrmObjectSearch) ([
...
+		err         error
+	)
+
+	for {
+		items, cursor, err = x.pool.SearchObjects(ctx, cid, filters, attributes, cursor, x.signer(ctx), opts)
+		if err != nil {
+			return nil, fmt.Errorf("search objects: %w", err)
+		}
+
+		resultItems = append(resultItems, items...)
```
roman-khimov, line 771: This defeats the purpose of searchv2 for object listings. In some cases you need all items, in some you don't. You can have different methods exposed here for these cases.

api/data/info.go
```diff
@@ -65,6 +65,18 @@ type (
 		Headers        map[string]string
 	}
 
+	// ObjectListResponseContent holds response data for object listing.
+	ObjectListResponseContent struct {
+		ID            oid.ID
+		IsDir         bool
+		DecryptedSize int64
```
roman-khimov, line 72: Not sure what's the difference with `Size`.

api/data/info.go
```diff
@@ -65,6 +65,18 @@ type (
 		Headers        map[string]string
 	}
 
+	// ObjectListResponseContent holds response data for object listing.
+	ObjectListResponseContent struct {
```
roman-khimov, line 69: What are the uses for `ObjectInfo`? Can this be embedded into `ObjectInfo`? Do we need both or just one is sufficient?

api/layer/object.go
```diff
@@ -435,25 +447,114 @@ func (n *layer) searchAllVersionsInNeoFS(ctx context.Context, bkt *data.BucketIn
...
+	)
+
+	if bt := bearerTokenFromContext(ctx, owner); bt != nil {
+		opts.WithBearerToken(*bt)
 	}
 
-	n.prepareAuthParameters(ctx, &prmSearch.PrmAuth, owner)
-	prmSearch.Filters.AddTypeFilter(object.MatchStringEqual, object.TypeRegular)
-	prmSearch.Filters.AddFilter(attributeTagsMetaObject, "", object.MatchNotPresent)
+	filters.AddFilter(object.FilterCreationEpoch, "0", object.MatchNumGT)
```
roman-khimov, line 573: Should be a prefix first of all, the most limiting one and coming from the user request directly.

api/layer/object.go
```diff
@@ -435,25 +447,114 @@ func (n *layer) searchAllVersionsInNeoFS(ctx context.Context, bkt *data.BucketIn
...
+		filters.AddFilter(object.AttributeFilePath, prefix, object.MatchCommonPrefix)
 	}
 
 	if onlyUnversioned {
-		prmSearch.Filters.AddFilter(attrS3VersioningState, data.VersioningUnversioned, object.MatchNotPresent)
+		filters.AddFilter(attrS3VersioningState, data.VersioningUnversioned, object.MatchNotPresent)
 	}
 
-	return n.searchObjects(ctx, bkt, prmSearch)
+	searchResultItems, err := n.neoFS.SearchObjectsV2(ctx, bkt.CID, filters, returningAttributes, opts)
```
roman-khimov, line 585: You don't need all items and you do need continuation token for listings.

api/data/info.go
smallhive: `ObjectInfo` in general appears with `GetExtendedObjectInfo` method to get full info about object. Moreover, it contains _all_ object attributes, which are used for responses and other data manipulations

`ObjectListResponseContent` is a shorter version, which is enough to show a bucket listing

api/layer/object.go
smallhive: Why?

api/data/info.go
smallhive: It should be. It is calculated and stored in a separate attribute when encryption is enabled

smallhive: At least one test agreed with me:
```
s3tests_boto3/functional/test_s3.py::test_encryption_sse_c_multipart_bad_download - assert 31472640 == 31457280
```

roman-khimov: I know, but there is a layering difference I fear we're breaking. You need to get this data from the DB, but you don't need this data in the output, it's just `Size` there. `api/data` is closer to the output to me, but maybe it's not.

api/layer/object.go
roman-khimov: Listing is exactly this: get 1000 items and a continuation token, return both to the user, request served OK, thanks.

api/data/info.go
smallhive: Do you mean I should hide this `DecryptedSize` in a request layer? Then rewrite size _in place_ and return to the output only one _Size_ field?

roman-khimov: Looks like that. From the user POV it's just an object size.

api/layer/multipart_upload.go
smallhive: Fixed in the node, updated in the gate

internal/neofs/neofs.go
smallhive: Reworked

api/data/info.go
smallhive: Removed from this layer, left it only on lower one

api/layer/object.go
smallhive: Fixed in the node, updated in the gate

api/layer/object.go
```diff
@@ -466,125 +488,208 @@
...
-		}
+	if bt := bearerTokenFromContext(ctx, owner); bt != nil {
+		opts.WithBearerToken(*bt)
 	}
-	for _, attr := range b.Attributes() {
-		if attr.Key() == object.AttributeTimestamp {
-			bCreated, _ = strconv.ParseInt(attr.Value(), 10, 64)
-			break
+
+	opts.SetCount(uint32(maxKeys))
```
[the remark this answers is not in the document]
smallhive: Updated

api/layer/object.go
```diff
@@ -435,25 +449,120 @@ func (n *layer) searchAllVersionsInNeoFS(ctx context.Context, bkt *data.BucketIn
...
 
-	if len(prefix) > 0 {
-		prmSearch.Filters.AddFilter(object.AttributeFilePath, prefix, object.MatchCommonPrefix)
+	if maxKeys <= 0 || maxKeys > 1000 {
+		maxKeys = 1000
 	}
 
+	opts.SetCount(uint32(maxKeys))
+
+	filters.AddFilter(object.AttributeFilePath, prefix, object.MatchCommonPrefix)
```
roman-khimov, line 580: Not sure how prefix here...

api/layer/object.go
```diff
@@ -435,25 +449,120 @@ func (n *layer) searchAllVersionsInNeoFS(ctx context.Context, bkt *data.BucketIn
...
+		if psr.DecryptedSize > 0 {
+			psr.PayloadSize = psr.DecryptedSize
+		}
+
+		psr.PayloadChecksum = item.Attributes[6]
+
+		searchResults = append(searchResults, psr)
+	}
+
+	sortFunc := func(a, b prefixSearchResult) int {
```
roman-khimov, line 652: ...works with this sorting function. It's perfectly fine for objects with the same `FilePath`, but otherwise they're just different objects as per S3. Consider prefix "a" and "ab", "ac" objects.

api/layer/object.go
```diff
@@ -432,25 +447,107 @@ func (n *layer) headLastVersionIfNotDeleted(ctx context.Context, bkt *data.Bucke
...
+
+	if bt := bearerTokenFromContext(ctx, owner); bt != nil {
+		opts.WithBearerToken(*bt)
+	}
 
 	if len(objectName) > 0 {
-		prmSearch.Filters.AddFilter(object.AttributeFilePath, objectName, object.MatchStringEqual)
+		filters.AddFilter(object.AttributeFilePath, objectName, object.MatchStringEqual)
+	} else {
+		filters.AddFilter(object.AttributeFilePath, "", object.MatchCommonPrefix)
```
roman-khimov, line 470: Can this ever really happen?

roman-khimov (approved):

api/layer/object.go
smallhive: At least one place exists in code, where we call with an empty object name - on the bucket removing. Any other place with user input may be involved as well

api/layer/object.go
smallhive: This method is used only for bucket listing. The prefix helps to filter similar objects or just list a "dir". There may be many objects with the same `FilePath`.
This sorting allows us to find the actual version of the object. But, maybe for this exact place, such multi-level sorting is too much. The question requires separate checks

roman-khimov: I'd expect this sorting to be prefix first and time second. Right now IIUC in a/ab/ac situation you can get _any_ order even though technically everyone expects exactly "a, ab, ac". Then _if_ there are multiple versions of "a", those should be sorted according to time stamps.

api/layer/object.go
roman-khimov: Bucket removal retrieves all objects from bucket?

smallhive: You can't remove a bucket with objects. In out case with regular objects

PR #1085, opened 2025-02-27
Streamline getobject

smallhive:


api/layer/layer.go
```diff
@@ -618,6 +655,137 @@ func (n *layer) GetExtendedObjectInfo(ctx context.Context, p *HeadObjectParams)
...
+	// if err != nil {
+	// 	return nil, nil, nil, fmt.Errorf("get head failed: %w", err)
+	// }
+	//
+	// objInfo := objectInfoFromMeta(p.BktInfo, meta)
+	//
+	// extObjInfo := &data.ExtendedObjectInfo{
+	// 	ObjectInfo:  objInfo,
+	// 	NodeVersion: &data.NodeVersion{},
+	// }
```
roman-khimov, line 772: Drop this.

api/handler/get.go
```diff
@@ -180,46 +192,52 @@ func (h *handler) GetObjectHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	getParams := &layer.GetObjectParams{
-		ObjectInfo: info,
-		Writer:     w,
-		Range:      params,
-		BucketInfo: bktInfo,
-		Encryption: encryptionParams,
+	if params != nil || encryptionParams.Enabled() {
+		// unfortunately this reader is useless for us in this case, we have to re-read another one.
```
roman-khimov, line 212: It can be reused for sure (just piping into the decryption layer), but that's not our current primary target.

api/handler/get.go
```diff
@@ -180,46 +192,52 @@ func (h *handler) GetObjectHandler(w http.ResponseWriter, r *http.Request) {
...
+		if err = h.obj.GetObject(r.Context(), getParams); err != nil {
+			h.logAndSendError(w, "could not get object", reqInfo, err)
+		}
+
+		return
 	}
-	if err = h.obj.GetObject(r.Context(), getParams); err != nil {
-		h.logAndSendError(w, "could not get object", reqInfo, err)
+
+	var bufferSize = min(info.Size, 1024*1024)
```
roman-khimov, line 229: Where this constant comes from?

smallhive: The old realization uses 32kb https://github.com/nspcc-dev/neofs-s3-gw/blob/30f920cb3324a4547292ea46343fa441c854aaba/api/layer/layer.go#L459, I just picked up a slightly bigger value. 

Do you think about different size?

api/layer/layer.go
smallhive: Ouch, removed

api/handler/get.go
roman-khimov: Let's do 128K. 1M is somewhat large wrt DOS scenario.

smallhive: Updated

Issue #996, opened 2024-09-03
Drop github.com/urfave/cli/v2 dependency


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we're using a lot of similar libraries, especially if we're using two different ones in the same project. S3 gateway itself doesn't use `github.com/urfave/cli/v2`, but authmate for whatever reason does.

## Describe the solution you'd like
Unify with S3 (and even REST) gateway, drop `github.com/urfave/cli/v2` dependency.

## Describe alternatives you've considered
Leave as is.

smallhive:
After discussions with @roman-khimov, we decided not to drop github.com/urfave/cli/v2 dependency in the gate. It allows Authmate to work as a regular CLI application and have separate well-structured commands.

PR #1095, opened 2025-03-11
urfavecliv2 dependency

smallhive:
After discussions with @roman-khimov, we decided not to drop `github.com/urfave/cli/v2` dependency in the gate. It allows Authmate to work as a regular CLI application and have separate well-structured commands.

This PR only update project dependencies


roman-khimov:
> go.mod: Upgrade github.com/urfave/cli/v2 v2.27.5 => v2.27.6
> Close #996.

It's not correct, it's OK to reference #996, but it needs to be closed as "not planned", not "completed".

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/nspcc-dev/neofs-s3-gw
 
-go 1.23
+go 1.23.0
+
+toolchain go1.23.1
```
roman-khimov, line 5: No.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/nspcc-dev/neofs-s3-gw
 
-go 1.23
+go 1.23.0
```
roman-khimov, line 3: No `.0` please.

go.mod
smallhive: `go mod` makes these changes when you upgrade golang.org/x/crypto

roman-khimov (approved):

go.mod
roman-khimov: I know, I know, Go modules are not that fun now. But I hope we'll drop x/crypto one day (#1080 + #1096)

PR #1097, opened 2025-03-11
Refactor

smallhive:
Some refactoring changes and removals after the last big updates

### DIFF
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -252,7 +252,7 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 		payloadReader = p.Reader
 		decSize       = p.Size
 		attributes    = map[string]string{
-			s3headers.ObjectKey: multipartInfo.Key,
+			s3headers.MultipartObjectKey: multipartInfo.Key,
 		}
 	)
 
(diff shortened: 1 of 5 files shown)


api/layer/multipart_upload.go
```diff
@@ -1050,7 +1049,6 @@ func (n *layer) multipartGetPartsList(ctx context.Context, bktInfo *data.BucketI
 			s3headers.ElementID,
 			s3headers.TotalSize,
 			AttributeDecryptedSize,
-			object.AttributeTimestamp,
```
roman-khimov, line 1070: What's the result if we're uploading the same part concurrently?

api/s3headers/headers.go
```diff
@@ -6,6 +6,25 @@ const (
...
+	// MultipartHash contains hash.Hash state to calculate final object hash.
+	MultipartHash = "s3mpHash"
+	// MultipartHomoHash contains hash.Hash state to calculate final object homo hash.
+	MultipartHomoHash = "s3mpHomoHash"
+	// MultipartPartHash contains hash for MultipartPartNumber.
+	MultipartPartHash = "s3mpPartHash"
+	// MultipartOwner contains object owner for uploading object.
+	MultipartOwner = "s3mpOwner"
+	// MultipartCopiesNumber contains CopiesNumber setting for uploading object.
+	MultipartCopiesNumber = "s3mpCopiesNumber"
```
roman-khimov, line 30: Why do you save it in attributes?

smallhive: It is an attribute for the MultipartInfo object. It stores only once

roman-khimov: Does it affect future UploadPart requests?

smallhive: Yes

roman-khimov (approved):

PR #1101, opened 2025-03-12
Listen new_epoch notifications

### LINKED ISSUE #1094: Subscribe to epoch changes in S3
In some cases, we need to havean  actual epoch

Listen to the epoch change event from the node

## Additional context
We already have cache for this, but we should improve the schema

smallhive:
Closes #1094.


internal/neofs/epoch.go
```diff
@@ -2,72 +2,159 @@ package neofs
...
+			select {
+			case <-ctx.Done():
+				l.log.Info("stopping. Context done")
+				return
+			default:
+			}
+
+			l.notifyChan = make(chan *state.ContainedNotificationEvent)
+
+			id, err := l.client.ReceiveExecutionNotifications(&neorpc.NotificationFilter{Contract: &l.netMapContract}, l.notifyChan)
```
roman-khimov, line 124: Add a filter for `NewEpoch` name here...

internal/neofs/epoch.go
```diff
@@ -2,72 +2,159 @@ package neofs
...
+}
+
+func (l *EpochListener) readNotifications() {
+	for {
+		notification, ok := <-l.notifyChan
+		if !ok {
+			break
+		}
+
+		if notification.Name != "NewEpoch" {
```
roman-khimov, line 147: ...to avoid checking it here.

internal/neofs/epoch.go
```diff
@@ -2,72 +2,159 @@ package neofs
...
+	}
+
+	return nil
+}
+
+// ListenNotifications start listening notifications. Reconnects in case of disconnect.
+func (l *EpochListener) ListenNotifications(ctx context.Context) {
+	go func() {
+		for {
+			if err := l.connect(ctx); err != nil {
```
roman-khimov, line 109: Current epoch must be updated after reconnection.

config/config.yaml
```diff
@@ -144,8 +144,6 @@ neofs:
   # Number of the object copies to consider PUT to NeoFS successful.
   # `0` means that object will be processed according to the container's placement policy
   set_copies_number: 0
-  # Timeout between retrieving actual epoch from NeoFS. Actual only if slicer.enabled = true.
-  epoch_update_interval: 2m
```
roman-khimov, line 148: Add a CHANGELOG entry to drop this from configurations.

roman-khimov:
Does it solve the test problem?

smallhive:
Yes, it solves the `test_multipart_upload_overwrite_existing_object` problem. Test ran 50 times:
* master: 16 failed, 34 passed
* branch: 50 passed

internal/neofs/epoch.go
smallhive: Updated

config/config.yaml
smallhive: Added

internal/neofs/epoch.go
```diff
@@ -2,72 +2,185 @@ package neofs
...
+				l.log.Error("connection failed", zap.Error(err))
+				time.Sleep(1 * time.Second)
+				continue
+			}
+
+			if err := l.forceGetEpoch(); err != nil {
```
roman-khimov, line 132: You can miss an event between this line...

internal/neofs/epoch.go
```diff
@@ -2,72 +2,185 @@ package neofs
...
+			select {
+			case <-ctx.Done():
+				l.log.Info("stopping. Context done")
+				return
+			default:
+			}
+
+			l.notifyChan = make(chan *state.ContainedNotificationEvent)
+
+			id, err := l.client.ReceiveExecutionNotifications(&neorpc.NotificationFilter{Contract: &l.netMapContract, Name: &eventName}, l.notifyChan)
```
roman-khimov, line 148: ... and this one (but just swapping them would lead to potential deadlock).

internal/neofs/epoch.go
smallhive: Reworked

internal/neofs/epoch.go
smallhive: Reworked

cmd/s3-gw/app.go
```diff
@@ -114,20 +119,48 @@ func newApp(ctx context.Context, log *Logger, v *viper.Viper) *App {
...
 	}
 
-	// If slicer is disabled, we should use "static" getter, which doesn't make periodic requests to the NeoFS.
-	var epochGetter neofs.EpochGetter = ni
+	var (
+		rpcHTTPEndpoints = v.GetStringSlice(cfgRPCEndpoints)
+		wsEndpoints      = make([]string, len(rpcHTTPEndpoints))
+		opt              rpcclient.Options
+		netMapContract   util.Uint160
+		emptyAddr        util.Uint160
```
roman-khimov, line 127: You're not using it as var.

cmd/s3-gw/app.go
```diff
@@ -114,20 +119,48 @@ func newApp(ctx context.Context, log *Logger, v *viper.Viper) *App {
...
+			log.logger.Warn("couldn't create inferred reader", zap.Error(err), zap.String("enpoint", endpoint))
+			continue
 		}
 
-		epochGetter = neofs.NewPeriodicGetter(ctx, ni.CurrentEpoch(), epochUpdateInterval, conns, log.logger)
+		netMapContract, err = nnsReader.ResolveFSContract(rpcNNS.NameNetmap)
+		if err != nil {
+			log.logger.Warn("couldn't resolve netmap contract", zap.Error(err), zap.String("endpoint", endpoint))
+			continue
+		}
```
roman-khimov, line 156: Close the client after use, don't leak resources.

internal/neofs/epoch.go
```diff
@@ -2,72 +2,163 @@ package neofs
...
 // CurrentEpoch returns actual epoch.
 //
 // CurrentEpoch implements [EpochGetter].
-func (g *PeriodicGetter) CurrentEpoch() uint64 {
-	return g.epoch.Load()
+func (l *EpochListener) CurrentEpoch() uint64 {
+	return l.epoch.Load()
+}
+
+func (l *EpochListener) Init(netMapContract util.Uint160) {
```
roman-khimov, line 55: Not sure why this is a separate method, can be a part of `New()` to me.

internal/neofs/epoch.go
```diff
@@ -2,72 +2,163 @@ package neofs
...
+			select {
+			case <-ctx.Done():
+				l.log.Info("stopping. Context done")
+				return
+			default:
+			}
+
+			if err := l.connect(ctx); err != nil {
+				l.log.Error("connection failed", zap.Error(err))
+				time.Sleep(1 * time.Second)
```
roman-khimov, line 116: You can use `github.com/cenkalti/backoff/v4` instead.

internal/neofs/epoch.go
```diff
@@ -2,72 +2,163 @@ package neofs
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				l.log.Info("stopping. Context done")
+				return
+			default:
+			}
+
+			if err := l.connect(ctx); err != nil {
```
roman-khimov, line 114: There is still a race between this connect...

internal/neofs/epoch.go
roman-khimov: ... and this subscription, you can lose an event.

cmd/s3-gw/app.go
smallhive: Updated

internal/neofs/epoch.go
smallhive: It makes sense, updated

internal/neofs/epoch.go
smallhive: Used

internal/neofs/epoch.go
smallhive: Moved `forceGet` right after subsciption

internal/neofs/epoch.go
```diff
@@ -2,72 +2,168 @@ package neofs
...
+			if err != nil {
+				l.log.Info("receive execution notifications failed", zap.Error(err))
+
+				_ = l.client.Unsubscribe(id)
+				l.client.Close()
+				l.client = nil
+				continue
+			}
+
+			if err = l.forceGetEpoch(); err != nil {
```
roman-khimov, line 128: Now the problem is that you're not reading notifications while doing this request which can lead to a deadlock.

internal/neofs/epoch.go
```diff
@@ -2,72 +2,168 @@ package neofs
...
+			select {
+			case <-ctx.Done():
+				l.log.Info("stopping. Context done")
+				return
+			default:
+			}
+
+			if err := l.connect(ctx); err != nil {
+				l.log.Error("connection failed", zap.Error(err))
+				time.Sleep(bo.NextBackOff())
```
roman-khimov, line 112: You need to reset it after successful connection. Otherwise it will grow even though there are no issues with establishing a new connection.

internal/neofs/epoch.go
smallhive: Hm, what order do you suggest?

roman-khimov: That's one option, the other is doing it in a separate routine.

smallhive: Updated

PR #1103, opened 2025-03-14
1076 versioning objects code simplification

### LINKED ISSUE #1076: Versioning objects code simplification
No, it is a polishing issue

After tree-service removal, there are many places to handle versioning, and I consider they can be simplified

Tree-service version assumed each object has a "version" in the tree. Some code have to live with this fact, I want to try to simplify this

smallhive:
Closes #1076


api/layer/layer.go
```diff
@@ -655,62 +655,45 @@ func (n *layer) GetExtendedObjectInfo(ctx context.Context, p *HeadObjectParams)
...
+		lockInfo       *data.LockInfo
+		isEmptyVersion = len(p.VersionID) == 0
+		isNullVersion  = p.VersionID == data.UnversionedObjectVersionID
 	)
 
-	if len(p.VersionID) == 0 {
-		versions, tags, locks, err = n.comprehensiveSearchAllVersionsInNeoFS(ctx, p.BktInfo, owner, p.Object, false)
-		if err != nil {
-			if errors.Is(err, ErrNodeNotFound) {
+	versions, tags, locks, err = n.comprehensiveSearchAllVersionsInNeoFS(ctx, p.BktInfo, owner, p.Object, isNullVersion)
```
roman-khimov, line 672: Why making any search at all if you have a particular version already? tags/locks? Likely it can be optimized in this case (you know the OID).

api/layer/neofs.go
```diff
@@ -301,3 +312,26 @@ type NeoFS interface {
 	// SearchObjectsV2WithCursor searches objects with corresponding filters and return objectID with requested attributes. It uses cursor to start from required point.
 	SearchObjectsV2WithCursor(ctx context.Context, cid cid.ID, filters object.SearchFilters, attributes []string, cursor string, opts client.SearchObjectsOptions) ([]client.SearchResultItem, string, error)
 }
+
+// WritePayload writes bts to each available hash.
+func (h *MultipartHashes) WritePayload(bts []byte) error {
+	if h.Hash != nil {
```
roman-khimov, line 318: Can be simplified with `for _, h := range []hash.Hash{h.Hash, h.HomoHash, h.PartHash}`

smallhive: Done

api/layer/layer.go
smallhive: Updated, but another search for tags/locks still exists

PR #1104, opened 2025-03-17
Remove tree-service usage for bucket tags

### LINKED ISSUE #1072: Tree-less bucket tagging
Same as #1068, but for bucket tags.

smallhive:
Closes #1072.

### DIFF
--- a/api/handler/tagging.go
+++ b/api/handler/tagging.go
@@ -219,7 +219,7 @@ func (h *handler) PutBucketTaggingHandler(w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	if err = h.obj.PutBucketTagging(r.Context(), bktInfo, tagSet); err != nil {
+	if err = h.obj.PutBucketTagging(r.Context(), bktInfo, tagSet, h.cfg.CopiesNumber); err != nil {
 		h.logAndSendError(w, "could not put object tagging", reqInfo, err)
 		return
 	}
--- a/api/layer/tagging.go
+++ b/api/layer/tagging.go
@@ -160,7 +160,7 @@ func (n *layer) GetObjectTagging(ctx context.Context, p *GetObjectTaggingParams)
 
 	slices.SortFunc(searchResults, sortFunc)
 
-	tags, err := n.getObjectTagsByOID(ctx, p.ObjectVersion.BktInfo, searchResults[0].ID)
+	tags, err := n.getTagsByOID(ctx, p.ObjectVersion.BktInfo, searchResults[0].ID)
 	if err != nil {
 		return "", nil, err
 	}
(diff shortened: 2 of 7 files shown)


api/handler/tagging.go
```diff
@@ -219,7 +219,7 @@ func (h *handler) PutBucketTaggingHandler(w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	if err = h.obj.PutBucketTagging(r.Context(), bktInfo, tagSet); err != nil {
+	if err = h.obj.PutBucketTagging(r.Context(), bktInfo, tagSet, h.cfg.CopiesNumber); err != nil {
```
roman-khimov, line 222: Where this (copies number) setting comes from?

api/layer/tagging.go
```diff
@@ -228,33 +228,118 @@ func (n *layer) DeleteObjectTagging(ctx context.Context, p *ObjectVersion) error
...
+	)
 
-	if tags := n.cache.GetTagging(owner, bucketTaggingCacheKey(bktInfo.CID)); tags != nil {
-		return tags, nil
+	if bt := bearerTokenFromContext(ctx, owner); bt != nil {
+		opts.WithBearerToken(*bt)
 	}
 
-	tags, err := n.treeService.GetBucketTagging(ctx, bktInfo)
+	filters.AddFilter(object.AttributeFilePath, bktInfo.Name, object.MatchStringEqual)
```
roman-khimov, line 251: Why using `FilePath` for these objects at all?

api/s3headers/headers.go
```diff
@@ -38,8 +38,9 @@ const (
 	// MetaType is a header name to identify meta containers for objects.
 	MetaType = "s3MetaType"
 
-	TypeLock = "lock"
-	TypeTags = "tags"
+	TypeLock       = "lock"
+	TypeTags       = "tags"
+	TypeBucketTags = "buketTags"
```
roman-khimov, line 43: bucket

api/handler/tagging.go
smallhive: This is the default value from the gate config

api/s3headers/headers.go
smallhive: Ouch, fixed

api/handler/tagging.go
smallhive: For regular objects, we take this value from the request headers. Otherwise, take this default value

api/layer/tagging.go
smallhive: Do you mean the bucket already is a restriction? Thus we don't need to put a bucket name to an object inside the same bucket

roman-khimov: 1. You can (and should) use a different attribute, this simplifies filtering in other cases, this is not a proper S3 (user-uploaded) object, those have a `FilePath`, this one doesn't need it.
2. Putting bucket name is not very helpful, it's known.

It can be like `S3BucketMetadata=tags` and then `S3BucketMetadata=notifications`, etc.

api/layer/tagging.go
```diff
@@ -228,33 +228,115 @@ func (n *layer) DeleteObjectTagging(ctx context.Context, p *ObjectVersion) error
 }
 
 func (n *layer) GetBucketTagging(ctx context.Context, bktInfo *data.BucketInfo) (map[string]string, error) {
-	owner := n.Owner(ctx)
+	var (
+		err   error
+		owner = n.Owner(ctx)
+	)
+
+	var (
```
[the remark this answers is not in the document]
smallhive: Merged

roman-khimov (approved):

PR #1106, opened 2025-03-18
Remove tree-service usage for bucket notifications config

### LINKED ISSUE #1071: Tree-less notification configuration
Same as #1068, but for notification configurations.

smallhive:
Closes #1071.

### DIFF
--- a/api/layer/notifications.go
+++ b/api/layer/notifications.go
@@ -4,12 +4,11 @@ import (
 	"bytes"
 	"context"
 	"encoding/xml"
-	"errors"
 	"fmt"
 
 	"github.com/nspcc-dev/neofs-s3-gw/api"
 	"github.com/nspcc-dev/neofs-s3-gw/api/data"
-	"go.uber.org/zap"
+	"github.com/nspcc-dev/neofs-s3-gw/api/s3headers"
 )
 
 type PutBucketNotificationConfigurationParams struct {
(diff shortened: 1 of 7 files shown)


api/layer/notifications.go
```diff
@@ -19,6 +27,12 @@ type PutBucketNotificationConfigurationParams struct {
 	CopiesNumber  uint32
 }
 
+type bucketNotifiConfigSearchResult struct {
```
roman-khimov, line 30: I think we have the same set in many places, can be unified.

api/layer/notifications.go
```diff
@@ -29,62 +43,109 @@ func (n *layer) PutBucketNotificationConfiguration(ctx context.Context, p *PutBu
...
 	}
 
-	objID, err := n.treeService.GetNotificationConfigurationNode(ctx, bktInfo)
-	objIDNotFound := errors.Is(err, ErrNodeNotFound)
-	if err != nil && !objIDNotFound {
-		return nil, err
+	filters.AddFilter(s3headers.MetaType, s3headers.TypeBucketNotifConfig, object.MatchStringEqual)
+	filters.AddTypeFilter(object.MatchStringEqual, object.TypeRegular)
+
+	searchResultItems, err := n.neoFS.SearchObjectsV2(ctx, bktInfo.CID, filters, returningAttributes, opts)
```
roman-khimov, line 83: This search can be generalized as well, the only thing differing is the first attribute value.

api/layer/notifications.go
```diff
@@ -29,62 +43,109 @@ func (n *layer) PutBucketNotificationConfiguration(ctx context.Context, p *PutBu
...
+			psr.CreationTimestamp, err = strconv.ParseInt(item.Attributes[2], 10, 64)
+			if err != nil {
+				return nil, fmt.Errorf("invalid creation timestamp %s: %w", item.Attributes[2], err)
+			}
+		}
+
+		searchResults = append(searchResults, psr)
+	}
+
+	sortFunc := func(a, b bucketNotifiConfigSearchResult) int {
```
roman-khimov, line 124: This one as well.

roman-khimov (commented):
DRY

smallhive:
I agree with all the lines. We have 3 almost identical PRs with similar changes. I am considering merging them and unifying inside another small PR. It would be faster and all checks can be done just ones

api/layer/notifications.go
smallhive: Unified

api/layer/notifications.go
smallhive: Unified

api/layer/notifications.go
smallhive: Unified

PR #1107, opened 2025-03-18
Remove tree-service usage for bucket settings

### LINKED ISSUE #1070: Tree-less bucket settings
Same as #1068, but for bucket-level settings.

smallhive:
Closes #1070.


smallhive:
Rebased, updated and tree completely removed @roman-khimov

api/layer/object.go
```diff
@@ -1232,13 +1232,15 @@ func (n *layer) searchBucketMetaObjects(ctx context.Context, containerID cid.ID,
 	)
 
 	if bt := bearerTokenFromContext(ctx, owner); bt != nil {
-		opts.WithBearerToken(*bt)
+		if owner == bktInfo.Owner || !bktInfo.BacisAcl.IsOpAllowed(acl.OpObjectSearch, acl.RoleOthers) {
```
roman-khimov, line 1235: If basic ACL has this disabled bearer can't help you at all.

api/data/locking.go
```diff
@@ -48,3 +51,62 @@ type (
 		ByPassedGovernance bool
 	}
 )
+
+// Encode marshal ObjectLockConfiguration to string.
+func (conf *ObjectLockConfiguration) Encode() string {
```
roman-khimov, line 56: Can be https://pkg.go.dev/encoding#TextMarshaler then. It's not very popular, but it exists.

api/data/locking.go
```diff
@@ -48,3 +51,62 @@ type (
...
+	return fmt.Sprintf("%s,%d,%s,%d", conf.ObjectLockEnabled, defaults.Days, defaults.Mode, defaults.Years)
+}
+
+// Decode unmarshal ObjectLockConfiguration from string.
+func (conf *ObjectLockConfiguration) Decode(value string) error {
+	if len(value) == 0 {
+		return nil
+	}
+
+	lockValues := strings.Split(value, ",")
```
roman-khimov, line 71: Better use https://pkg.go.dev/strings#SplitSeq then, it's safer.

api/layer/system_object.go
```diff
@@ -279,22 +280,68 @@ func (n *layer) GetBucketSettings(ctx context.Context, bktInfo *data.BucketInfo)
...
+	id, err := n.searchBucketMetaObjects(ctx, bktInfo.CID, s3headers.TypeBucketSettings)
+	if err != nil {
+		return nil, fmt.Errorf("search: %w", err)
+	}
+
+	if id.IsZero() {
+		return &settings, nil
+	}
+
+	settigsObj, err := n.objectGet(ctx, bktInfo, id)
```
roman-khimov, line 297: settigs?

api/data/info.go
```diff
@@ -34,6 +35,7 @@ type (
 		Created            time.Time
 		LocationConstraint string
 		ObjectLockEnabled  bool
+		BacisAcl           acl.Basic
```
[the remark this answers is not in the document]
smallhive: Correct, but this structure is used everywhere as BucketInfo, inside the cache as well.

api/data/locking.go
smallhive: added in 1.24.0, we still have 1.23

api/layer/object.go
roman-khimov: And the most correct check would be `bt.Owner == bktInfo.Owner`, btw. If they're not matched everything else is irrelevant.

api/layer/system_object.go
smallhive: Updated

api/layer/object.go
smallhive: We don't have such cases in S3 right now. All buckets have EACL enabled. Anyway, this problem is out of the scope of this task.
Maybe we should think of creating _separate_ bucket for all bucket settings to be able to get this data without any restrictions for the gate or user. Another option is using contract to store this data, to avoid this restriction

smallhive:
The builds are fixed. There were some _tree_ in workflows

api/data/locking.go
roman-khimov: OK!

api/layer/object.go
roman-khimov: `bt.Issuer`, of course. Just check of Issuer==Owner, that's it, you don't need anything else right now.

roman-khimov: If you're to do that you don't need any basic ACL checks as well.

api/layer/system_object.go
smallhive: Oops, push failed. Fixed

api/layer/object.go
smallhive: Updated

api/data/info.go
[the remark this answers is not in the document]
smallhive: Removed

roman-khimov (approved):

PR #1110, opened 2025-03-24
Fix tags and locks search

smallhive:


roman-khimov (commented):
Please explain (in the commit message).

roman-khimov (approved):

PR #1113, opened 2025-04-02
Migrate to aws sdk v2

smallhive:
Tested with the latest versions:
```
boto3                     1.37.24
botocore                  1.37.24
```
```
aws-cli/2.25.3
```


api/auth/center.go
```diff
@@ -108,7 +115,7 @@ func New(neoFS tokens.NeoFS, key *keys.PrivateKey, prefixes []string, config *ca
 	}
 }
 
-func (c *center) parseAuthHeader(header string) (*authHeader, error) {
+func (c *center) parseAuthHeader(header, amzContent string) (*authHeader, error) {
```
roman-khimov, line 120: amzContentHash? It's a bit unobvious here.

api/auth/center.go
```diff
@@ -161,7 +169,13 @@ func (c *center) Authenticate(r *http.Request) (*Box, error) {
 			SignedFields: queryValues[AmzSignedHeaders],
 			Date:         creds[1],
 			IsPresigned:  true,
+			PayloadHash:  r.Header.Get(AmzContentSha256),
+		}
+
+		if authHdr.PayloadHash == "" {
+			authHdr.PayloadHash = UnsignedPayload
 		}
+
```
roman-khimov, line 180: Useless change.

api/auth/center.go
```diff
@@ -220,8 +234,13 @@ func (c *center) Authenticate(r *http.Request) (*Box, error) {
 			return nil, fmt.Errorf("DecodeString: %w", err)
 		}
 
-		awsCreds := credentials.NewStaticCredentials(authHdr.AccessKeyID, box.Gate.AccessKey, "")
-		streamSigner := v4.NewChunkSigner(authHdr.Region, authHdr.Service, sig, signatureDateTime, awsCreds)
+		appCreds := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(authHdr.AccessKeyID, box.Gate.AccessKey, ""))
```
roman-khimov: What's the purpose of this cache, can it be reused?

api/auth/center.go
smallhive: Renamed var

api/auth/center.go
smallhive: Removed

api/auth/center.go
smallhive: The [docs](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-gosdk.html) says
```
This enables caching and credential rotation that is concurrency safe.
```

For our purposes, it is useless, because we check these credentials on each request for each user. It would work well for a client application.

Removed cache from code

roman-khimov (approved):

PR #1121, opened 2025-04-09
Check cache usage

smallhive:


api/layer/system_object.go
```diff
@@ -291,6 +290,9 @@ func (n *layer) GetBucketSettings(ctx context.Context, bktInfo *data.BucketInfo)
 	}
 
 	if id.IsZero() {
+		settings.LockConfiguration = &data.ObjectLockConfiguration{}
```
roman-khimov, line 293: Suggests non-pointer `LockConfiguration` to me.

smallhive: Some code checks `!= nil`, in this case the configuration is empty. Just left it empty by default

roman-khimov (approved):

roman-khimov (approved):

PR #1127, opened 2025-04-17
Add request processing time histograms in metrics

### LINKED ISSUE #1099: Request processing time histograms in metrics
I'm always frustrated when I don't see request processing time histograms in metrics. NeoGo provides them, NeoFS node provides them, S3 gateway should provide them.

Provide histograms for all requests in and out. Per API. And pool too. It gives both counters and time perspective which is much more valuable than counters and average requests.

smallhive:
Closes #1099


api/metrics/collector.go
```diff
@@ -23,7 +27,7 @@ var (
 	)
 
 	statsMetrics = &stats{
-		desc: prometheus.NewDesc("neofs_s3_stats", "Statistics exposed by NeoFS S3 Gate instance", nil, nil),
+		desc: prometheus.NewDesc(namespace+"stats", "Statistics exposed by NeoFS S3 Gate instance", nil, nil),
```
roman-khimov, line 30: `+"_stats"` to be compatible.

api/metrics/pool_metrics.go
```diff
@@ -0,0 +1,314 @@
...
+			Name:      "object_hash",
+			Help:      "Object hash request handling time",
+		}),
+		ContainerAnnounceUsedSpace: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: poolSubsystem,
+			Name:      "container_announce_used_space",
+			Help:      "Container announce used space request handling time",
+		}),
+		AnnounceIntermediateTrust: prometheus.NewHistogram(prometheus.HistogramOpts{
```
roman-khimov, line 165: It never calls `Announce*`, btw.

roman-khimov (commented):
How about S3 api statistics?

api/metrics/collector.go
smallhive: Ouch, updated

api/metrics/pool_metrics.go
smallhive: Do you want to remove it? What about `ContainerAnnounceUsedSpace` and `AnnounceLocalTrust`?

smallhive:
> How about S3 api statistics?

Already [covered](https://github.com/nspcc-dev/neofs-s3-gw/blob/master/api/metrics/api.go)

api/metrics/pool_metrics.go
roman-khimov: I think so, S3 will never use them, they're all SN-specific.

smallhive: Removed

roman-khimov (approved):

Issue #1130, opened 2025-04-21
Unstable TestObjectsListCache/lifetime


roman-khimov:
## Current Behavior
https://github.com/nspcc-dev/neofs-s3-gw/actions/runs/14570740667/job/40867449345?pr=1128
```
    objectslist_test.go:50: 
        	Error Trace:	/home/runner/work/neofs-s3-gw/neofs-s3-gw/api/cache/objectslist_test.go:50
        	Error:      	Condition satisfied
        	Test:       	TestObjectsListCache/lifetime
```

## Expected Behavior
Green.

## Possible Solution
Fix.

## Steps to Reproduce
Unknown, but it can happen on GH.

PR #1132, opened 2025-04-22
Stabilize test

smallhive:
Sometimes, the runtime took a bit more time, and data was wiped out from the cache. Therefore, sometimes we get nil from the condition. We reduced the condition check for 1 second to ensure we had no nil result for the Never check. Simultaneously, we increased Eventually check interval. This condition checks that the cache is going to be empty during these two seconds. In one second and a few milliseconds in fact.


roman-khimov (commented):
1. Commit message formatting is weird
2. #1130 isn't mentioned.
3. Looks like the test is inherently racy, but OK.

smallhive:
Updated

roman-khimov (approved):

PR #1136, opened 2025-04-24
accessbox: use AES instead of ChaCha20-Poly1305

### LINKED ISSUE #1096: Drop chacha20poly1305 dependency
I'm always frustrated when we have dependencies we don't really need. chacha20poly1305 was introduced in #80 along with ed25519 crypto. Then the key scheme was changed to 256r1 since that's what we have in Neo, NeoFS, etc. But chacha20poly1305 remained.

There is nothing inherently wrong with it, but at the same time it's an additional dependency that can be easily avoided. We've got https://pkg.go.dev/crypto/cipher@go1.24.1#AEAD that works fine over standard AES (and handles random nonce internally as well).

Keep things as is, import and maintain x/crypto dependency that also pulls x/net that always has some bugs @dependabot wants to fix.

## Context
#1080, the other thing we're using from x/crypto is hkdf, but it's a part of the standard Go 1.24.

roman-khimov:
Chacha is fun, but it's not something provided by standard libraries. AES works fine as well. Fixes #1096.

### DIFF
--- a/creds/accessbox/accessbox.go
+++ b/creds/accessbox/accessbox.go
@@ -2,6 +2,7 @@ package accessbox
 
 import (
 	"bytes"
+	"crypto/aes"
 	"crypto/cipher"
 	"crypto/elliptic"
 	"crypto/rand"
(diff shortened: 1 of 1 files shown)


smallhive (approved):

roman-khimov:
This will sit in the queue for while (till 0.37.0) because merging it immediately can break tests.

PR #1150, opened 2025-05-05
Get metadata consistency policy from contract

### LINKED ISSUE #1148: Accept metadata consistency policy from contract callback
I'm always frustrated when I can't configure metadata consistency policy per-container.  #1116 was OK for quick tests, but doesn't work well for production.

Accept the following structure from `resolvePolicy(userAddr interop.Hash160, policyName string) []byte`:
```
{
    "version": 0, // Versioning
    "placement": {}, // netmap-compatible policy object as was returned previously
    "consistency": "strict" // or "optimistic" or no value, optional field
}
```
Keep compatiblity! Unmarshal as a simple netmap policy first and accept with the default consistency setting it it's OK.

Take consistency setting from this structure otherwise and create container with it.

smallhive:
Closes #1148.

### DIFF
--- a/api/handler/put.go
+++ b/api/handler/put.go
@@ -825,20 +825,32 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (h handler) setPolicy(prm *layer.CreateBucketParams, userAddr util.Uint160, locationConstraint string, userPolicies []*accessbox.ContainerPolicy) {
-	prm.Policy = h.cfg.Policy.Default()
+	prm.Policy = layer.PlacementPolicy{
+		Version:     layer.PlacementPolicyV1,
+		Placement:   h.cfg.Policy.Default(),
+		Consistency: h.cfg.ContainerMetadataPolicy,
+	}
 
 	if locationConstraint == "" {
 		return
 	}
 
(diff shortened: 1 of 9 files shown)


api/handler/put.go
```diff
@@ -825,20 +825,32 @@ func (h *handler) CreateBucketHandler(w http.ResponseWriter, r *http.Request) {
...
 	if locationConstraint == "" {
 		return
 	}
 
 	if policy, ok := h.cfg.Policy.Get(locationConstraint); ok {
-		prm.Policy = policy
+		prm.Policy = layer.PlacementPolicy{
+			Version:     layer.PlacementPolicyV1,
+			Placement:   policy,
+			Consistency: h.cfg.ContainerMetadataPolicy,
```
roman-khimov, line 842: Seems like you're overriding policy value here which is not correct. You've already handled defaults in `Get()`, so if it's `ok` then it's just `prm.Policy = policy`.

smallhive: `get` returns `netmap.PlacementPolicy`. But `prm.Policy` now is a new `layer.PlacementPolicy`. We reads data from contract in the end of the function

roman-khimov: OK, I see. Looks like we have many choices.

roman-khimov (approved):

PR #1162, opened 2025-05-06
Attribute documentation fixes

roman-khimov:


docs/attribute_mapping.md
```diff
@@ -279,15 +288,16 @@ Stores settings configuration for a bucket.
...
-It canbe empty string or `%s,%d,%s,%d` string.
-
-* `string`. Is object lock enabled. Possible values: `` (empty string), `Enabled`.
-* `int64`. Retention days for bucket objects.
-* `string`. Retention mode. Possible values: `GOVERNANCE`, `COMPLIANCE`.
-* `int64`. Retention years for bucket objects.
-
-`Days` and `Years` represent durations and cannot be used simultaneously. Only one of them can be specified at a time.
+  Deprecated, replaced by meta version.
+- `S3-BucketSettings-MetaVersion`
```
smallhive, line 292: I'm not sure it is deprecated. This field was planned to be used to understand which version of the configuration we work with. We don't have any alternative right now

roman-khimov: It's not deprecated. It's `S3-BucketSettings-Versioning` that is deprecated.

Issue #1158, opened 2025-05-06
Tag objects are not deleted when deleting an object


roman-khimov:
## Current Behavior
Object is deleted, but tag object for it is left.

## Expected Behavior
No objects left behind.

## Possible Solution
Find and drop appropriate tag object. Do not make additional searches.

## Steps to Reproduce
Put an object. Put tags for it. Delete object. Check container contents.

PR #1166, opened 2025-05-13
Store object tags as attributes along with a payload

### LINKED ISSUE #1153: Extend tag storage scheme
I'm always frustrated when I see out storage scheme is not future-proof. S3 tags are not very useful on their own, setting/getting them is not the real purpose. They become useful in various filters, mostly lifecycle, but also policy, replication and notifications. We support none of these functions, but for them to be possible we need tag object scheme to be flexible enough already.

1. Store tags in `S3-Meta-Tag-XXX: YYY` attributes along with JSON-based body when using separate tag objects. This makes it possible to do tag-based searches.
2. Do not create additional objects when tags are provided along with the object itself (store in attributes only). This optimizes out tag retrieval in most cases, tags are more likely to be provided at creation stage rather than separately.

Keeping things as is.

## Additional context
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/tagging-and-policies.html

smallhive:
Closes #1153.

It would help to make tag based searches in the future.

### DIFF
--- a/api/layer/tagging.go
+++ b/api/layer/tagging.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"slices"
 	"strconv"
+	"strings"
 
 	"github.com/nspcc-dev/neofs-s3-gw/api/data"
 	"github.com/nspcc-dev/neofs-s3-gw/api/s3errors"
@@ -35,6 +36,7 @@ type taggingSearchResult struct {
 	ID                oid.ID
 	FilePath          string
 	CreationTimestamp int64
+	IsOriginalObject  bool
 }
 
 func (n *layer) PutObjectTagging(ctx context.Context, p *PutObjectTaggingParams) error {
@@ -88,6 +90,7 @@ func (n *layer) GetObjectTagging(ctx context.Context, p *GetObjectTaggingParams)
 		returningAttributes = []string{
 			object.AttributeFilePath,
 			object.AttributeTimestamp,
+			s3headers.MetaType,
 		}
 
 		opts client.SearchObjectsOptions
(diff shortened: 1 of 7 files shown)


api/layer/tagging.go
```diff
@@ -158,6 +160,55 @@ func (n *layer) GetObjectTagging(ctx context.Context, p *GetObjectTaggingParams)
...
+func (n *layer) getTagsFromOriginalObject(ctx context.Context, bktInfo *data.BucketInfo, name, version string) (map[string]string, error) {
+	var (
+		id             oid.ID
+		isEmptyVersion = len(version) == 0
+		isNullVersion  = version == data.UnversionedObjectVersionID
+		tagSet         = make(map[string]string)
+	)
+
+	if isEmptyVersion || isNullVersion {
+		versions, _, _, err := n.comprehensiveSearchAllVersionsInNeoFS(ctx, bktInfo, n.Owner(ctx), name, isNullVersion)
```
roman-khimov, line 172: Additional searches can and should be avoided.

roman-khimov (commented):
How about tag deletion?

api/layer/tagging.go
smallhive: Do you suggest searching all object versions + all meta objects for this version and making a filter in the code? Is it cheaper than two searches?

smallhive:
> How about tag deletion?

It is a responsibility of #1158

roman-khimov:
I mean deleting tags of an object that had them set on put.

api/layer/tagging.go
smallhive: Reworked

smallhive:
I consider you mean [this](https://github.com/nspcc-dev/neofs-s3-gw/pull/1166/files#diff-68e5a2a921a951a39c819d0470049094a5b68135095040955e0633319c785f78R281) about tag removing.
To handle tag removal from the object itself, we store the tags' meta object with empty tags. Any tags meta object has priority against tags inside object attributes

roman-khimov (approved):

Issue #1156, opened 2025-05-06
Optimize tag and lock fetching when doing searches


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we do requests we don't need. `n.comprehensiveSearchAllVersionsInNeoFS` or `n.searchTagsAndLocksInNeoFS` both have lock/tag OIDs after their searches, but we don't utilize them and doing `n.GetObjectTagging` and `n.getLockDataFromObjects` again in `ComprehensiveObjectInfo()` (likely in other places as well).

## Describe the solution you'd like
1. For locks we can request lock-specific attributes and be done with it, that's everything needed.
2. For tags just reuse known tag OIDs and GET them, do not SEARCH again.

PR #1168, opened 2025-05-14
Store lock meta in single attribute

### LINKED ISSUE #1159: Merge lock info into a single attribute
I'm always frustrated when we're using more attributes than we could. We have `S3-BucketSettings-RetentionUntil` and `S3-BucketSettings-ComplianceMode`. Effectively it's all lock metadata, there never a case when we need one of them, but don't need the other, they're never used for searches.

Have a single `S3-Lock-Meta` with JSONised values inside (use more compact time representation like timestamp). This allows to retrieve this attribute alone when searching for objects and get all the lock data for free as well (no need to HEAD the object or perform additional searches).

smallhive:
Closes #1159.

### DIFF
--- a/api/layer/system_object.go
+++ b/api/layer/system_object.go
@@ -108,8 +108,7 @@ func (n *layer) getLockDataFromObjects(ctx context.Context, bkt *data.BucketInfo
 			object.AttributeFilePath,
 			object.AttributeTimestamp,
 			object.AttributeExpirationEpoch,
-			s3headers.AttributeComplianceMode,
-			s3headers.AttributeRetentionUntilMode,
+			s3headers.AttributeLockMeta,
 		}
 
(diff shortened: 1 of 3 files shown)


roman-khimov (changes requested):
1. Documentation isn't updated.
2. It's not complete, `ComprehensiveObjectInfo` can be optimized to not perform additional search (`getLockDataFromObjects`) and `GetObjectTaggingAndLock` can be simplified to deal with tags only while locks can be handled in `GetExtendedObjectInfo` (or just replaced with `ComprehensiveObjectInfo`).

smallhive:
Docs updated.
ComprehensiveObjectInfo optimisation is a #1156 responsibility

roman-khimov (approved):

PR #1171, opened 2025-05-16
Remove old object versions in unversioned container

### LINKED ISSUE #1160: Old objects are not deleted in unversioned container
I'm always frustrated when unversioned container behavior is not compliant to S3, when PUT happens we never delete an old object version which means they pile up forever (and can't even be properly deleted via S3 gw).

Drop the old version when adding a new one.

smallhive:
Closes #1160.

### DIFF
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -17,6 +17,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/minio/sio"
@@ -1556,6 +1557,22 @@ func (n *layer) CompleteMultipartUpload(ctx context.Context, p *CompleteMultipar
 		return nil, nil, fmt.Errorf("couldn't get versioning settings object: %w", err)
 	}
 
+	var (
+		oldVersions    []allVersionsSearchResult
+		oldVersionsErr error
+		wg             sync.WaitGroup
+	)
+
+	if !bktSettings.VersioningEnabled() {
+		wg.Add(1)
+
+		go func() {
+			defer wg.Done()
+
+			oldVersions, oldVersionsErr = n.searchAllVersionsInNeoFS(ctx, p.Info.Bkt, n.Owner(ctx), p.Info.Key, true)
+		}()
+	}
+
 	header, err := n.prepareMultipartHeadObject(ctx, prmHeaderObject, multipartHash, homoHash, uint64(multipartObjetSize), bktSettings.VersioningEnabled())
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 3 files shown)


roman-khimov:
It seems to be legit, but we need to test it for performance, so it'll stay in a queue for a moment.

roman-khimov:
Rebase, please

api/layer/multipart_upload.go
```diff
@@ -1611,6 +1628,42 @@ func (n *layer) CompleteMultipartUpload(ctx context.Context, p *CompleteMultipar
...
+	// Until this point the new object is not in the system.
+	wg.Wait()
+
+	if oldVersionsErr != nil {
+		n.log.Error("search old object versions failed",
+			zap.String("object", p.Info.Key),
+			zap.String("bucket", p.Info.Bkt.Name),
+			zap.Stringer("cid", p.Info.Bkt.CID),
+			zap.Error(err))
+	} else {
```
roman-khimov, line 1640: `if len(oldVersions) != 0`? To avoid messing with routines in this case.

api/layer/object.go
```diff
@@ -280,15 +281,65 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
...
+			oldVersions, oldVersionsErr = n.searchAllVersionsInNeoFS(ctx, p.BktInfo, owner, p.Object, true)
+		}()
 	}
 
 	id, hash, err := n.objectPutAndHash(ctx, prm, p.BktInfo)
 	if err != nil {
 		return nil, err
 	}
 
+	// We need new object id to filter out it.
```
roman-khimov, line 307: to filter it out

api/layer/object.go
```diff
@@ -280,15 +281,65 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
...
+	wg.Wait()
+
+	if oldVersionsErr != nil {
+		n.log.Error("search old object versions failed",
+			zap.String("object", p.Object),
+			zap.String("bucket", p.BktInfo.Name),
+			zap.Stringer("cid", p.BktInfo.CID),
+			zap.Error(err))
+	} else {
```
roman-khimov, line 316: `if len() > 0`?

api/layer/object.go
```diff
@@ -280,15 +281,65 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
...
 	id, hash, err := n.objectPutAndHash(ctx, prm, p.BktInfo)
 	if err != nil {
 		return nil, err
 	}
 
+	// We need new object id to filter out it.
+	wg.Wait()
+
+	if oldVersionsErr != nil {
+		n.log.Error("search old object versions failed",
```
roman-khimov, line 311: Warning here as well? It doesn't prevent gw to work correctly anyway.

api/layer/object.go
smallhive: Fixed

api/layer/object.go
smallhive: Updated

roman-khimov (approved):

Issue #1188, opened 2025-07-11
Use "associate" attribute for tags and locks


roman-khimov:
https://github.com/nspcc-dev/neofs-api/pull/337 introduced a new attribute that associates some object with another one. While https://github.com/nspcc-dev/neofs-s3-gw/pull/1186 used a new SDK, but has not adopted new attribute, eventually we need to move to using it with new APIs introduced by https://github.com/nspcc-dev/neofs-sdk-go/pull/721. Update documentation accordingly.

PR #1189, opened 2025-07-23
api/lock: LOCK objects using object attributes

carpawell:
This is how it should be with API 2.18+.


roman-khimov:
https://github.com/nspcc-dev/neofs-s3-gw/issues/1188

carpawell:
Local results (LOCK tests only)


image

PR #1191, opened 2025-07-24
CHANGELOG: release 0.40.0

roman-khimov:


carpawell (approved):
Why not 39.1? It is like a patch for me.

roman-khimov:
Incompatible changes.

carpawell:
An inevitable fix to me that should have been done (as we now know) in 0.39.

PR #1206, opened 2025-08-19
Update metrics status accorrding to pool

smallhive:


cmd/s3-gw/app_metrics.go
```diff
@@ -63,18 +72,44 @@ func newStateMetrics() *stateMetrics {
...
+	cl, err := m.p.RawClient()
+	// Only "no healthy client" error is possible.
+	if err != nil {
+		m.SetHealth(healthStatusUnhealthy)
+		return
+	}
+
+	// Check actual connection is really alive.
+	_, err = cl.NetworkInfo(ctx, client.PrmNetworkInfo{})
+	if err != nil && strings.Contains(err.Error(), "connection error") {
```
cthulhu-rider, line 96: i'd avoid error stringing approach AMAP, it can break at any t. If u need to catch this error, it should be exposed from the package

cmd/s3-gw/app_metrics.go
```diff
@@ -63,18 +72,44 @@ func newStateMetrics() *stateMetrics {
...
 }
 
+func (m *stateMetrics) updateHealthStatus() {
+	ctx, cancel := context.WithTimeout(context.Background(), checkNeoFSConnectionTimeOut)
+	defer cancel()
+
+	cl, err := m.p.RawClient()
+	// Only "no healthy client" error is possible.
+	if err != nil {
+		m.SetHealth(healthStatusUnhealthy)
```
cthulhu-rider, line 90: i'd dedicate separate metric for this. Some kind of conn probe, unlike existing one reflecting app stages. Moreover, it relies on single connection. How this gonna behave? If we pool 9 healthy conns and 1 unhealthy one, can we unhealthy here? Cuz if so, the metric will poorly reflect the likelihood of OK request

cmd/s3-gw/app_metrics.go
```diff
@@ -63,18 +72,44 @@ func newStateMetrics() *stateMetrics {
...
+		return
+	}
+
+	// Check actual connection is really alive.
```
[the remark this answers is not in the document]
smallhive: Removed

cthulhu-rider (approved):

PR #1219, opened 2025-09-16
metrics: don't exclude PUT/POST requests from histograms

roman-khimov:
Give more data to user.


carpawell (commented):
Dont we need different metrics for different http requests?

roman-khimov:
Yes, different S3 methods should be (and are) tracked separately. GetObject is very much different from PutObject.

PR #1273, opened 2026-04-01
metrics: use atomics instead of shared map for counters

roman-khimov:
The map itself is initialized sequentially when API handlers are instantiated, then counters don't need to lock anything, atomic increments work fine there.

### DIFF
--- a/api/metrics/api.go
+++ b/api/metrics/api.go
@@ -2,7 +2,7 @@ package metrics
 
 import (
 	"io"
-	"maps"
+	"math"
 	"net/http"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


api/metrics/api.go
```diff
@@ -67,48 +64,57 @@ var (
...
-			float64(value),
+			float64(value.Load()),
 			api,
 		)
 	}
 }
 
 // APIStats wraps http handler for api with basic statistics collection.
 func APIStats(api string, f http.HandlerFunc) http.HandlerFunc {
+	if httpStatsMetric.currentS3Requests[api] == nil {
```
roman-khimov: Maybe these ifs are not needed at all, btw.

api/metrics/api.go
```diff
@@ -67,48 +64,57 @@ var (
...
+		httpStatsMetric.totalS3Requests[api] = new(atomic.Uint64)
+	}
+	if httpStatsMetric.totalS3Errors[api] == nil {
+		httpStatsMetric.totalS3Errors[api] = new(atomic.Uint64)
+	}
 	return func(w http.ResponseWriter, r *http.Request) {
-		httpStatsMetric.currentS3Requests.Inc(api)
-		defer httpStatsMetric.currentS3Requests.Dec(api)
+		httpStatsMetric.currentS3Requests[api].Add(1)
+		defer httpStatsMetric.currentS3Requests[api].Add(math.MaxUint64)
```
smallhive, line 112: math.MaxUint64?

api/metrics/api.go
smallhive: It seems we may remove it, each `APIStats(apiName...` call seems unique

api/metrics/api.go
roman-khimov: It's unsigned and you can't `Add(-1)`.

smallhive: Right, unit64

smallhive (dismissed):

api/metrics/api.go
roman-khimov: Removed.

smallhive (approved):

PR #1280, opened 2026-04-23
api: drop extra buffers on object reads

End-rey:
Refs https://github.com/nspcc-dev/neofs-sdk-go/pull/809.

### DIFF
--- a/api/handler/get.go
+++ b/api/handler/get.go
@@ -2,7 +2,6 @@ package handler
 
 import (
 	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -23,10 +22,6 @@ type conditionalArgs struct {
 	IfMatch           string
 	IfNoneMatch       string
 }
-
-const (
-	defaultBufferSize = 128 * 1024
-)
 
 func fetchRangeHeader(headers http.Header, fullSize uint64) (*layer.RangeParams, error) {
 	const prefix = "bytes="
@@ -281,13 +276,7 @@ func (h *handler) GetObjectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var bufferSize = min(info.Size, defaultBufferSize)
-	if bufferSize == 0 {
-		bufferSize = defaultBufferSize
-	}
-
-	buf := make([]byte, bufferSize)
-	if _, err = io.CopyBuffer(w, objectWithPayloadReader.Payload, buf); err != nil {
+	if _, err = objectWithPayloadReader.Payload.WriteTo(w); err != nil {
 		h.logAndSendError(w, "could write object output", reqInfo, err)
 	}
 
--- a/api/layer/layer.go
+++ b/api/layer/layer.go
@@ -209,7 +209,7 @@ type (
 	// ObjectWithPayloadReader is a response for Client.GetObjectWithPayloadReader.
 	ObjectWithPayloadReader struct {
 		Head       *object.Object
-		Payload    io.ReadCloser
+		Payload    PayloadReadCloser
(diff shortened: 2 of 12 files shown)


api/handler/get.go
```diff
@@ -281,13 +277,7 @@ func (h *handler) GetObjectHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	var bufferSize = min(info.Size, defaultBufferSize)
-	if bufferSize == 0 {
-		bufferSize = defaultBufferSize
-	}
-
-	buf := make([]byte, bufferSize)
-	if _, err = io.CopyBuffer(w, objectWithPayloadReader.Payload, buf); err != nil {
+	if _, err = io.Copy(w, objectWithPayloadReader.Payload); err != nil {
```
roman-khimov: Just `objectWithPayloadReader.Payload.WriteTo(w)`, it's 10ns faster this way.

api/layer/layer.go
```diff
@@ -560,7 +552,7 @@ func (n *layer) GetObject(ctx context.Context, p *GetObjectParams) error {
 	}
 
 	// copy full payload
-	written, err := io.CopyBuffer(p.Writer, r, buf)
+	written, err := io.Copy(p.Writer, r)
```
roman-khimov: Same here. `io.ReaderCloser` is used anywhere it can be combined with `io.WriterTo` in some new interface, but I'd like to see it all working correctly, `io.Copy()` doesn't guarantee that.

internal/neofs/neofs.go
```diff
@@ -552,6 +552,19 @@ func (x payloadReader) Read(p []byte) (int, error) {
 	return n, err
 }
 
+func (x payloadReader) WriteTo(w io.Writer) (int64, error) {
+	if wt, ok := x.ReadCloser.(io.WriterTo); ok {
```
roman-khimov: Make `x.ReadCloser` include and demand `WriteTo`, we don't want fallbacks.

roman-khimov (commented):
Please fix linter, homomorphic things can be stubbed out with `//nolint`, they'll be removed after node 0.53.0 release.

api/handler/get.go
End-rey: Done.

api/layer/layer.go
End-rey: Made new interface.

internal/neofs/neofs.go
End-rey: Fixed.

internal/neofs/neofs.go
```diff
@@ -187,11 +188,6 @@ func (x *NeoFS) CreateContainer(ctx context.Context, prm layer.PrmContainerCreat
 		cnr.SetAttribute(containerMetaDataPolicyAttribute, prm.Policy.Consistency)
 	}
 
-	err := client.SyncContainerWithNetwork(ctx, &cnr, x.pool)
```
roman-khimov, line 190: I guess this one gives side-effects for us in tests, needs to be fixed, there is exactly one thing to check for --- homomorphic hash and respective attribute.

https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/948-1776977553/index.html#suites/3a7ef537e45a1b5bc9d9513ff7884beb/e671415615b347a7/

roman-khimov (commented):
Nice otherwise.

internal/neofs/neofs.go
End-rey: The tests seem to have gone well, everything is working.

roman-khimov (approved):

PR #1284, opened 2026-05-08
Drop settings meta objects

smallhive:
Closes #1265.

There is one issue with `[test_100_continue](https://github.com/nspcc-dev/s3-tests/blob/eba16d41ce7da553897531a4bd6edae79463abbe/s3tests_boto3/functional/test_s3.py#L8151)` in `s3-tests`.

It expects 403 on the first step. Previously, we had a search for settings in a private container and got 403. Now we don't have a search, and the test fails on [this](https://github.com/nspcc-dev/s3-tests/blob/eba16d41ce7da553897531a4bd6edae79463abbe/s3tests_boto3/functional/test_s3.py#L8164) check.

I consider, in this case, removing two strings with a 403 check from the test is a good decision. In another case, make an empty search to get the correct error - sounds strange and affects performance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This document outlines major changes between releases.
 
 ### Removed
 - Support for legacy access boxes carrying session token v1 and bearer tokens (#1271)
+- Backward-compatibility code for legacy bucket settings, CORS, tags and notification meta objects (#1265)
 
 ### Upgrading from 0.44.0
 
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -10,16 +10,9 @@ import (
 )
 
 const (
-	bktSettingsObject                  = ".s3-settings"
-	bktCORSConfigurationObject         = ".s3-cors"
-	bktNotificationConfigurationObject = ".s3-notifications"
-
 	VersioningUnversioned = "Unversioned"
 	VersioningEnabled     = "Enabled"
 	VersioningSuspended   = "Suspended"
-
-	// BucketSettingsV1 describes v1 version identifier for the bucket settings file.
-	BucketSettingsV1 = "1"
 )
 
 const (
@@ -123,16 +116,6 @@ func NotificationInfoFromObject(objInfo *ObjectInfo) *NotificationInfo {
 		Size:    objInfo.Size,
 		HashSum: objInfo.HashSum,
 	}
-}
-
-// SettingsObjectName is a system name for a bucket settings file.
-func (b *BucketInfo) SettingsObjectName() string { return bktSettingsObject }
-
-// CORSObjectName returns a system name for a bucket CORS configuration file.
-func (b *BucketInfo) CORSObjectName() string { return bktCORSConfigurationObject }
-
-func (b *BucketInfo) NotificationConfigurationObjectName() string {
-	return bktNotificationConfigurationObject
 }
 
 // VersionID returns object version from ObjectInfo.
--- a/api/data/info.go
+++ b/api/data/info.go
@@ -1,7 +1,9 @@
 package data
 
 import (
+	"encoding/json"
 	"encoding/xml"
+	"fmt"
 	"time"
 
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
@@ -45,6 +47,10 @@ type (
 		AttributeTags          string
 		AttributeSettings      string
 		AttributeNotifications string
+		Settings               *BucketSettings
+		CORS                   *CORSConfiguration
+		Tags                   map[string]string
+		Notifications          *NotificationConfiguration
 	}
 
 	// ObjectInfo holds S3 object data.
@@ -139,4 +145,44 @@ func (b BucketSettings) VersioningEnabled() bool {
 
 func (b BucketSettings) VersioningSuspended() bool {
 	return b.Versioning == VersioningSuspended
+}
+
+// ParseSettings parses the container attributes into Settings, CORS, Tags and Notifications.
+func (b *BucketInfo) ParseSettings() error {
+	b.Settings = &BucketSettings{Versioning: VersioningUnversioned}
+	b.Notifications = &NotificationConfiguration{}
+
+	if b.AttributeSettings != "" {
+		settings := &BucketSettings{Versioning: VersioningUnversioned}
+		if err := json.Unmarshal([]byte(b.AttributeSettings), settings); err != nil {
+			return fmt.Errorf("malformed bucket settings: %w", err)
+		}
+		b.Settings = settings
+	}
+
+	if b.AttributeCors != "" {
+		var corsRules []CORSRule
+		if err := json.Unmarshal([]byte(b.AttributeCors), &corsRules); err != nil {
+			return fmt.Errorf("malformed bucket CORS: %w", err)
+		}
+		b.CORS = &CORSConfiguration{CORSRules: corsRules}
+	}
+
+	if b.AttributeTags != "" {
+		var tags map[string]string
+		if err := json.Unmarshal([]byte(b.AttributeTags), &tags); err != nil {
+			return fmt.Errorf("malformed bucket tags: %w", err)
+		}
+		b.Tags = tags
+	}
+
+	if b.AttributeNotifications != "" {
+		conf := &NotificationConfiguration{}
+		if err := json.Unmarshal([]byte(b.AttributeNotifications), conf); err != nil {
+			return fmt.Errorf("malformed bucket notifications: %w", err)
+		}
+		b.Notifications = conf
+	}
+
+	return nil
 }
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -297,13 +297,7 @@ func (h *handler) PutBucketACLHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	settings, err := h.obj.GetBucketSettings(r.Context(), bktInfo)
-	if err != nil {
-		h.logAndSendError(w, "could not get bucket settings", reqInfo, err)
-		return
-	}
-
-	if settings.BucketOwner == data.BucketOwnerEnforced {
+	if bktInfo.Settings.BucketOwner == data.BucketOwnerEnforced {
 		if !isValidOwnerEnforced(r) {
 			h.logAndSendError(w, "access control list not supported", reqInfo, s3errors.GetAPIError(s3errors.ErrAccessControlListNotSupported))
 			return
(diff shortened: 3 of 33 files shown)


roman-khimov:
> In another case, make an empty search to get the correct error - sounds strange and affects performance

The test is about PUT and the problem is that there is no NeoFS-level PUT yet. Why there is no PUT if we already have all headers and just waiting for the data?

smallhive:
> The test is about PUT and the problem is that there is no NeoFS-level PUT yet. Why there is no PUT if we already have all headers and just waiting for the data?

It is a good question. I'm not sure why test tries to do this. Maybe @evgeniiz321 can help us understand this

api/layer/cache.go
```diff
@@ -188,20 +194,20 @@ func (c *Cache) PutLockInfo(owner user.ID, key string, lockInfo *data.LockInfo)
 }
 
 func (c *Cache) GetSettings(bktInfo *data.BucketInfo) *data.BucketSettings {
-	key := bktInfo.Name + bktInfo.SettingsObjectName()
+	key := bktInfo.Name + settingsCacheKeySuffix
```
roman-khimov: We no longer need a separate cache. The same bucket cache can hold this data and follow bucket data update rules.

api/layer/cors.go
```diff
@@ -85,71 +80,27 @@ func (n *layer) PutBucketCORS(ctx context.Context, p *PutCORSParams) error {
...
-		}
-		if err = n.deleteBucketCORS(ctx, bktInfo); err != nil {
-			return nil, fmt.Errorf("delete bucket CORS: %w", err)
-		}
+	var corsRules []data.CORSRule
+	if err := json.Unmarshal([]byte(bktInfo.AttributeCors), &corsRules); err != nil {
+		return nil, fmt.Errorf("malformed data: %w", err)
 	}
 
+	cors := &data.CORSConfiguration{CORSRules: corsRules}
```
roman-khimov: Should be like this: if you have `bktInfo` you have cors/settings/etc.

roman-khimov (changes requested):
docs still mention old objects

roman-khimov:
> It is a good question. I'm not sure why test tries to do this.

I'm not talking about test, test does everything correctly here. I'm talking about gateway.

api/layer/cache.go
smallhive: Updated

smallhive:
> > It is a good question. I'm not sure why test tries to do this.
> 
> I'm not talking about test, test does everything correctly here. I'm talking about gateway.

Long and deep debugging eventually gets results. The issue was in checking access. After removing meta objects check check was too late

smallhive:
> docs still mention old objects

Removed

api/layer/cache.go
```diff
@@ -81,6 +77,19 @@ func (c *Cache) PutBucket(bktInfo *data.BucketInfo) {
 
 func (c *Cache) DeleteBucket(name string) {
 	c.bucketCache.Delete(name)
+	c.systemCache.Delete(name + settingsCacheKeySuffix)
+}
+
+// GetSettings returns cached bucket settings.
+func (c *Cache) GetSettings(bktInfo *data.BucketInfo) *data.BucketSettingsCacheItem {
+	return c.systemCache.GetSettings(bktInfo.Name + settingsCacheKeySuffix)
```
roman-khimov: It's still a separate cache, why not a part of `data.BucketInfo`?

api/handler/put.go
```diff
@@ -218,6 +218,23 @@ func (h *handler) PutObjectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Anonymous PUT must be decided on headers alone, before any body read,
+	// so a private bucket gets a clean 403 instead of a 100-Continue followed
+	// by a late rejection (see s3-tests->test_100_continue).
+	if api.IsAnonymousRequest(r.Context()) {
```
roman-khimov: Non-anonymous requests are the same wrt 100 expectations.

api/handler/acl.go
```diff
@@ -1753,6 +1753,32 @@ func isValidOwnerPreferred(r *http.Request) bool {
...
+// Called from PutObjectHandler before any body read so that an anonymous PUT
+// against a private bucket is rejected with 403 without first triggering
+// net/http's 100-Continue response.
+func eaclAllowsAnonPut(table *eacl.Table) bool {
+	if table == nil {
+		return false
+	}
+
+	for _, r := range table.Records() {
+		if r.Operation() != eacl.OperationPut || len(r.Filters()) != 0 {
```
roman-khimov: Interpreting EACL in gateway is a bad idea. You should be able to invoke PUT and receive an error if something is wrong with headers.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ This document outlines major changes between releases.
 ### Updated
 
 ### Removed
+- Backward-compatibility code for legacy bucket settings, CORS, tags and
+  notification meta objects (#1265)
```
roman-khimov: Usually these are one-liners.

api/handler/acl.go
smallhive: Yeah, but there are some moments to think:
* The test sends a request with some [headers](https://github.com/nspcc-dev/s3-tests/blob/4d9103e476101f86918d85fafc911e412bbe66f9/s3tests_boto3/functional/test_s3.py#L8118) via socket.
* The gate tries to put the object; trying to read body; in this moment `net/http` sends http 100 automatically. (https://pkg.go.dev/net/http, The server will automatically send a 100 (Continue) header)
* The test closes connection; the gate has context cancelled, but 100 has already been sent. Here we have the test failure.

In case we ignore client context (with context.Backgroud() for instance), we get an unexpected EOF, because the test [announced](https://github.com/nspcc-dev/s3-tests/blob/4d9103e476101f86918d85fafc911e412bbe66f9/s3tests_boto3/functional/test_s3.py#L8118) 123 bytes of payload, but connection already closed.

Before, it worked, because we had a search (meta object) which failed, and it didn't read the payload.

Thus, we should either update the test to make more accurate requests or have a possibility in the gate to understand: could we put the object or not without reading the request payload. All these moment prevents EACL check on the node side

api/layer/cache.go
smallhive: Merged

CHANGELOG.md
smallhive: Updated

api/handler/acl.go
smallhive: Another option is to make a fake search for 1 element (to reproduce previous behaviour) before payload read. It will allow to check EACL by the SN node in this case; I consider it should work as well

roman-khimov: We already perform a search in PUT, likely it can be just synchronized with the rest of the flow in a way that it's complete before we process the data. Can affect performance, but let's check.

smallhive: This search was removed with the settings meta object. I've added a lightweight search.
The test is green
```
257 passed, 105 skipped, 191 deselected, 4485 warnings in 602.26s (0:10:02) = branch
257 passed, 105 skipped, 191 deselected, 4485 warnings in 596.86s (0:09:56) = master
```
It seems the performance is the same for the s3-tests collection

roman-khimov: What kind of lightweight search we're talking about? There is a search done in PUT already, I don't see any need for an additional one.

smallhive: Reused old versions search

api/data/info.go
```diff
@@ -45,6 +47,7 @@ type (
 		AttributeTags          string
 		AttributeSettings      string
 		AttributeNotifications string
+		SettingsItem           *BucketSettingsCacheItem
```
roman-khimov: Why a pointer here? All buckets have settings of some kind.

api/data/info.go
```diff
@@ -106,6 +109,14 @@ type (
 		ExposeHeaders  []string `xml:"ExposeHeader" json:"ExposeHeaders"`
 		MaxAgeSeconds  int      `xml:"MaxAgeSeconds,omitempty" json:"MaxAgeSeconds,omitempty"`
 	}
+
+	// BucketSettingsCacheItem holds the bucket-level configuration.
+	BucketSettingsCacheItem struct {
```
roman-khimov: I'm not sure you need a separate structure, it's all a part of BucketInfo.

api/data/info.go
```diff
@@ -140,3 +151,45 @@ func (b BucketSettings) VersioningEnabled() bool {
 func (b BucketSettings) VersioningSuspended() bool {
 	return b.Versioning == VersioningSuspended
 }
+
+// NewBucketSettingsCacheItem parses the bucket configuration container attributes into a cache item.
+func NewBucketSettingsCacheItem(b *BucketInfo) (*BucketSettingsCacheItem, error) {
```
roman-khimov: This helper can populate `b` fields and return an error, likely it can even be unexported (implemented where appropriate).

api/layer/system_object.go
```diff
@@ -232,20 +232,21 @@ func lockObjectKey(objVersion *ObjectVersion) string {
...
+	if bktInfo.SettingsItem != nil {
+		return bktInfo.SettingsItem, nil
 	}
 
-	var settings = &data.BucketSettings{Versioning: data.VersioningUnversioned}
-	if bktInfo.AttributeSettings != "" {
-		if err := json.Unmarshal([]byte(bktInfo.AttributeSettings), settings); err != nil {
-			return nil, fmt.Errorf("malformed data: %w", err)
-		}
+	return data.NewBucketSettingsCacheItem(bktInfo)
```
roman-khimov: Because you don't need it here, `bktInfo` MUST have the data.

api/layer/object.go
```diff
@@ -242,6 +242,17 @@ func (n *layer) PutObject(ctx context.Context, p *PutObjectParams) (*data.Extend
 		p.Size = int64(encSize)
 	}
 
+	oldVersions, oldVersionsErr := n.searchAllVersionsInNeoFS(ctx, p.BktInfo, p.Object, true)
```
roman-khimov, line 242: Why searching for versioned buckets?

api/data/info.go
smallhive: Moved fields from `BucketSettingsCacheItem` into `BucketInfo`

api/data/info.go
smallhive: Removed

api/data/info.go
smallhive: Removed

api/layer/system_object.go
smallhive: Done

api/layer/object.go
smallhive: Vice versa, we search only in the unversioned bucket to be able to remove previous versions of the object if you reupload the object. We move the search from another place, but save the functionality

roman-khimov: I mean you always do the search now, previously it was for unversioned buckets only.

roman-khimov:
Conflicts.

api/layer/object.go
smallhive: Yes, but for this test we need to make _something_ to be able to check EACL. And as we discussed, this Search is a good candidate for this action

roman-khimov (approved):

PR #1285, opened 2026-05-08
Drop session v1 and bearer tokens support

smallhive:
Closes #1271.


api/handler/acl.go
```diff
@@ -22,7 +22,6 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/eacl"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
-	"github.com/nspcc-dev/neofs-sdk-go/session"
 	session2 "github.com/nspcc-dev/neofs-sdk-go/session/v2"
```
roman-khimov: This should not be renamed on import now.

smallhive: Renamed and updated

roman-khimov (approved):

PR #1287, opened 2026-05-22
Route GET through SN HTTP/2 API

End-rey:
Refs https://github.com/nspcc-dev/neofs-node/pull/3973.


roman-khimov:
Experiment done, won't be merged, see https://github.com/nspcc-dev/neofs-node/issues/3972

PR #1293, opened 2026-06-15
Group encryption parameters under a single attribute

### LINKED ISSUE #1279: Group encryption parameters under a single attribute
I'm always frustrated when we're using multiple attributes without any real need for it. We've got:
 * S3-Algorithm
 * S3-Decrypted-Size
 * S3-HMAC-Salt
 * S3-HMAC-Key

No unencrypted object has any of them. All encrypted objects have them. All of these attributes are needed at the same time when processing objects, we never SEARCH for any specific of them.

Replace this set with a single `S3-Encryption-Meta` similar to `S3-Lock-Meta`. Easy to set, easy to get.

smallhive:
Closes #1279

### DIFF
--- a/api/layer/multipart_upload.go
+++ b/api/layer/multipart_upload.go
@@ -178,8 +178,15 @@ func (n *layer) CreateMultipartUpload(ctx context.Context, p *CreateMultipartPar
 	}
 
 	if p.Info.Encryption.Enabled() {
-		if err := addEncryptionHeaders(info.Meta, p.Info.Encryption); err != nil {
-			return "", fmt.Errorf("add encryption header: %w", err)
+		hmacKey, hmacSalt, err := p.Info.Encryption.HMAC()
+		if err != nil {
+			return "", fmt.Errorf("encryption get hmac: %w", err)
+		}
+
+		info.EncryptionMeta = &data.EncryptionMeta{
+			Algorithm: AESEncryptionAlgorithm,
+			HMACSalt:  hmacSalt,
+			HMACKey:   hmacKey,
 		}
 	}
 
@@ -234,7 +241,7 @@ func (n *layer) UploadPart(ctx context.Context, p *UploadPartParams) (string, er
 }
 
 func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInfo, p *UploadPartParams) (*data.ObjectInfo, error) {
-	encInfo := FormEncryptionInfo(multipartInfo.Meta)
+	encInfo := FormEncryptionInfoFromMeta(multipartInfo.EncryptionMeta)
 	if err := p.Info.Encryption.MatchObjectEncryption(encInfo); err != nil {
 		n.log.Warn("mismatched obj encryptionInfo", zap.Error(err))
 		return nil, s3errors.GetAPIError(s3errors.ErrInvalidEncryptionParameters)
@@ -258,7 +265,7 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 		if err != nil {
 			return nil, fmt.Errorf("failed to create ecnrypted reader: %w", err)
 		}
-		attributes[s3headers.AttributeDecryptedSize] = strconv.FormatInt(p.Size, 10)
+		attributes[s3headers.AttributeEncryptionMeta] = packEncryptionMeta(data.EncryptionMeta{DecryptedSize: p.Size})
 		payloadReader = r
 		p.Size = int64(encSize)
 	}
@@ -389,7 +396,7 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
 }
 
 func (n *layer) uploadZeroPart(ctx context.Context, multipartInfo *data.MultipartInfo, p *UploadInfoParams) (*data.PartInfo, error) {
-	encInfo := FormEncryptionInfo(multipartInfo.Meta)
+	encInfo := FormEncryptionInfoFromMeta(multipartInfo.EncryptionMeta)
 	if err := p.Encryption.MatchObjectEncryption(encInfo); err != nil {
 		n.log.Warn("mismatched obj encryptionInfo", zap.Error(err))
 		return nil, s3errors.GetAPIError(s3errors.ErrInvalidEncryptionParameters)
@@ -409,10 +416,6 @@ func (n *layer) uploadZeroPart(ctx context.Context, multipartInfo *data.Multipar
 		creationTime  = TimeNow(ctx)
 	)
 
-	if p.Encryption.Enabled() {
-		attributes[s3headers.AttributeDecryptedSize] = "0"
-	}
-
 	attrs := make([]object.Attribute, 0, len(multipartInfo.Meta)+1)
 	attrs = append(attrs, object.NewAttribute(object.AttributeTimestamp, strconv.FormatInt(creationTime.Unix(), 10)))
 
@@ -422,10 +425,8 @@ func (n *layer) uploadZeroPart(ctx context.Context, multipartInfo *data.Multipar
 		}
 	}
 
-	if encInfo.Enabled {
-		attrs = append(attrs, object.NewAttribute(s3headers.AttributeEncryptionAlgorithm, encInfo.Algorithm))
-		attrs = append(attrs, object.NewAttribute(s3headers.AttributeHMACKey, encInfo.HMACKey))
-		attrs = append(attrs, object.NewAttribute(s3headers.AttributeHMACSalt, encInfo.HMACSalt))
+	if multipartInfo.EncryptionMeta != nil {
+		attrs = append(attrs, object.NewAttribute(s3headers.AttributeEncryptionMeta, packEncryptionMeta(*multipartInfo.EncryptionMeta)))
 	}
 
 	var hashlessHeaderObject object.Object
@@ -636,6 +637,10 @@ func (n *layer) createMultipartInfoObject(ctx context.Context, bktInfo *data.Buc
 		payloadMap[s3headers.MultipartMeta] = base64.StdEncoding.EncodeToString(objectAttributes)
 	}
 
+	if info.EncryptionMeta != nil {
+		payloadMap[s3headers.AttributeEncryptionMeta] = packEncryptionMeta(*info.EncryptionMeta)
+	}
+
 	payload, err := json.Marshal(payloadMap)
 	if err != nil {
 		return fmt.Errorf("meta marshaling: %w", err)
@@ -796,6 +801,14 @@ func (n *layer) parseMultipartInfoObject(uploadID string, obj *object.Object) (d
 		}
 	}
 
+	if encRaw, ok := payloadMap[s3headers.AttributeEncryptionMeta]; ok {
+		encMeta, err := unpackEncryptionMeta(encRaw)
+		if err != nil {
+			return data.MultipartInfo{}, fmt.Errorf("unpack encryption meta %s: %w", uploadID, err)
+		}
+		multipartInfo.EncryptionMeta = encMeta
+	}
+
 	return multipartInfo, nil
 }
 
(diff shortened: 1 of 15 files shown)


api/layer/util.go
```diff
@@ -140,6 +151,49 @@ func addEncryptionHeaders(meta map[string]string, enc encryption.Params) error {
 	return nil
 }
 
+func packEncryptionMeta(meta map[string]string) error {
```
roman-khimov: Why a map? It's a natural struct to me.

api/layer/object.go
```diff
@@ -858,9 +867,22 @@ func (n *layer) searchAllVersionsInNeoFSByPrefix(ctx context.Context, bkt *data.
...
 
 		if item.Attributes[3] != "" {
-			psr.DecryptedSize, err = strconv.ParseInt(item.Attributes[3], 10, 64)
+			fields := make(map[string]string)
+			if err = json.Unmarshal([]byte(item.Attributes[3]), &fields); err != nil {
+				return nil, "", fmt.Errorf("unmarshal encryption meta: %w", err)
+			}
+
+			if ds := fields[s3headers.FieldDecryptedSize]; ds != "" {
+				psr.DecryptedSize, err = strconv.ParseInt(ds, 10, 64)
```
roman-khimov: Why duplicating it?

api/layer/multipart_upload.go
```diff
@@ -982,10 +1000,22 @@ func (n *layer) multipartGetPartsList(ctx context.Context, bktInfo *data.BucketI
...
+				return nil, fmt.Errorf("unmarshal encryption meta for multipart %s: %w", uploadID, err)
+			}
+
+			if ds := fields[s3headers.FieldDecryptedSize]; ds != "" {
+				element.TotalSize, err = strconv.ParseInt(ds, 10, 64)
+				if err != nil {
+					return nil, fmt.Errorf("convert multipart %s decryptedSize from enc meta: %w", uploadID, err)
+				}
+			}
+		} else if decSize := item.Attributes[5]; decSize != "" {
```
roman-khimov: The goal is to simplify the scheme.

api/headers.go
```diff
@@ -97,4 +99,6 @@ var SystemMetadata = map[string]struct{}{
 	ContentType:        {},
 	LastModified:       {},
 	ETag:               {},
+
+	s3headers.AttributeEncryptionMeta: {},
```
roman-khimov: Why old headers were not there?

api/layer/util.go
smallhive: We worked with attributes in the map, and it was logical to use them here as well

api/layer/object.go
smallhive: It's backwards compatibility: write new fields only, but read new and old fields as well. If we don't need it, I will remove old code

api/layer/multipart_upload.go
smallhive: It's backwards compatibility: write new fields only, but read new and old fields as well. If we don't need it, I will remove old code

api/headers.go
smallhive: Ouch. Because neither the old headers nor this one needs to be added here. Removed. It is a DEBUG code in some way

api/layer/object.go
roman-khimov: Compatibility can be dropped in this case.

smallhive: Do I need to remove all backward compatibility, just ignoring old objects?

api/layer/util.go
roman-khimov: We have a limited number of well-known attributes here, passing them via a map is just adding overhead.

api/layer/object.go
roman-khimov: I think so. It was never used.

api/layer/util.go
smallhive: Replaced with struct

api/layer/object.go
smallhive: Removed

docs/attribute_mapping.md
```diff
@@ -184,21 +184,40 @@ Contains JSON encoded lock metadata in the next fields:
...
+
+### S3-Meta-Algorithm
+
+Deprecated, replaced by `S3-Encryption-Meta`.
+
+Contains encryption algorithm for object payload, if encryption is enabled.
+
+### S3-Meta-Decrypted-Size
+
+Deprecated, replaced by `S3-Encryption-Meta`.
```
roman-khimov: I think it's removed (as well as others). Also, docs should be updated in the same commit.

api/layer/multipart_upload.go
```diff
@@ -258,7 +258,16 @@ func (n *layer) uploadPart(ctx context.Context, multipartInfo *data.MultipartInf
...
 		}
-		attributes[s3headers.AttributeDecryptedSize] = strconv.FormatInt(p.Size, 10)
+		packed, err := packEncryptionMeta(encryptionMeta{
+			DecryptedSize: strconv.FormatInt(p.Size, 10),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("pack encryption meta: %w", err)
+		}
+
+		if packed != "" {
```
roman-khimov: Why would it be?

api/layer/multipart_upload.go
```diff
@@ -410,7 +419,14 @@ func (n *layer) uploadZeroPart(ctx context.Context, multipartInfo *data.Multipar
 	)
 
 	if p.Encryption.Enabled() {
-		attributes[s3headers.AttributeDecryptedSize] = "0"
+		packed, err := packEncryptionMeta(encryptionMeta{DecryptedSize: "0"})
+		if err != nil {
+			return nil, fmt.Errorf("pack encryption meta: %w", err)
+		}
+
+		if packed != "" {
```
roman-khimov: Here as well.

api/layer/multipart_upload.go
```diff
@@ -982,10 +1020,15 @@ func (n *layer) multipartGetPartsList(ctx context.Context, bktInfo *data.BucketI
...
 
-		if decSize := item.Attributes[4]; decSize != "" {
-			element.TotalSize, err = strconv.ParseInt(decSize, 10, 64)
+		encMeta, err := unpackEncryptionMeta(item.Attributes[4])
+		if err != nil {
+			return nil, fmt.Errorf("unmarshal encryption meta for multipart %s: %w", uploadID, err)
+		}
+
+		if encMeta != nil && encMeta.DecryptedSize != "" {
+			element.TotalSize, err = strconv.ParseInt(encMeta.DecryptedSize, 10, 64)
```
roman-khimov: `unpackEncryptionMeta` should handle atoi

api/layer/util.go
```diff
@@ -48,13 +49,24 @@ type (
 		DeleteMarker        []*data.ExtendedObjectInfo
 		VersionIDMarker     string
 	}
+
+	encryptionMeta struct {
+		Algorithm     string `json:"algo,omitempty"`
+		DecryptedSize string `json:"decSize,omitempty"`
```
roman-khimov: Why not an integer here?

api/layer/util.go
```diff
@@ -48,13 +49,24 @@ type (
...
+		HMACSalt      string `json:"salt,omitempty"`
+		HMACKey       string `json:"key,omitempty"`
```
roman-khimov: Can be `[]byte` with auto-base64.

api/layer/util.go
```diff
@@ -48,13 +49,24 @@ type (
...
+	}
 )
 
 var (
 	// ErrLinkingObjectNotFound means there is no linking object for parent.
 	ErrLinkingObjectNotFound = errors.New("linking object not found")
 )
 
+func (m encryptionMeta) IsEmty() bool {
```
roman-khimov: Emty?

api/layer/util.go
```diff
@@ -140,6 +164,46 @@ func addEncryptionHeaders(meta map[string]string, enc encryption.Params) error {
...
+		Algorithm:     AESEncryptionAlgorithm,
+		DecryptedSize: decryptedSize,
+		HMACSalt:      hex.EncodeToString(hmacSalt),
+		HMACKey:       hex.EncodeToString(hmacKey),
+	}, nil
+}
+
+func packEncryptionMeta(meta encryptionMeta) (string, error) {
+	if meta.IsEmty() {
+		return "", nil
```
roman-khimov: Can we just not try packing empty meta? Because the only case where it's really relevant is size==0 and unset other fields. OK, you can avoid adding the header at all in this case, but otherwise you're always passing something real here and can't get an empty string in the output.

api/s3headers/headers.go
```diff
@@ -40,6 +40,9 @@ const (
 	// Result: S3-Lock-Meta.
 	AttributeLockMeta = attributePrefix + "Lock-Meta"
 
+	// Result: S3-Encryption-Meta.
+	AttributeEncryptionMeta = attributePrefix + "Encryption-Meta"
+
```
roman-khimov, line 45: Old ones not deleted?

api/layer/multipart_upload.go
smallhive: No way, updated

api/layer/multipart_upload.go
smallhive: Updated

api/layer/util.go
smallhive: Fixed

api/s3headers/headers.go
smallhive: Reworked and simplified

docs/attribute_mapping.md
smallhive: Squashed and removed

api/layer/util.go
smallhive: Changed to int

api/layer/multipart_upload.go
smallhive: Done

api/layer/util.go
smallhive: Updated

api/layer/util.go
smallhive: Reworked

api/layer/util.go
```diff
@@ -106,38 +113,66 @@ func objectInfoFromMeta(bkt *data.BucketInfo, meta *object.Object) *data.ObjectI
...
+
 	return encryption.ObjectEncryption{
-		Enabled:   len(algorithm) > 0,
-		Algorithm: algorithm,
-		HMACKey:   headers[s3headers.AttributeHMACKey],
-		HMACSalt:  headers[s3headers.AttributeHMACSalt],
+		Enabled:   true,
+		Algorithm: meta.Algorithm,
+		HMACKey:   hex.EncodeToString(meta.HMACKey),
+		HMACSalt:  hex.EncodeToString(meta.HMACSalt),
```
roman-khimov: Why encoding this again?

api/layer/util.go
```diff
@@ -106,38 +113,66 @@ func objectInfoFromMeta(bkt *data.BucketInfo, meta *object.Object) *data.ObjectI
...
+		DecryptedSize: decryptedSize,
+		HMACSalt:      hmacSalt,
+		HMACKey:       hmacKey,
+	}, nil
+}
+
+func packEncryptionMeta(meta data.EncryptionMeta) (string, error) {
+	payload, err := json.Marshal(meta)
+	if err != nil {
+		return "", fmt.Errorf("marshal encryption meta: %w", err)
```
roman-khimov: The structure is pretty trivial, I think it's OK to panic here and simplify `packEncryptionMeta()` to not have error returned.

api/layer/util.go
smallhive: Leftovers, updated

api/layer/util.go
smallhive: Simplified

roman-khimov (approved):

PR #1295, opened 2026-06-17
Simplify FS getter

roman-khimov:


smallhive (dismissed):
It looks as good improvements. But something strange with tests, all of them with "Access Denied". Maybe something with test env?

roman-khimov:
Fixed (to be investigated in node).

smallhive (approved):

PR #1296, opened 2026-06-18
Drop object ACL support

### LINKED ISSUE #1275: Drop object ACL support
Judging from https://github.com/minio/minio/issues/8195 Minio doesn't have any support for object ACLs and it's just not an issue. We do have some support for it, but it has known problems (#685, #904, #906 open and we known EACL table can grow too big with the current implementation like in #1055, and we have a list of things in https://github.com/nspcc-dev/neofs-s3-gw/blob/426f732fd72df508248021f39e8261ee62040a43/docs/aws_s3_compat.md#acl). We can drop some kludges (not touch EACL table at all for object operations) and just mimic enough of API for compatibility. No real applications should notice (see Minio example).

This can also speedup some things since EACL won't be touched at all (even for reads) in most cases (let SN handle it).

smallhive:
Closes #1275.

### DIFF
--- a/api/handler/acl.go
+++ b/api/handler/acl.go
@@ -383,44 +383,40 @@ func (h *handler) GetObjectACLHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	bucketACL, err := h.obj.GetBucketACL(r.Context(), bktInfo)
-	if err != nil {
-		h.logAndSendError(w, "could not fetch bucket acl", reqInfo, err)
-		return
-	}
-
 	prm := &layer.HeadObjectParams{
 		BktInfo:   bktInfo,
 		Object:    reqInfo.ObjectName,
 		VersionID: reqInfo.URL.Query().Get(api.QueryVersionID),
 	}
 
-	objInfo, err := h.obj.GetObjectInfo(r.Context(), prm)
-	if err != nil {
+	if _, err = h.obj.GetObjectInfo(r.Context(), prm); err != nil {
 		h.logAndSendError(w, "could not object info", reqInfo, err)
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


roman-khimov:
Too many broken unrelated tests for now. Also, bucket ACLs are very much different from object ACLs. Buckets ACLs can be supported in general (with the minimum number of limitations).

api/handler/acl.go
```diff
@@ -369,6 +397,98 @@ func (h *handler) updateBucketACL(r *http.Request, astChild *ast, bktInfo *data.
...
+		}
+	}
+
+	if cannedACL := r.Header.Get(api.AmzACL); cannedACL != "" {
+		if !specID.IsZero() {
+			// Check if we have special rules.
+			for _, grant := range acl.AccessControlList {
+				for _, id := range aclAccounts {
+					// Bucket and new object has the same canned ACL.
+					if id.String() == grant.Grantee.ID && specID == id {
```
roman-khimov: Why storing it in EACL? These lists can be modified via Morph and everything would break immediately. Bucket-level settings can be stored in bucket-level settings, that is in container attributes.

smallhive: Yeah, it makes the picture a bit simpler. Reworked

roman-khimov: But this makes us dependent on set attribute permission which is not a good thing.

roman-khimov:
Let's use https://github.com/minio/minio/blob/master/cmd/acl-handlers.go for inspiration here, the amount of support done there is sufficient in general. At the same time we can do better. We already have partial support for Bucket ACLs and this can be left as is (put/get). We can at the same time stub out object acls, accept/return "private" for them and do nothing. Do not try to evaluate canned ACL against bucket preset, do not store bucket preset at all. Document the behavior.

The scheme can be adjusted in future if needed.

roman-khimov (approved):

PR #1300, opened 2026-06-23
Support bucket namespace

### LINKED ISSUE #1299: Handle user namespaces
Users can have preferences for their bucket subdomains, this needs to be accounted for when dealing with them.

smallhive:
Closes #1299.

### DIFF
--- a/creds/tokens/credentials.go
+++ b/creds/tokens/credentials.go
@@ -11,6 +11,7 @@ import (
 	"github.com/nspcc-dev/neofs-s3-gw/api/cache"
 	"github.com/nspcc-dev/neofs-s3-gw/creds/accessbox"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/session/v2"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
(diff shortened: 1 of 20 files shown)


internal/neofs/neofs.go
```diff
@@ -657,20 +667,31 @@ func (x *AuthmateNeoFS) CreateContainer(ctx context.Context, prm authmate.PrmCon
 	})
 }
 
-// ReadObjectPayload implements authmate.NeoFS interface method.
-func (x *AuthmateNeoFS) ReadObjectPayload(ctx context.Context, addr oid.Address) ([]byte, error) {
+// ReadObject implements authmate.NeoFS interface method.
+func (x *AuthmateNeoFS) ReadObject(ctx context.Context, addr oid.Address) ([]byte, string, error) {
```
roman-khimov: Too specific at this level.

creds/tokens/credentials.go
```diff
@@ -97,26 +96,28 @@ func (c *cred) GetBox(ctx context.Context, addr oid.Address) (*accessbox.Box, er
...
 		return nil, fmt.Errorf("put box into cache: %w", err)
 	}
 
 	return cachedBox, nil
 }
 
-func (c *cred) getAccessBox(ctx context.Context, addr oid.Address) (*accessbox.AccessBox, error) {
-	data, err := c.neoFS.ReadObjectPayload(ctx, addr)
+func (c *cred) getAccessBox(ctx context.Context, addr oid.Address) (*accessbox.AccessBox, string, error) {
+	data, namespace, err := c.neoFS.ReadObject(ctx, addr)
```
roman-khimov: You can have namespace written into access box data as well.

roman-khimov (commented):
How about object fetching and other things that can be done with buckets?

creds/tokens/credentials.go
smallhive: As I understand it, there is no such functionality, and the namespace can be found only in attributes

internal/neofs/neofs.go

smallhive (commented):
Updated bucket search

roman-khimov (approved):

PR #1305, opened 2026-06-30
Bind gateway to namespace

### LINKED ISSUE #1302: Namespace-bound gateway
We should have a configurable mode where all requests are treated as belonging to some namespace and no other namespaces are allowed at all (as per accessbox data).

smallhive:
Closes #1302.

### DIFF
--- a/api/auth/center.go
+++ b/api/auth/center.go
@@ -50,6 +50,7 @@ type (
 		postReg                    *RegexpSubmatcher
 		cli                        tokens.Credentials
 		allowedAccessKeyIDPrefixes []string // empty slice means all access key ids are allowed
+		namespace                  string
 	}
 
 	prs int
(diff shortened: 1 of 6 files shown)


api/auth/center.go
```diff
@@ -192,6 +202,11 @@ func (c *center) Authenticate(r *http.Request) (*Box, error) {
 				return c.checkFormData(r)
 			}
 
+			// The gateway is bound to a namespace, so anonymous requests are not allowed.
```
roman-khimov: Why? That's one of the features of namespaced gateway, it processed unauthenticated requests relative to the configured namespace.

smallhive: I wanted stricter rules, removed this check

roman-khimov (approved):

PR #1307, opened 2026-07-01
Update authmate

### LINKED ISSUE #1306: Authmate is outdated as well as its documentation
## Current Behavior
1. `--bearer-rules` flag which does nothing.
2. `--gate-public-key` really demands public keys.
3. `--session-tokens` only works with container-level ops.
4. `--session-tokens` does not support setattribute/removeattribute.

## Expected Behavior
1. Bearer rules gone forever.
2. Accounts are accepted instead of public keys.
3. Object-level operations should be configurable for session tokens.
4. All container-level operations should be supported.
5. docs/authmate.md should be updated.

## Possible Solution
Fix it.

## Steps to Reproduce
Just try to do something with authmate according to docs and flags.

## Context
It's unbelievable we're in this state after token v2 adoption and token v1 deletion.

## Regression
Technically, yes. Some years ago it was OK. Now it's not.

## Your Environment
* Version of the product used: 0.44.0+

smallhive:
Closes #1306

### DIFF
--- a/authmate/authmate.go
+++ b/authmate/authmate.go
@@ -106,7 +106,6 @@ type (
 		Container             ContainerOptions
 		NeoFSKey              *keys.PrivateKey
 		GatesPublicKeys       []*keys.PublicKey
-		EACLRules             []byte
 		SessionTokenRules     []byte
 		SkipSessionRules      bool
 		Lifetime              time.Duration
--- a/authmate/authmate.go
+++ b/authmate/authmate.go
@@ -403,20 +403,12 @@ func buildSessionTokenV2(key *keys.PrivateKey, lifetime lifetimeOptions, context
 func buildSessionTokensV2(key *keys.PrivateKey, lifetime lifetimeOptions, ctxs []sessionTokenContext, gatesKeys []*keys.PublicKey, ephemeralKey *keys.PrivateKey, secret []byte) ([]session2.Token, error) {
 	var (
 		verbsByCnr = make(map[cid.ID][]session2.Verb)
-
-		objectOperations = []session2.Verb{
-			session2.VerbObjectPut,
-			session2.VerbObjectGet,
-			session2.VerbObjectHead,
-			session2.VerbObjectSearch,
-			session2.VerbObjectDelete,
-			session2.VerbObjectRange,
-		}
 	)
 
 	for _, c := range ctxs {
 		switch c.verb {
 		case session2.VerbContainerPut, session2.VerbContainerDelete, session2.VerbContainerSetEACL, session2.VerbContainerSetAttribute, session2.VerbContainerRemoveAttribute:
+		case session2.VerbObjectPut, session2.VerbObjectGet, session2.VerbObjectHead, session2.VerbObjectSearch, session2.VerbObjectDelete, session2.VerbObjectRange:
 		default:
 			return nil, fmt.Errorf("unknown verb: %v", c.verb)
 		}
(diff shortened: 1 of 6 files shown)


smallhive:
Replacing pub-keys with accounts is a bit complicated. They are used in AccessBox [cryptography](https://github.com/nspcc-dev/neofs-s3-gw/blob/38ace4107e1fb2dd0a0c1d2f5e431b21cceb67a4/internal/accessbox/accessbox.go#L23). IIRC, we already tried to eliminate keys, but eventually decided to leave this part as is, because we didn't want to leave this data without protection

roman-khimov:
> Replacing pub-keys with accounts is a bit complicated.

OK.

authmate/session_tokens.go
```diff
@@ -20,11 +20,20 @@ type (
...
+	containerSessionVerbSetEACL         = "SETEACL"
+	containerSessionVerbSetAttribute    = "SETATTRIBUTE"
+	containerSessionVerbRemoveAttribute = "REMOVEATTRIBUTE"
+
+	objectSessionVerbPut    = "OBJECT_PUT"
+	objectSessionVerbGet    = "OBJECT_GET"
+	objectSessionVerbHead   = "OBJECT_HEAD"
+	objectSessionVerbSearch = "OBJECT_SEARCH"
+	objectSessionVerbDelete = "OBJECT_DELETE"
+	objectSessionVerbRange  = "OBJECT_RANGE"
```
roman-khimov: Don't you have strings defined in session package?

authmate/authmate.go
```diff
@@ -426,7 +418,7 @@ func buildSessionTokensV2(key *keys.PrivateKey, lifetime lifetimeOptions, ctxs [
 			continue
 		}
 
-		verbsByCnr[c.containerID] = append(verbsByCnr[c.containerID], append(objectOperations, c.verb)...)
+		verbsByCnr[c.containerID] = append(verbsByCnr[c.containerID], c.verb)
```
roman-khimov, line 421: The default has changed here. It's kinda OK for constrained setups (you get what you have configured), but one thing I'd like to have is token issued with all privileges granted by default. This simplifies tests a lot.

authmate/session_tokens.go
smallhive: Not as constants, but I've managed to reuse them as well

authmate/authmate.go
smallhive: It is not an issue, because I've [extended/moved](https://github.com/nspcc-dev/neofs-s3-gw/pull/1307/changes#diff-700a3b052133237a00e490c478282e72e1c1605d9f580128b2a40d4666864676R117) these operations to the default case.

I double-checked if not passing the rules, the session token with 11 verbs (5 container and 6 object) was generated

authmate/session_tokens.go
```diff
@@ -20,11 +20,13 @@ type (
...
+// JSON strings for supported session verbs.
 const (
-	containerSessionVerbPut     = "PUT"
-	containerSessionVerbDelete  = "DELETE"
-	containerSessionVerbSetEACL = "SETEACL"
+	containerSessionVerbPut             = "PUT"
+	containerSessionVerbDelete          = "DELETE"
+	containerSessionVerbSetEACL         = "SETEACL"
+	containerSessionVerbSetAttribute    = "SETATTRIBUTE"
+	containerSessionVerbRemoveAttribute = "REMOVEATTRIBUTE"
```
roman-khimov: Make it accept SDK strings for container ops as well (in addition to old ones), we'll migrate to them.

smallhive: Ok, updated

roman-khimov (approved):

PR #1320, opened 2026-07-31
Use io.ReaderFrom

### LINKED ISSUE #1318: Use io.ReaderFrom in object write paths
Update to SDK RC21, use `ReadFrom()` in all object-related paths (maybe except when we have a complete payload in buffer already and `Write()` is not much different then).

smallhive:
Closes #1318.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This document outlines major changes between releases.
 ### Added
 
 ### Changed
+- Use io.ReaderFrom in object write paths (#1318)
```
roman-khimov: No one can understand it.

internal/neofs/neofs.go
```diff
@@ -458,27 +458,16 @@ func (x *NeoFS) putReadyObject(ctx context.Context, signer user.Signer, sessionv
...
-	var (
-		chunk        *[]byte
-		returnToPool bool
-	)
-
-	if pldSize > 0 && pldSize < uint64(x.MaxObjectSize()/2) {
-		c := make([]byte, pldSize)
-		chunk = &c
+	// Payloads that are already fully buffered (like bytes.Reader) are sent as is.
+	if wt, ok := pldRdr.(io.WriterTo); ok {
```
roman-khimov: Not worth an additional check. I thought we have some internal things like `write_into_object([]byte)` where a simple `Write()` would suffice (instead of creating `bytes.Reader`), but if that's not the case and we always have a reader from upper layers then `ReadFrom()` does the job as well.

CHANGELOG.md
roman-khimov: Something like "optimized object write buffering" is more appropriate for general audience.

internal/neofs/neofs.go

CHANGELOG.md
smallhive: Updated

roman-khimov (approved):

PR #1321, opened 2026-07-31
Hide NeoFS system attributes from GetObject.Metadata

smallhive:


api/handler/get.go
```diff
@@ -106,7 +106,7 @@ func writeHeaders(h http.Header, requestHeader http.Header, info *data.ObjectInf
 	}
 
 	for key, val := range info.Headers {
-		if layer.IsSystemHeader(key) {
+		if layer.IsSystemHeader(key) || layer.IsNeoFSSystemAttribute(key) {
```
roman-khimov: I thought `IsSystemHeader()` is supposed to do that already, why a separate function?

smallhive: I had the same opinion, but it seems _System_ in this context is the S3 gate (it filters our gate's internal things), not NeoFS. Removed function, moved check into

roman-khimov (approved):

